PR #108, opened 2024-06-27
Update the bids state which are not submitted to chain for an auction

danimhr:
This PR aims to update the bids state for those which lost the auction without being submitted to the chain.

### DIFF
--- a/auction-server/src/state.rs
+++ b/auction-server/src/state.rs
@@ -747,15 +747,29 @@ impl Store {
             }
             BidStatus::Lost { result: _, index } => {
                 if let Some(auction) = auction {
-                    query_result = sqlx::query!(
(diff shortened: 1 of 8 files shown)


auction-server/src/state.rs
```diff
@@ -747,15 +747,29 @@ impl Store {
...
+                            query_result = sqlx::query!(
+                                "UPDATE bid SET status = $1, auction_id = $2 WHERE id = $3 AND status = 'pending'",
+                                updated_status as _,
+                                auction.id,
+                                bid.id
+                            )
+                            .execute(&self.db)
+                            .await?;
+                        }
+                    }
```
m30m, line 772: out of curiosity, is it possible for index to not be None but auction to be None? The query on line 775 seems weird since there is no auction but we are updating the bundle index. Perhaps we should enforce these relationships in another PR

m30m (approved):

anihamde (approved):

PR #110, opened 2024-06-28
Add more logs

danimhr:
This PR aims to add more logs for auction server

### DIFF
--- a/auction-server/src/state.rs
+++ b/auction-server/src/state.rs
@@ -1006,8 +1006,8 @@ impl Store {
         from_time: Option<OffsetDateTime>,
     ) -> Result<Vec<OpportunityParamsWithMetadata>, RestError> {
         let mut query = QueryBuilder::new("SELECT * from opportunity where chain_id = ");
-        query.push_bind(chain_id);
-        if let Some(permission_key) = permission_key {
+        query.push_bind(chain_id.clone());
+        if let Some(permission_key) = permission_key.clone() {
             query.push(" AND permission_key = ");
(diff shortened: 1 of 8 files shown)


auction-server/src/state.rs
```diff
@@ -1065,11 +1091,15 @@ impl Store {
...
-                let result: anyhow::Result<SimulatedBid> = (b, auction).try_into();
+                let result: anyhow::Result<SimulatedBid> = (b.clone(), auction).try_into();
                 match result {
                     Ok(bid) => Some(bid),
                     Err(e) => {
-                        tracing::error!("Failed to convert bid to SimulatedBid: {}", e);
+                        tracing::error!(
+                            "Failed to convert bid to SimulatedBid: {} - bid: {:?}",
+                            e,
+                            b
```
m30m, line 1101: nit: rename to bid

auction-server/src/state.rs
```diff
@@ -1027,8 +1052,9 @@ impl Store {
             .collect();
         parsed_opps.map_err(|e| {
             tracing::error!(
-                "Failed to convert opportunity to OpportunityParamsWithMetadata: {}",
-                e
+                "Failed to convert opportunity to OpportunityParamsWithMetadata: {} - query: {:?}",
+                e,
+                query.sql(),
```
m30m, line 1057: this is not gonna include the actual parameters in the query but just $1, $2, etc. Not sure if this is actually what you wanted

m30m (approved):

PR #111, opened 2024-06-28
make fix, add getOwner

anihamde:


m30m (approved):
Thanks for updating the docs

PR #113, opened 2024-06-29
Add opportunity provider script

danimhr:
This PR aims to provide a script for the opportunity provider


scripts/opportunity-provider/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "build": "tsc",
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "provide-opportunity": "npm run build && node dist/index.js"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
m30m, line 14: Apache

scripts/opportunity-provider/index.ts
```diff
@@ -0,0 +1,223 @@
...
+  console.log(data);
+}
+
+const argv = yargs(hideBin(process.argv))
+  .option("private-key", {
+    description:
+      "Private key to provide opportunity in hex format with 0x prefix. e.g: 0xdeadbeef...",
+    type: "string",
+    demandOption: true,
+  })
```
m30m, line 208: maybe add options for config and opportunity path with default values set to `config.json` and `opportunity.json`. I can imagine having multiple pre-built opportunities for testing different scenarios and I'd rather not change the same file over and over again.

m30m (approved):

scripts/opportunity-provider/index.ts
anihamde: could also add multiple opportunities to `opportunity.json`. i guess it depends in part on whether we want to create the same set of opportunities again and again. if so, then i'd say just add to a single json.

contracts/script/Vault.s.sol
```diff
@@ -149,6 +150,24 @@ contract VaultScript is Script {
...
+
+        OpportunityProvider provider = new OpportunityProvider(
+            admin,
+            expressRelay,
+            permit2
+        );
+        vm.stopBroadcast();
+        console.log("Provoder at", address(provider));
+        return address(provider);
+    }
```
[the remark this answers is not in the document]
danimhr: Where do we use deployAll?

PR #118, opened 2024-07-30
add multicall adapter

anihamde:


contracts/test/multicall-adapter/MulticallAdapter.sol
```diff
@@ -0,0 +1,97 @@
...
+    function _callTargetContract(
+        address targetContract,
+        bytes calldata targetCalldata,
+        uint256 targetCallValue
+    ) internal {
+        (bool success, bytes memory returnData) = targetContract.call{
+            value: targetCallValue
+        }(targetCalldata);
+        if (!success) {
+            revert MulticallAdapterTargetCallFailed(returnData);
```
m30m, line 50: maybe we can add the targetCall index to the revert for easier debugging

contracts/test/multicall-adapter/MulticallAdapter.sol
```diff
@@ -0,0 +1,97 @@
...
+    function multicall(
+        MulticallParams calldata params
+    ) public payable nonReentrant {
+        _transferSellTokens(params.sellTokens);
+
+        for (uint i = 0; i < params.targetCalls.length; i++) {
+            _makeTargetCall(params.targetCalls[i]);
+        }
+
+        _sweepTokens(params.buyTokens);
```
m30m, line 91: The fact that we don't look at the buyTokens `amount` field and we just use the `token` field for sweeping is unnatural.
I would rather have an `address[] tokenToSweep` in `MulticallParams` and use that instead of sweeping every token address we got.
We can do this later on and just add a TODO for now

contracts/test/OpportunityAdapterIntegration.t.sol
```diff
@@ -2,12 +2,12 @@
 pragma solidity ^0.8.13;
 
 import {Test} from "forge-std/Test.sol";
-import "forge-std/console.sol";
 import "openzeppelin-contracts/contracts/proxy/ERC1967/ERC1967Proxy.sol";
 
 import "src/express-relay/Errors.sol";
 import "src/opportunity-adapter/OpportunityAdapter.sol";
 import {OpportunityAdapterFactory} from "src/opportunity-adapter/OpportunityAdapterFactory.sol";
+import {MulticallAdapter, TargetCall as TargetCall, TokenToSend as TokenToSend, TokenAmount as MulticallAdapterTokenAmount, MulticallParams} from "test/multicall-adapter/MulticallAdapter.sol";
```
m30m, line 10: `TargetCall as TargetCall, TokenToSend as TokenToSend` this part can be simplified

contracts/test/OpportunityAdapterIntegration.t.sol
```diff
@@ -293,6 +306,173 @@ contract OpportunityAdapterIntegrationTest is
         );
     }
 
+    function testExecutionWithBidAndCallValueViaSwaps(
```
m30m, line 309: I know that the scenario in this test case is complicated and I think it's worth a few lines of comment to explain what's going on. Maybe even some ascii art to show the flow

PR #120, opened 2024-08-01
Add logs for bid limits on amount and gas

danimhr:
This PR aims to add logging for low bid amount or high gas usage of submitted bids.

### DIFF
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -648,6 +648,13 @@ where
     if bid_amount >= minimum_bid_amount {
         Ok(())
     } else {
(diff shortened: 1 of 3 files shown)


auction-server/src/auction.rs
```diff
@@ -662,6 +669,11 @@ async fn verify_bid_under_gas_limit(
 ) -> Result<(), RestError> {
     if chain_store.block_gas_limit < estimated_gas * multiplier {
```
anihamde, line 670: nit: can we rename `multiplier` here to something better?

anihamde (approved):

PR #122, opened 2024-08-07
dump swap beacon

anihamde:


danimhr (commented):
@anihamde @m30m  I debugged it and will clean in on Monday. It's tested and working. Only need to add deployment and CI stuff + logging.

scripts/swap-beacon/src/adapter/odos.ts
```diff
@@ -0,0 +1,171 @@
...
+    const decimalsOut = tokenMap[tokenOut.toLowerCase()].decimals as number;
+
+    const [priceIn, priceOut] = await Promise.all([
+      this.getPrice(chainId, tokenIn),
+      this.getPrice(chainId, tokenOut),
+    ]);
+
+    return BigInt(
+      Math.ceil(
+        ((((Number(amountOut) * priceOut) / priceIn) * 10 ** decimalsIn) /
```
m30m, line 66: I find this formula a bit unintuitive. maybe something like this is easier to understand?

`amountOut * (priceOut/(10**decimalsOut) / (priceIn/(10**decimalsIn))`
And then we can store `(priceOut/(10**decimalsOut) / (priceIn/(10**decimalsIn))` as conversion rate?

scripts/swap-beacon/src/adapter/odos.ts
```diff
@@ -0,0 +1,171 @@
...
+        },
+        destination: responseTx.data.transaction.to,
+      })
+    );
+    const tokensToReceive: TokenAmount[] = responseTx.data.outputTokens.map(
+      (outputToken: { tokenAddress: Address; amount: string }) => ({
+        token: outputToken.tokenAddress as Address,
+        amount: amountOut
+          ? outputToken.amount
+          : Math.floor(Number(outputToken.amount) * 0.995).toString(),
```
m30m, line 157: If this number is related to the `1.005` constant defined above, let's derive them from the same constant

scripts/swap-beacon/src/index.ts
```diff
@@ -0,0 +1,308 @@
...
+  .parseSync();
+
+async function run() {
+  const beacon = new SwapBeacon(argv.endpoint, argv["chain-ids"].split(","));
+  await beacon.start();
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore: Unreachable code error
+BigInt.prototype.toJSON = function (): string {
```
m30m, line 304: is it worth doing this if it's only for  `console.log`?

scripts/swap-beacon/src/index.ts
```diff
@@ -0,0 +1,308 @@
...
+    if (
+      opportunity.targetContract.toLowerCase() ===
+      swapAdapterConfig.multicallAdapter.toLowerCase()
+    ) {
+      return;
+    }
+
+    await Promise.all(
+      swapAdapterConfig.liquidAssets.map(async (base) => {
+        const convertedOpportunity = await this.convertOpportunity(
```
m30m, line 247: does a single error here cause the whole script to exit? Is this a good behavior?

scripts/swap-beacon/src/adapter/odos.ts
```diff
@@ -0,0 +1,171 @@
...
+    );
+  }
+
+  private getTokenMap(chainId: string, tokenIn: Address, tokenOut: Address) {
+    const tokenMap = this.tokenMap[chainId];
+    if (!tokenMap) {
+      throw new Error("Chain Id not supported");
+    }
+    if (!tokenMap[tokenIn.toLowerCase()]) {
+      throw new Error("Token In not supported");
```
m30m, line 33: We can define a custom error here similar to `SwapBeaconError`

scripts/swap-beacon/src/index.ts
```diff
@@ -0,0 +1,308 @@
...
+    const originalTargetCall = {
+      targetContract: opportunity.targetContract,
+      targetCalldata: opportunity.targetCalldata,
+      targetCallValue: opportunity.targetCallValue,
+      tokensToSend: opportunity.sellTokens.map((tokenAmount) => ({
+        tokenAmount: tokenAmount,
+        destination: opportunity.targetContract,
+      })),
+    };
+    const swapsSellTargetCalls = swapsSell.map((swap) => ({
```
m30m, line 69: why are these maps necessary?

scripts/swap-beacon/src/index.ts
```diff
@@ -0,0 +1,308 @@
...
+  private createSwapOpportunity(
+    opportunity: Opportunity,
+    base: Address,
+    swapsSell: ExtendedTargetCall[],
+    swapsBuy: ExtendedTargetCall[]
+  ): Opportunity | undefined {
+    const targetContract =
+      SWAP_ADAPTER_CONFIGS[opportunity.chainId].multicallAdapter;
+    const targetCallValue = BigInt(
+      BigInt(
```
m30m, line 102: no need to wrap around BigInt

scripts/swap-beacon/src/index.ts
```diff
@@ -0,0 +1,308 @@
...
+        swapsSell.reduce(
+          (prev, curr) => BigInt(prev + curr.targetCallValue),
+          0n
+        ) +
+          swapsBuy.reduce(
+            (prev, curr) => BigInt(prev + curr.targetCallValue),
```
m30m, line 108: no need to wrap around BigInt

danimhr: On my local it won't work without wrapping.
I'll use docker and test if I can get rid of these bigint wrapping

danimhr: It was a bug

m30m (approved):
nicely done

PR #123, opened 2024-08-11
add express relay skeleton

anihamde:


express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        let balance_protocol_fee_receiver = protocol_fee_receiver.to_account_info().lamports();
+        // TODO: use actual protocol pda datalen here?
+        let rent_protocol_fee_receiver = Rent::default().minimum_balance(0).max(1);
+        let amount_protocol: u64;
+        if balance_protocol_fee_receiver >= rent_protocol_fee_receiver {
+            amount_protocol = fee_protocol;
+        } else {
+            amount_protocol = fee_protocol + (rent_protocol_fee_receiver - balance_protocol_fee_receiver);
```
m30m, line 119: this would fail if the bid amount can not cover the rent amounts. I think it's ok to be dumber here and just check if this account is already initialized and owned by the protocol. It also forces the protocol to have some instructions for initializing this account and hopefully, for withdrawing fees too

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        let balance_relayer_fee_receiver = relayer_fee_receiver.to_account_info().lamports();
+        // TODO: use actual relayer fee receiver datalen here?
+        let rent_relayer_fee_receiver = Rent::default().minimum_balance(0).max(1);
```
m30m, line 124: what's the purpose of the `max(1)`?

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+        validate_fee_split(data.split_protocol)?;
+
+        ctx.accounts.protocol_config.split = data.split_protocol;
+
+        Ok(())
+    }
+
+    pub fn permission(ctx: Context<Permission>, data: PermissionArgs) -> Result<()> {
+        if data.deadline < Clock::get()?.unix_timestamp as u64 {
+            return Err(ExpressRelayError::DeadlinePassed.into());
```
m30m, line 67: I guess you can use `err!` here for consistency?

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+
+        let express_relay_metadata_account_info = express_relay_metadata.to_account_info();
+        let rent_express_relay_metadata = Rent::default().minimum_balance(express_relay_metadata_account_info.data_len()).max(1);
+
+        if express_relay_metadata_account_info.lamports() <= rent_express_relay_metadata {
+            return err!(ExpressRelayError::InsufficientWithdrawalFunds);
+        }
+
+        let amount = express_relay_metadata.to_account_info().lamports() - rent_express_relay_metadata;
+        transfer_lamports(&express_relay_metadata_account_info, &admin.to_account_info(), amount)
```
m30m, line 192: I'm not sure how convenient it is to transfer the funds to the admin account. In case of governance, this is probably a PDA and it might not be easy to transfer it from there. The easier way is to just accept another account here and use that instead

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// CHECK: this is the protocol/router address
+    pub protocol: UncheckedAccount<'info>,
+    /// CHECK: this cannot be checked against ConfigProtocol bc it may not be initialized bc anchor :(
+    #[account(seeds = [SEED_CONFIG_PROTOCOL, protocol.key().as_ref()], bump)]
+    pub protocol_config: UncheckedAccount<'info>,
+
+    /// CHECK: this is just a PK for the relayer to receive fees at
+    #[account(mut)]
+    pub fee_receiver_relayer: UncheckedAccount<'info>,
+	/// CHECK: don't care what this PDA looks like
```
m30m, line 286: fix indentation

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+#[derive(Accounts)]
+pub struct Permission<'info> {
+    #[account(mut)]
+    pub searcher: Signer<'info>,
+    #[account(mut)]
+    pub relayer_signer: Signer<'info>,
+
+    /// CHECK: this is the permission_key
+    pub permission: UncheckedAccount<'info>,
+
```
m30m, line 276: spacing is not consistent

m30m (commented):
I still need to review the tests.
Later on we may need to add optimizations for hot accounts. We now have a writable account per protocol which makes the transactions within a protocol unparallelisable.  We also have the relayer fee receiver which makes the all the transactions unparallelisable. Both are fixable but would take time

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub split_protocol_default: u64,
+    pub split_relayer: u64
+}
+
+#[derive(Accounts)]
+pub struct Initialize<'info> {
+    #[account(mut)]
+    pub payer: Signer<'info>,
+    #[account(init, payer = payer, space = RESERVE_EXPRESS_RELAY_METADATA, seeds = [SEED_METADATA], bump)]
+    pub express_relay_metadata: Account<'info, ExpressRelayMetadata>,
```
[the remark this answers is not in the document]
anihamde: the word `config` is already used for the protocol config, i think it would be more confusing to use `config` rather than stick with `metadata`

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+            split_protocol = split_protocol_default;
+        }
+
+        let fee_protocol = bid_amount * split_protocol / FEE_SPLIT_PRECISION;
+        if fee_protocol > bid_amount {
+            return err!(ExpressRelayError::FeesTooHigh);
+        }
+
+        let fee_relayer = bid_amount.saturating_sub(fee_protocol) * split_relayer / FEE_SPLIT_PRECISION;
+        if fee_relayer.checked_add(fee_protocol).unwrap() > bid_amount {
```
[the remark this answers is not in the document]
anihamde: checked_add returns an Option, so can't use ? in this function (which returns a Result) to unwrap

express_relay/programs/express_relay/src/lib.rs
anihamde: i agree. there is a chance we may want to make it possible for `fee_receiver_protocol` to be an arbitrary address (e.g. for frontends), but not 100% sure yet. will leave it as a specified pda for now and we can change it in the future if needed

express_relay/programs/express_relay/tests/test_permission.rs
```diff
@@ -0,0 +1,103 @@
...
+    let split_relayer: u64 = 5000;
+
+    let searcher = Keypair::new();
+
+    let protocol = Keypair::new().pubkey();
+    let fee_receiver_protocol = Pubkey::find_program_address(&[SEED_EXPRESS_RELAY_FEES], &protocol).0;
+
+    program_test.add_account(
+        payer.pubkey(),
+        Account {
```
m30m, line 31: we can reduce the duplicate logic here. the Account structure is the same in all these `add_account`s

express_relay/programs/express_relay/tests/test_set_admin.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use anchor_lang::prelude::*;
+use solana_program_test::{ProgramTest, tokio};
+use solana_sdk::{account::Account, signature::Keypair, signer::Signer};
+use express_relay::state::ExpressRelayMetadata;
+use helpers::{initialize, set_admin};
+
+#[tokio::test]
+async fn test_set_admin() {
+    let mut program_test = ProgramTest::new(
```
m30m, line 11: about 50% of this test is the setup. We can make another function that initializes the program and returns all the necessary setup args

express_relay/programs/express_relay/tests/test_withdraw_fees.rs
```diff
@@ -0,0 +1,63 @@
...
+    program_context
+        .banks_client
+        .process_transaction(transfer_tx)
+        .await
+        .unwrap();
+
+    let fee_receiver_admin = Keypair::new().pubkey();
+    let (balance_express_relay_metadata, balance_admin) = withdraw_fees(&mut program_context, admin, fee_receiver_admin).await;
+
+    let rent_express_relay_metadata = Rent::default().minimum_balance(RESERVE_EXPRESS_RELAY_METADATA).max(1);
```
m30m, line 59: no need to have max

express_relay/programs/express_relay/src/lib.rs
m30m: it seems that this line has a tab character instead of 4 space characters which is why it's shown like this:
image

express_relay/testing/src/express_relay/initialize.rs
```diff
@@ -0,0 +1,39 @@
...
+                split_relayer: split_relayer,
+            }
+        }.data(),
+        accounts: Initialize {
+            payer: payer.pubkey(),
+            express_relay_metadata: express_relay_metadata,
+            admin: admin,
+            relayer_signer: relayer_signer,
+            fee_receiver_relayer: fee_receiver_relayer,
+            system_program: system_program_pk,
```
m30m, line 33: no need to define the `system_program_pk` variable. Just use `system_program::ID` directly

express_relay/testing/tests/initialize.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let setup_params = SetupParams {
+        split_protocol_default,
+        split_relayer,
+    };
+    let setup_result = setup(setup_params);
+
+    match setup_result {
+        Ok(_) => panic!("expected setup to fail"),
+        Err(err) => assert_custom_error(err, 0, 6000)
```
m30m, line 38: these assertions are a bit fragile and unreadable. We can use something like this `express_relay::error::ErrorCode::FeeSplitLargerThanPrecision.into()` instead of `6000`

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+        Ok(())
+    }
+
+    pub fn permission(ctx: Context<Permission>, data: PermissionArgs) -> Result<()> {
+        if data.deadline < Clock::get()?.unix_timestamp {
+            return err!(ErrorCode::DeadlinePassed);
+        }
+
+        // check that not cpi
+        let instruction = get_instruction_relative(0, &ctx.accounts.sysvar_instructions.to_account_info())?;
```
m30m, line 79: I just found this:
https://docs.rs/solana-program/latest/solana_program/instruction/fn.get_stack_height.html

Seems like an easier way to detect cpi by doing something like this:
```
if get_stack_height() != TRANSACTION_LEVEL_STACK_HEIGHT {
      return err!(ErrorCode::InvalidCPIPermission);
  }
```

m30m (commented):
sorry I haven't still read all the tests but the structure looks solid now 👍

express_relay/yarn.lock
danimhr, line 1: as we are using npm as package manager inside the per, I think it's better to use npm here as well instead of yarn. Is it sth that we setup or other tools are using yarn by default?

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,359 @@
...
+            &searcher.to_account_info(),
+            &express_relay_metadata.to_account_info(),
+            bid_amount.saturating_sub(fee_protocol).saturating_sub(fee_relayer),
+            ctx.accounts.system_program.to_account_info()
+        )?;
+
+        Ok(())
+    }
+
+    pub fn check_permission(ctx: Context<CheckPermission>) -> Result<()> {
```
guibescos, line 162: have you considered making this an sdk method an not a cpi? This doesn't need to be a CPI.

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,359 @@
...
+        if bid_amount + rent_searcher > searcher.lamports() {
+            return err!(ErrorCode::InsufficientSearcherFunds);
+        }
+
+        let express_relay_metadata = &ctx.accounts.express_relay_metadata;
+        let split_relayer = express_relay_metadata.split_relayer;
+        let split_protocol_default = express_relay_metadata.split_protocol_default;
+
+        let split_protocol: u64;
+        let protocol_config = &ctx.accounts.protocol_config;
```
guibescos, line 98: You can use the anchor feature `Option<Account<'info,ConfigProtocol>` for this case

express_relay/programs/express_relay/src/lib.rs
anihamde: we want integrated contracts to have a check for permissioning within their program logic, that way express relay permissioning cannot be foregone

express_relay/programs/express_relay/src/lib.rs
anihamde: nice will take a look

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,359 @@
...
+
+#[derive(Accounts)]
+pub struct Permission<'info> {
+    #[account(mut)]
+    pub searcher: Signer<'info>,
+
+    pub relayer_signer: Signer<'info>,
+
+    /// CHECK: this is the permission_key
+    pub permission: UncheckedAccount<'info>,
```
guibescos, line 306: I find it weird that this account is not checked anywhere

express_relay/programs/express_relay/src/lib.rs
guibescos: What I mean is that the integrated program can perform these checks without CPIng

express_relay/programs/express_relay/src/lib.rs
anihamde: i think this actually doesn't work here unfortunately. even when we aren't using this account bc the protocol doesn't have a `protocol_config`, we still need to check it to make sure that a user hasn't just excluded it from the instruction.

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,366 @@
...
+        validate_fee_split(data.split_protocol)?;
+
+        ctx.accounts.protocol_config.protocol = *ctx.accounts.protocol.key;
+        ctx.accounts.protocol_config.split = data.split_protocol;
+
+        Ok(())
+    }
+
+    // Permissions a particular (protocol, permission) pair and distributes bids according to splits
+    pub fn permission(ctx: Context<Permission>, data: PermissionArgs) -> Result<()> {
```
danimhr, line 81: @anihamde as a suggestion, maybe we could extract multiple functions here?

express_relay/programs/express_relay/src/lib.rs
jayantk: one advantage of the CPI is that the DAO controls the code, so we can upgrade it later. Harder to do that if it's in a client SDK used by a bunch of different programs.

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,366 @@
...
+        } else {
+            split_protocol = split_protocol_default;
+        }
+
+        let fee_protocol = bid_amount * split_protocol / FEE_SPLIT_PRECISION;
+        if fee_protocol > bid_amount {
+            return err!(ErrorCode::FeesHigherThanBid);
+        }
+
+        let fee_relayer = bid_amount.saturating_sub(fee_protocol) * split_relayer / FEE_SPLIT_PRECISION;
```
jayantk, line 119: the subtraction here seems like it will make it harder to configure the fees (?) I get that this makes it a cascade, but seems simpler to make the splits add up to 100% conceptually.

express_relay/programs/express_relay/src/lib.rs
```diff
@@ -0,0 +1,366 @@
...
+        if fee_relayer.checked_add(fee_protocol).ok_or(ProgramError::ArithmeticOverflow)? > bid_amount {
```
jayantk, line 120: I think this condition can only trigger if the splits are > 1, which I think is guarded against when the configuration is set. Both fee_relayer and fee_protocol are truncated, i.e., rounded down, so there's no way for them to be larger than the bid.

It's fine to leave this check here to be extra safe, but it's worth leaving a comment that this is defensive programming and really shouldn't happen.

jayantk (commented):
I just read the core piece of it and you can see inline comments there. I'll send follow up thoughts in slack.

express_relay/programs/express_relay/src/lib.rs
anihamde: that's fair. in a way i think it makes sense for the relayer split to be a percentage of what isn't taken by the protocol (so that relayer revenue remains linear with Express Relay revenue).

i think there are merits to both, but mostly, this structure is preferred bc it's consistent with the evm side

Tiltfile
```diff
@@ -123,3 +129,43 @@ local_resource(
...
+    # check readiness by sending a health GET query to the RPC url
+    readiness_probe=probe(
+        period_secs=10,
+        http_get = http_get_action(port=int(rpc_port_solana), host="localhost", scheme="http", path="/health")
+    ),
+    resource_deps=["build-programs"],
+)
+
+local_resource(
+    "airdrop",
```
m30m, line 154: I would call this resource and script `setup-accounts`. airdrop is a bit misleading since it actually checks and creates the keypairs if they don't exist

per_sdk/solana/airdrop.py
```diff
@@ -0,0 +1,64 @@
...
+
+    client = AsyncClient(args.rpc_url)
+
+    for account in ["searcher", "admin", "relayer_signer"]:
+        file_path = f"keypairs/{account}.json"
+        if not os.path.exists(file_path):
+            kp = Keypair()
+            with open(file_path, "w") as f:
+                json.dump(kp.to_bytes_array(), f)
+                logger.debug(f"Created and saved {account} keypair")
```
m30m, line 51: I think it's nicer to log this message by default without the need to change the logging level. Maybe change it to info or change tilt to run it with more verbosity

per_sdk/solana/dummy_tx.py
```diff
@@ -0,0 +1,182 @@
...
+    )
+    parser.add_argument(
+        "--submit-on-chain",
+        action="store_true",
+        default=False,
+        help="Submit the transaction directly on-chain instead of submitting to the server",
+    )
+    args = parser.parse_args()
+
+    logger.setLevel(logging.INFO if args.verbose == 0 else logging.DEBUG)
```
m30m, line 77: this is the same logic in all 3 scripts. we can refactor and extract it

Tiltfile
```diff
@@ -123,3 +129,43 @@ local_resource(
     serve_cmd=searcher_command,
     resource_deps=["deploy-contracts", "auction-server", "create-configs"],
 )
+
+
+# Solana resources
```
m30m, line 134: let's add svm prefix to these resources to easily distinguish them on the tilt ui
same for evm

per_sdk/solana/airdrop.py
```diff
@@ -0,0 +1,64 @@
...
+    log_handler.setFormatter(formatter)
+    logger.addHandler(log_handler)
+
+    client = AsyncClient(args.rpc_url)
+
+    for account in ["searcher", "admin", "relayer_signer"]:
+        file_path = f"keypairs/{account}.json"
+        if not os.path.exists(file_path):
+            kp = Keypair()
+            with open(file_path, "w") as f:
```
m30m, line 49: `keypairs` directory might not exist and we have to create that first if necessary

Tiltfile
m30m: another way to do this is to just create the keypairs and then start the `solana-localnet` with these accounts already topped up. This is possible via the `--account` flag and by providing such a json:
```
{
  "pubkey": "8WpVYzy6XZQy7HiRdRD4kNhGx7yr8K55QumUG2BVTyeA",
  "account": {
    "lamports": 1000000000,
    "data": [
      "",
      "base64"
    ],
    "owner": "11111111111111111111111111111111",
    "executable": false,
    "rentEpoch": 1,
    "space": 0
  }
}
```

I noticed it takes more than 30 seconds to have the airdrops so it may be worth it.

per_sdk/solana/airdrop.py
```diff
@@ -0,0 +1,64 @@
...
+    logger.setLevel(logging.INFO if args.verbose == 0 else logging.DEBUG)
+    log_handler = logging.StreamHandler()
+    formatter = logging.Formatter(
+        "%(asctime)s %(levelname)s:%(name)s:%(module)s %(message)s",
+        datefmt="%Y-%m-%d %H:%M:%S",
+    )
+    log_handler.setFormatter(formatter)
+    logger.addHandler(log_handler)
+
+    client = AsyncClient(args.rpc_url)
```
m30m, line 43: use "confirmed" as the default commitment level of the clients to speedup tx confirmations

per_sdk/solana/dummy_tx.py
```diff
@@ -0,0 +1,182 @@
...
+    )[0]
+    pk_fee_receiver_dummy = Pubkey.find_program_address(
+        [b"express_relay_fees"], dummy_pid
+    )[0]
+    pk_express_relay_metadata = Pubkey.find_program_address(
+        [b"metadata"], express_relay_pid
+    )[0]
+
+    discriminator_submit_bid = hashlib.sha256(b"global:submit_bid").digest()[:8]
+    data_submit_bid = struct.pack(
```
m30m, line 109: there must be a better way to do these in python. probably via anchor-py

per_sdk/svm/dummy_tx.py
```diff
@@ -0,0 +1,175 @@
...
+from solders.sysvar import INSTRUCTIONS as sysvar_ixs_pid
+
+from per_sdk.svm.helpers import configure_logger, read_kp_from_json
+
+logger = logging.getLogger(__name__)
+
+DEADLINE_MAX = 2**63 - 1
+
+
+async def main():
```
m30m, line 23: function is a bit too long. you can extract the parser part into a separate function.

m30m (approved):

PR #126, opened 2024-08-28
Exit on no ping

m30m:
We rely on the heartbeat the server sends to the client to make sure the connection is still alive.
By calling the terminate function on the websocket, the whole process will exit. 
Solution based on:
https://github.com/websockets/ws?tab=readme-ov-file#how-to-detect-and-close-broken-connections

### DIFF
--- a/scripts/swap-beacon/src/index.ts
+++ b/scripts/swap-beacon/src/index.ts
@@ -25,8 +25,10 @@ export function getSwapAdapterConfig(chainId: string) {
 
 export class SwapBeacon {
   private client: Client;
-  private adapters: Adapter[];
-  private chainIds: string[];
+  private pingTimeout: NodeJS.Timeout | undefined;
+  private readonly adapters: Adapter[];
+  private readonly chainIds: string[];
+  private readonly PING_INTERVAL = 30000;
 
   constructor(endpoint: string, _chainIds: string[]) {
     this.client = new Client(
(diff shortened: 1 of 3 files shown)


scripts/swap-beacon/src/index.ts
```diff
@@ -246,12 +249,25 @@ export class SwapBeacon {
     );
```
anihamde, line 249: could you add a log to state that `convertedOpportunity` with a specific ID was successfully submitted?

danimhr (approved):

scripts/swap-beacon/src/index.ts
```diff
@@ -246,12 +249,25 @@ export class SwapBeacon {
...
   }
 
+  heartbeat() {
+    if (this.pingTimeout !== undefined) clearTimeout(this.pingTimeout);
+
+    this.pingTimeout = setTimeout(() => {
+      console.error("Received no ping. Terminating connection.");
+      this.client.websocket.terminate();
```
anihamde, line 257: instead of just terminating in this case, what about retrying to establish the connection, at least once or twice? if you have multiple ping timeouts in a session, you can just terminate the websocket and end the session?

scripts/swap-beacon/src/index.ts
```diff
@@ -246,12 +249,25 @@ export class SwapBeacon {
...
+    }, this.PING_INTERVAL + 2000); // 2 seconds for latency
```
anihamde, line 258: if this terminates, does the session not just continue, with a broken WebSocket meaning no opportunities received? in that case, shouldn't we actually throw an error and exit the script, so that it can restart?

m30m: nope, when we terminate the websocket the whole script exits. Seems a bit magical though

scripts/swap-beacon/src/index.ts
m30m: When the process exits, k8s will just spawn a new one which is effectively retrying. The service is pretty stateless and a complete restart seems fine but we can add retry mechanisms if it's costly to do a full container restart.

PR #129, opened 2024-08-30
Add submit_bid instruction existance check for svm bid

danimhr:
This PR aims to check if svm bid contains exactly a single submit_bid instruction to ER program

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@ RUN rustup default nightly-2024-04-10
 RUN curl -L https://foundry.paradigm.xyz | bash
 ENV PATH="${PATH}:/root/.foundry/bin/"
 RUN foundryup
+RUN cargo install --git https://github.com/coral-xyz/anchor --tag v0.30.1 anchor-cli --locked
 
 # Add solidity dependencies
 WORKDIR /src
@@ -27,6 +28,12 @@ RUN forge install OpenZeppelin/openzeppelin-contracts@v5.0.2 --no-git --no-commi
 RUN forge install OpenZeppelin/openzeppelin-contracts-upgradeable@v4.9.6 --no-git --no-commit
 RUN forge install Uniswap/permit2@0x000000000022D473030F116dDEE9F6B43aC78BA3 --no-git --no-commit
 RUN forge install nomad-xyz/ExcessivelySafeCall@be417ab0c26233578b8d8f3a37b87bd1fcb4e286 --no-git --no-commit
+
+# Add solana dependencies
+WORKDIR /src
+COPY express_relay express_relay
+WORKDIR /src/express_relay
+RUN anchor build
 
 # Build auction-server
 WORKDIR /src
--- a/Tiltfile
+++ b/Tiltfile
@@ -80,10 +80,9 @@ cmd_button(
 )
 
 local_resource(
-    "evm-create-configs", "python3 evm_integration.py %s %s" % (rpc_url_anvil, ws_url_anvil), resource_deps=["evm-deploy-contracts"]
+    "evm-create-configs", "python3 integration.py %s %s" % (rpc_url_anvil, ws_url_anvil), resource_deps=["evm-deploy-contracts"]
 )
 
-# svm resources
 local_resource(
     "svm-build-programs",
     "cargo build-sbf && anchor build",
--- a/Tiltfile
+++ b/Tiltfile
@@ -83,11 +83,18 @@ local_resource(
     "evm-create-configs", "python3 evm_integration.py %s %s" % (rpc_url_anvil, ws_url_anvil), resource_deps=["evm-deploy-contracts"]
 )
 
+# svm resources
+local_resource(
+    "svm-build-programs",
+    "cargo build-sbf && anchor build",
+    dir="express_relay",
+)
+
 local_resource(
     "auction-server",
     serve_cmd="source ../tilt-resources.env; source ./.env; cargo run -- run --database-url $DATABASE_URL --subwallet-private-key $RELAYER_PRIVATE_KEY --secret-key $SECRET_KEY",
     serve_dir="auction-server",
-    resource_deps=["evm-create-configs"],
+    resource_deps=["evm-create-configs", "svm-build-programs"],
     readiness_probe=probe(period_secs=5, http_get=http_get_action(port=9000)),
 )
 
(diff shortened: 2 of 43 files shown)


evm_integration.py
```diff
@@ -39,6 +38,8 @@ def main():
     adapter_factory_contract: {latest_env['adapterFactory']}
     legacy_tx: false
     poll_interval: 1
+  solana:
+    express_relay_program_id: GwEtasTAxdS9neVE4GPUpcwR7DB7AizntQSPcG36ubZM
```
danimhr, line 42: @anihamde @m30m What do you thing about this part?
I couldn't find a resources file for svm tilt setup. In other files we used the actual variable and didn't read the contract address from any kind of generated config file. Should I continue with it or it's easy to create a config file from the generated accounts and addresses? (Like what we have for evm)

anihamde: The solana program addresses are being set in the `test-validator-params.sh` file. they don't have to be set there, we can add those as arguments to the call in tilt, but if so, we should create a separate `svm_integration.py` file and `local_resource` to create env vars for express relay and dummy program addresses.

danimhr: Thank you Ani.
As this value is not going to be updated, I will continue with the current version.

auction-server/src/server.rs
```diff
@@ -265,6 +312,7 @@ pub async fn start_server(run_options: RunOptions) -> anyhow::Result<()> {
         secret_key: run_options.secret_key.clone(),
         access_tokens: RwLock::new(access_tokens),
         metrics_recorder: setup_metrics_recorder()?,
+        express_relay_idl: load_express_relay_idl()?,
```
danimhr, line 315: Is this value different for each svm chain?

anihamde: it technically will be different just in the address field. but if we keep a separate config for the program id/address by chain, then we can just reuse the same IDL, similar to how it's done in the SDK PR

evm_integration.py
```diff
@@ -39,6 +38,8 @@ def main():
     adapter_factory_contract: {latest_env['adapterFactory']}
     legacy_tx: false
     poll_interval: 1
+  solana:
```
anihamde, line 41: maybe we should rename this to something like `development-solana` or `localnet`

anihamde: and can we rename this file to `chain_integration.py` or `test_chain_integration.py`? or move the solana stuff to a separate `svm_integration.py` file?

evm_integration.py
anihamde: for code quality purposes, we should probably ensure we are using the same program ID across the `test-validator-params.sh` file and here. probably easiest would be to just read what's in the `test-validator-params.sh` file here

auction-server/src/auction.rs
```diff
@@ -880,16 +882,71 @@ pub async fn run_tracker_loop(store: Arc<Store>, chain_id: String) -> Result<()>
...
+    transaction: VersionedTransaction,
+    name: String,
+    total: usize,
+) -> Result<(), RestError> {
+    if transaction
+        .message
+        .instructions()
+        .iter()
+        .filter(|instruction| {
+            let program_id = instruction.program_id(transaction.message.static_account_keys());
```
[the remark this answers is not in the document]
danimhr: I think the code is readable and I don't know what comment would help here and make it more understandable.
Would you mind if suggest what comment would help here?

auction-server/src/auction.rs
```diff
@@ -880,16 +882,71 @@ pub async fn run_tracker_loop(store: Arc<Store>, chain_id: String) -> Result<()>
     Ok(())
 }
 
+// Checks that the transaction includes exactly "total" "name" instruction to the Express Relay on-chain program
```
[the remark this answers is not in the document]
danimhr: 1. I have to either include `submit_bid` in the function name or add it as args. I prefer to add it as args as it's not creating any kind of complexity here.
2. As we don't need these fields and are not going to use it for this version of server, I prefer to add these args when we need them.

auction-server/src/state.rs
```diff
@@ -188,7 +190,7 @@ pub struct ChainStore {
     pub chain_id_num:           u64,
```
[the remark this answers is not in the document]
danimhr: Make it as enum would change lots of file and I'm planning to do these refactoring later.
I'll rename the variable

Dockerfile
```diff
@@ -28,6 +29,12 @@ RUN forge install OpenZeppelin/openzeppelin-contracts-upgradeable@v4.9.6 --no-gi
 RUN forge install Uniswap/permit2@0x000000000022D473030F116dDEE9F6B43aC78BA3 --no-git --no-commit
 RUN forge install nomad-xyz/ExcessivelySafeCall@be417ab0c26233578b8d8f3a37b87bd1fcb4e286 --no-git --no-commit
 
+# Add solana dependencies
+WORKDIR /src
+COPY express_relay express_relay
```
m30m, line 34: At some point we should probably move this to another folder since the naming is poor. We can do it separately but the later we do it, the more changes we need to make

Tiltfile
```diff
@@ -83,11 +83,18 @@ local_resource(
     "evm-create-configs", "python3 evm_integration.py %s %s" % (rpc_url_anvil, ws_url_anvil), resource_deps=["evm-deploy-contracts"]
 )
 
+# svm resources
```
m30m, line 86: this comment was meant to be for a section of the file which contains svm resources. But now since all of them have the svm prefix, we can just remove the comment

auction-server/src/server.rs
```diff
@@ -140,60 +147,97 @@ pub fn setup_metrics_recorder() -> anyhow::Result<PrometheusHandle> {
...
+                }
+            }),
+    )
     .await
     .into_iter()
     .collect()
 }
 
+pub fn load_express_relay_idl() -> anyhow::Result<Idl> {
+    let idl = fs::read("../express_relay/target/idl/express_relay.json")?;
```
m30m, line 236: I think this is a bad idea, similar to what we have on the evm side, we should include this on compile time and have proper type checking.

Dockerfile
danimhr: Not sure about naming / structure for this part. Do you have anything in mind?

auction-server/src/auction.rs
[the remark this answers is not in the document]
danimhr: This part is removed

auction-server/src/auction.rs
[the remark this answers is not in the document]
danimhr: This part is simplified and updated

m30m (approved):

PR #130, opened 2024-08-30
rename protocol -> router and reorg

anihamde:


express_relay/programs/express_relay/src/utils.rs
```diff
@@ -30,26 +30,19 @@ pub fn transfer_lamports_cpi<'info>(
...
-    if pda != &pda_calculated {
-        return err!(ErrorCode::InvalidPDAProvided);
+        transfer(CpiContext::new(system_program, cpi_accounts), amount)?;
     }
 
     Ok(())
 }
 
-pub fn num_permissions_in_tx(sysvar_instructions: UncheckedAccount, permission: Option<Pubkey>, protocol: Option<Pubkey>) -> Result<u16> {
+pub fn num_permissions_in_tx(sysvar_instructions: UncheckedAccount, permission: Option<Pubkey>, router: Option<Pubkey>) -> Result<u16> {
```
m30m, line 45: I think this function interface is a bit shaky. We have 2 usecases, one is to make sure there is only 1 instruction and the other is to make sure that it's with a specific permission and router. It should not be possible to provide only permission or only router since it doesn't make a lot of sense.

express_relay/programs/express_relay/src/utils.rs
```diff
@@ -96,72 +89,67 @@ pub fn handle_bid_payment(ctx: Context<SubmitBid>, bid_amount: u64) -> Result<()
...
-    let rent_fee_receiver_protocol = Rent::get()?.minimum_balance(0);
-    if balance_fee_receiver_protocol+fee_protocol < rent_fee_receiver_protocol {
-        return err!(ErrorCode::InsufficientProtocolFeeReceiverRent);
+    let fee_relayer = bid_amount.saturating_sub(fee_router) * split_relayer / FEE_SPLIT_PRECISION;
+    if fee_relayer.checked_add(fee_router).ok_or(ProgramError::ArithmeticOverflow)? > bid_amount {
+        // this error should never be reached due to fee split checks, but kept as a matter of defensive programming
+        return err!(ErrorCode::FeesHigherThanBid);
     }
+    if fee_relayer > 0 {
+        let fee_receiver_relayer = &ctx.accounts.fee_receiver_relayer;
```
m30m, line 134: the next 4 lines are somewhat duplicate. We can extract a function to check if fee > rent for any account

express_relay/programs/express_relay/src/utils.rs
```diff
@@ -30,26 +30,19 @@ pub fn transfer_lamports_cpi<'info>(
     amount: u64,
     system_program: AccountInfo<'info>,
 ) -> Result<()> {
-    let cpi_accounts = Transfer {
-        from: from.clone(),
-        to: to.clone(),
-    };
+    if amount > 0 {
```
m30m, line 33: it seems that we are doing this check externally too. let's remove it from here if it's already checked

express_relay/testing/src/express_relay/helpers.rs
```diff
@@ -13,24 +13,20 @@ pub fn get_express_relay_metadata(svm: litesvm::LiteSVM) -> ExpressRelayMetadata
...
-    let protocol_config_key = get_protocol_config_key(protocol);
-    let protocol_config_acc = svm.get_account(&protocol_config_key);
-    match protocol_config_acc {
-        Some(protocol_config_acc) => {
-            let protocol_config = ConfigProtocol::try_deserialize(&mut protocol_config_acc.data.as_ref()).expect("Account is not of struct ConfigProtocol");
-            return Some(protocol_config);
+pub fn get_router_config(svm: litesvm::LiteSVM, router: Pubkey) -> Option<ConfigRouter> {
+    let router_config_key = get_router_config_key(router);
+    let router_config_acc = svm.get_account(&router_config_key);
+    match router_config_acc {
```
m30m, line 23: maybe using map is cleaner

express_relay/testing/src/express_relay/set_router_split.rs
```diff
@@ -0,0 +1,32 @@
+use anchor_lang::{InstructionData, ToAccountMetas};
+use solana_sdk::{instruction::Instruction, pubkey::Pubkey, signature::Keypair, signer::Signer, system_program};
+use express_relay::{accounts::SetRouterSplit, SetRouterSplitArgs};
+
+use super::helpers::{get_express_relay_metadata_key, get_router_config_key};
+
+pub fn get_set_router_split_instruction(
```
m30m, line 7: get_set is a bit confusing. It has a different meaning in some [contexts](https://redis.io/docs/latest/commands/getset/).  maybe `make_set` or `create_set`?

express_relay/testing/src/dummy/do_nothing.rs
```diff
@@ -14,7 +15,7 @@ pub fn get_do_nothing_instruction(
             express_relay: express_relay::ID,
             sysvar_instructions: sysvar_instructions_id(),
             permission: permission_key,
-            protocol: dummy::ID,
+            router: router
```
m30m, line 18: `router: router` -> `router` ?

m30m (approved):

PR #134, opened 2024-09-04
feat: Add removal reason to db

m30m:


anihamde (approved):
lgtm, just make sure precommit checks are passing. i think the migrations haven't gone through on the build side?

auction-server/src/opportunity_adapter.rs
```diff
@@ -541,14 +547,14 @@ pub async fn run_verification_loop(store: Arc<Store>) -> Result<()> {
                     // check each of the opportunities for this permission key for validity
                     for opportunity in opportunities.iter() {
                         match verify_with_store(opportunity.clone(), &store).await {
```
danimhr, line 549: Can't we use if let here?

danimhr (commented):
Please update the `models.rs` as well.
You can use the `BidStatus` as an example.

Also isn't there an easy way to refactor both BidStatus and RemovalReason update queries in `state.rs` file with the types in models.rs?

PR #136, opened 2024-09-09
Add svm bid data extraction from transaction

danimhr:
This PR aims to extract bid data from svm transaction

### DIFF
--- a/auction-server/build.rs
+++ b/auction-server/build.rs
@@ -1,4 +1,44 @@
-use std::process::Command;
+use {
+    anchor_lang_idl::{
+        convert::convert_idl,
+        types::IdlInstructionAccountItem,
+    },
+    std::{
+        fs,
+        process::Command,
+    },
+};
+
+const SUBMIT_BID_INSTRUCTION_SVM: &str = "submit_bid";
+const PERMISSION_ACCOUNT_SVM: &str = "permission";
+
+fn verify_idl() {
+    let idl_json = fs::read("../contracts/svm/target/idl/express_relay.json")
+        .expect("Failed to read IDL JSON");
+    let express_relay_idl =
+        convert_idl(idl_json.as_slice()).expect("Failed to convert IDL to Rust");
+    match express_relay_idl
+        .instructions
+        .iter()
+        .find(|i| i.name == SUBMIT_BID_INSTRUCTION_SVM)
+    {
+        Some(instruction) => {
+            if !instruction.accounts.iter().any(|a| match a {
+                IdlInstructionAccountItem::Single(a) => a.name == PERMISSION_ACCOUNT_SVM,
+                IdlInstructionAccountItem::Composite(a) => a.name == PERMISSION_ACCOUNT_SVM,
+            }) {
+                panic!(
+                    "{} account not found in {} instruction",
+                    PERMISSION_ACCOUNT_SVM, SUBMIT_BID_INSTRUCTION_SVM
+                );
+            }
+        }
+        None => panic!(
+            "{} instruction not found in IDL",
+            SUBMIT_BID_INSTRUCTION_SVM
+        ),
+    }
+}
 
 fn main() {
     let contract_setup = r#"
@@ -24,4 +64,6 @@ fn main() {
             String::from_utf8_lossy(&output.stdout)
         );
     }
+
+    verify_idl();
 }
--- a/auction-server/build.rs
+++ b/auction-server/build.rs
@@ -9,45 +9,12 @@ use {
     },
 };
 
-const SUBMIT_BID_INSTRUCTION_SVM: &str = "submit_bid";
-const PERMISSION_ACCOUNT_SVM: &str = "permission";
-
-fn verify_idl() {
-    let idl_json = fs::read("../contracts/svm/target/idl/express_relay.json")
-        .expect("Failed to read IDL JSON");
-    let express_relay_idl =
-        convert_idl(idl_json.as_slice()).expect("Failed to convert IDL to Rust");
-    match express_relay_idl
-        .instructions
-        .iter()
-        .find(|i| i.name == SUBMIT_BID_INSTRUCTION_SVM)
-    {
-        Some(instruction) => {
-            if !instruction.accounts.iter().any(|a| match a {
-                IdlInstructionAccountItem::Single(a) => a.name == PERMISSION_ACCOUNT_SVM,
-                IdlInstructionAccountItem::Composite(a) => a.name == PERMISSION_ACCOUNT_SVM,
-            }) {
-                panic!(
-                    "{} account not found in {} instruction",
-                    PERMISSION_ACCOUNT_SVM, SUBMIT_BID_INSTRUCTION_SVM
-                );
-            }
-        }
-        None => panic!(
-            "{} instruction not found in IDL",
-            SUBMIT_BID_INSTRUCTION_SVM
-        ),
-    }
-}
-
-fn main() {
+fn build_evm_contracts() {
     let contract_setup = r#"
         cd ../contracts/evm
         forge build --via-ir
     "#;
     println!("cargo:rerun-if-changed=../contracts/evm");
-    println!("cargo:rerun-if-changed=migrations");
-
     // Build the contracts and generate the ABIs. This is required for abigen! macro expansions to work.
     let output = Command::new("sh")
         .args(["-c", contract_setup])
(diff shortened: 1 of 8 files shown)


auction-server/src/auction.rs
```diff
@@ -895,15 +908,48 @@ pub fn verify_submit_bid_instruction_svm(
...
+        _ => Err(RestError::BadParameters(
             "Bid has to include exactly one submit_bid instruction to Express Relay program"
                 .to_string(),
-        ));
+        )),
     }
-    Ok(())
+}
+
+fn find_account_position_svm(idl: Idl, instruction: &str, account: &str) -> Option<usize> {
```
m30m, line 923: I'd rather find the index in build.rs and pass it here as a constant instead of finding it here again

auction-server/src/auction.rs
```diff
@@ -895,15 +908,48 @@ pub fn verify_submit_bid_instruction_svm(
...
+    let discriminator = express_relay_svm::instruction::SubmitBid::discriminator();
+    let submit_bid_data = express_relay_svm::SubmitBidArgs::try_from_slice(
+        &instruction.data.as_slice()[discriminator.len()..],
+    )
+    .map_err(|e| RestError::BadParameters(format!("Invalid submit_bid instruction data: {}", e)))?;
+    let permission_account_position =
+        find_account_position_svm(idl, SUBMIT_BID_INSTRUCTION_SVM, PERMISSION_ACCOUNT_SVM).ok_or(
+            RestError::BadParameters("Invalid submit_bid instruction accounts".to_string()),
+        )?;
+    let account_position = instruction.accounts[permission_account_position];
```
m30m, line 948: this can lead to out of bound panics

auction-server/src/auction.rs
```diff
@@ -895,15 +908,48 @@ pub fn verify_submit_bid_instruction_svm(
...
+    Ok((
+        submit_bid_data.bid_amount,
+        accounts[account_position as usize],
```
m30m, line 951: not sure, but this may also lead to out of bound panics

m30m (approved):

auction-server/src/auction.rs
```diff
@@ -895,15 +908,48 @@ pub fn verify_submit_bid_instruction_svm(
...
+    idl: Idl,
+    accounts: &[Pubkey],
+    instruction: CompiledInstruction,
+) -> Result<(u64, Pubkey), RestError> {
+    let discriminator = express_relay_svm::instruction::SubmitBid::discriminator();
+    let submit_bid_data = express_relay_svm::SubmitBidArgs::try_from_slice(
+        &instruction.data.as_slice()[discriminator.len()..],
+    )
+    .map_err(|e| RestError::BadParameters(format!("Invalid submit_bid instruction data: {}", e)))?;
+    let permission_account_position =
```
anihamde, line 944: i would prefer you also get the router account and index by that. generally, permission keys _should_ be unique across different router opportunities, but the way the program is structured, permissioning is checked for a tuple `(permission, router)`. it would be more accurate to also extract the router account as part of the bid data

auction-server/src/auction.rs
```diff
@@ -895,15 +908,48 @@ pub fn verify_submit_bid_instruction_svm(
...
+}
+
+const SUBMIT_BID_INSTRUCTION_SVM: &str = "submit_bid";
+const PERMISSION_ACCOUNT_SVM: &str = "permission";
+
+fn extract_bid_data_svm(
+    idl: Idl,
+    accounts: &[Pubkey],
+    instruction: CompiledInstruction,
+) -> Result<(u64, Pubkey), RestError> {
```
anihamde, line 938: i think you also should extract the bid `deadline`? this could be relevant for deciding whether to include this bid in the auction or not (if deadline has already passed)

auction-server/src/auction.rs
```diff
@@ -895,15 +908,48 @@ pub fn verify_submit_bid_instruction_svm(
...
+    let discriminator = express_relay_svm::instruction::SubmitBid::discriminator();
+    let submit_bid_data = express_relay_svm::SubmitBidArgs::try_from_slice(
+        &instruction.data.as_slice()[discriminator.len()..],
+    )
+    .map_err(|e| RestError::BadParameters(format!("Invalid submit_bid instruction data: {}", e)))?;
+    let permission_account_position =
+        find_account_position_svm(idl, SUBMIT_BID_INSTRUCTION_SVM, PERMISSION_ACCOUNT_SVM).ok_or(
```
anihamde, line 945: i think you can probably run this method once upon startup and not have to run it each time?

auction-server/src/auction.rs
```diff
@@ -895,15 +908,48 @@ pub fn verify_submit_bid_instruction_svm(
...
+        IdlInstructionAccountItem::Single(a) => a.name == account,
+        IdlInstructionAccountItem::Composite(a) => a.name == account,
+    })
+}
+
+const SUBMIT_BID_INSTRUCTION_SVM: &str = "submit_bid";
+const PERMISSION_ACCOUNT_SVM: &str = "permission";
+
+fn extract_bid_data_svm(
+    idl: Idl,
```
anihamde, line 935: if you run the account index method once upon startup, you don't need to pass the IDL here

auction-server/src/auction.rs
danimhr: We will run a simulation after the extraction. If the deadline is invalid, then the simulation will be failed.

auction-server/src/auction.rs
danimhr: Is there any chance that the permission be same for two different opportunity (auction)?

auction-server/src/server.rs
```diff
@@ -313,6 +313,9 @@ pub async fn start_server(run_options: RunOptions) -> anyhow::Result<()> {
         access_tokens: RwLock::new(access_tokens),
         metrics_recorder: setup_metrics_recorder()?,
         express_relay_idl: load_express_relay_idl()?,
```
m30m: I don't think we need this anymore

auction-server/src/auction.rs
```diff
@@ -895,15 +904,53 @@ pub fn verify_submit_bid_instruction_svm(
...
+    accounts: &[Pubkey],
+    instruction: CompiledInstruction,
+) -> Result<(u64, Pubkey), RestError> {
+    let discriminator = express_relay_svm::instruction::SubmitBid::discriminator();
+    let submit_bid_data = express_relay_svm::SubmitBidArgs::try_from_slice(
+        &instruction.data.as_slice()[discriminator.len()..],
+    )
+    .map_err(|e| RestError::BadParameters(format!("Invalid submit_bid instruction data: {}", e)))?;
+
+    if instruction.accounts.len() <= permission_account_position {
```
m30m: I think it's cleaner to use `instruction.accounts.get` instead of manual out of bound check

auction-server/src/auction.rs
```diff
@@ -895,15 +904,53 @@ pub fn verify_submit_bid_instruction_svm(
...
+            permission_account_position,
+        );
+        return Err(RestError::BadParameters(
+            "Permission account not found in submit_bid instruction".to_string(),
         ));
     }
-    Ok(())
+
+    let account_position = instruction.accounts[permission_account_position] as usize;
+    if account_position >= accounts.len() {
```
m30m, line 938: use `accounts.get` instead of manual check

auction-server/src/auction.rs
```diff
@@ -895,15 +904,53 @@ pub fn verify_submit_bid_instruction_svm(
...
+            "Permission account not found in transaction accounts: {:?} - {}",
+            accounts,
+            account_position,
+        );
+        return Err(RestError::BadParameters(
+            "Permission account not found in transaction accounts".to_string(),
+        ));
+    }
+
+    Ok((submit_bid_data.bid_amount, accounts[account_position]))
```
m30m: we should do bound check here

auction-server/build.rs
```diff
@@ -1,13 +1,23 @@
...
+};
 
-fn main() {
+fn build_evm_contracts() {
     let contract_setup = r#"
         cd ../contracts/evm
         forge build --via-ir
     "#;
     println!("cargo:rerun-if-changed=../contracts/evm");
-    println!("cargo:rerun-if-changed=migrations");
```
m30m, line 9: why is this removed? I added this because of this:
https://docs.rs/sqlx/latest/sqlx/macro.migrate.html

auction-server/src/server.rs
```diff
@@ -267,6 +268,15 @@ pub async fn start_server(run_options: RunOptions) -> anyhow::Result<()> {
     };
 
     let chains_svm = setup_chain_store_svm(config_map);
+    let express_relay_svm = ExpressRelaySvm {
+        idl:                         load_express_relay_idl()?,
```
m30m, line 272: we never use this

auction-server/src/server.rs
```diff
@@ -267,6 +268,15 @@ pub async fn start_server(run_options: RunOptions) -> anyhow::Result<()> {
...
+        permission_account_position: env!("SUBMIT_BID_PERMISSION_ACCOUNT_POSITION")
```
m30m, line 273: I'd rather put this inside the function that is using it. Soon, we can make these constant and fully compile time:
https://github.com/rust-lang/rust/pull/124941

auction-server/src/auction.rs
```diff
@@ -895,15 +906,73 @@ pub fn verify_submit_bid_instruction_svm(
...
+            instruction,
+            position,
+        );
+        return Err(RestError::BadParameters(
+            "Permission account not found in submit_bid instruction".to_string(),
         ));
     }
-    Ok(())
+
+    let account_position = instruction.accounts[position] as usize;
```
m30m, line 937: I think using `into` is safer than `as`:
`let account_position: usize = instruction.accounts[position].into();`

auction-server/src/auction.rs
```diff
@@ -895,15 +906,73 @@ pub fn verify_submit_bid_instruction_svm(
...
+        )),
+    }
+}
+
+fn extract_account_svm(
+    accounts: &[Pubkey],
+    instruction: CompiledInstruction,
+    position: usize,
+) -> Result<Pubkey, RestError> {
+    if instruction.accounts.len() <= position {
```
m30m, line 926: use `accounts.get` instead of manual boundry check

auction-server/build.rs
danimhr: It isn't removed. It's at the beginning of the `main` function

auction-server/src/server.rs
danimhr: Why not here?
If the value doesn't exist, then the run will be failed.

auction-server/build.rs
m30m: my bad

auction-server/src/server.rs
m30m: fair enough, but hopefully we can get rid of this struct later on when we can make these const values

PR #137, opened 2024-09-10
Feat/svm simulation

m30m:
This PR adds additional checks for svm bids before accepting them as valid bids:
- Simulation without signature verification on processed commitment should pass
- All signatures should be valid on the transaction with the exception of the relayer itself

### DIFF
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -269,8 +269,8 @@ fn decode_logs_for_receipt(receipt: &TransactionReceipt) -> Vec<MulticallIssuedF
         .collect()
 }
 
-
 const AUCTION_MINIMUM_LIFETIME: Duration = Duration::from_secs(1);
+
 // An auction is ready if there are any bids with a lifetime of AUCTION_MINIMUM_LIFETIME
 fn is_ready_for_auction(bids: Vec<SimulatedBid>, bid_collection_time: OffsetDateTime) -> bool {
     bids.iter()
@@ -991,6 +991,32 @@ pub async fn handle_bid_svm(
         submit_bid_instruction,
     )?;
 
+    simulate_bid_svm(chain_store, &bid).await?;
     // TODO implement this
     Err(RestError::NotImplemented)
+}
+
+async fn simulate_bid_svm(chain_store: &ChainStoreSvm, bid: &BidSvm) -> Result<(), RestError> {
+    let response = chain_store
+        .client
+        .simulate_transaction(&bid.transaction)
+        .await;
+    let result = response.map_err(|e| {
+        tracing::error!("Error while simulating bid: {:?}", e);
+        RestError::TemporarilyUnavailable
+    })?;
+    match result.value.err {
+        Some(err) => {
+            tracing::error!(
+                "Error while simulating bid: {:?}, context: {:?}",
+                err,
+                result.context
+            );
+            Err(RestError::SimulationError {
+                result: Default::default(),
+                reason: err.to_string(),
+            })
+        }
+        None => Ok(()),
+    }
 }
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -92,6 +92,7 @@ use {
     solana_sdk::{
         instruction::CompiledInstruction,
         pubkey::Pubkey,
+        signature::Signer as SvmSigner,
         transaction::VersionedTransaction,
     },
     sqlx::types::time::OffsetDateTime,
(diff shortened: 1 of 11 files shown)


auction-server/src/config.rs
```diff
@@ -62,6 +62,11 @@ pub struct RunOptions {
     #[arg(env = "SUBWALLET_PRIVATE_KEY")]
     pub subwallet_private_key: String,
```
anihamde, line 63: maybe prefix with `evm`?

auction-server/src/config.rs
```diff
@@ -62,6 +62,11 @@ pub struct RunOptions {
...
 
+    /// SVM relayer private key in base58 format
+    #[arg(long = "svm-private-key")]
+    #[arg(env = "SVM_PRIVATE_KEY")]
+    pub svm_private_key: Option<String>,
+
     #[arg(long = "secret-key")]
     #[arg(env = "SECRET_KEY")]
     pub secret_key: String,
```
anihamde, line 72: ditto, prefix with `evm`?

auction-server/src/server.rs
```diff
@@ -248,16 +257,22 @@ pub async fn start_server(run_options: RunOptions) -> anyhow::Result<()> {
...
     tracing::info!("Using wallet address: {:?}", wallet.address());
 
-    let chains = match setup_chain_store(config_map.clone(), wallet.clone()).await {
-        Ok(chain_store) => chain_store,
-        Err(err) => {
-            tracing::error!("Failed to set up chain store: {:?}", err);
-            return Err(err);
-        }
-    };
+    let chains = setup_chain_store(config_map.clone(), wallet.clone()).await?;
```
anihamde, line 260: again, maybe the naming here --> `evm` suffix

anihamde: since svm config is an option, we should probably also make evm config an option in the long run?

auction-server/src/state.rs
```diff
@@ -297,6 +302,7 @@ pub struct BidStatusWithId {
 
 #[derive(Clone)]
 pub struct ExpressRelaySvm {
```
anihamde, line 304: feels weird that this is stored separately from the `chains_svm` info. maybe it should be combined with that somehow? e.g. `chains_svm_info: SvmInfo`, with `SvmInfo` containing both the ExpressRelaySvm stuff and the hash map to specific chain information?

anihamde: realize it's not related to this PR, but just something i'm noticing now

per_sdk/svm/dummy_tx.py
```diff
@@ -3,13 +3,14 @@
 import base64
 import hashlib
 import logging
+import pathlib
```
anihamde, line 6: nit, just import `Path`

auction-server/src/server.rs
```diff
@@ -153,6 +158,10 @@ fn setup_chain_store_svm(config_map: ConfigMap) -> HashMap<ChainId, ChainStoreSv
                 chain_id.clone(),
                 ChainStoreSvm {
                     express_relay_program_id: chain_config.express_relay_program_id,
+                    client:                   RpcClient::new_with_commitment(
+                        chain_config.rpc_addr.clone(),
+                        CommitmentConfig::processed(),
```
anihamde, line 163: this client may be used for other purposes beyond simulation, maybe makes sense to not set the commitment here, and instead set it in the simulation/sending?

anihamde (approved):

auction-server/src/auction.rs
```diff
@@ -990,7 +991,49 @@ pub async fn handle_bid_svm(
...
     Err(RestError::NotImplemented)
 }
+
+fn verify_signatures_svm(bid: &BidSvm, relayer_pubkey: &Pubkey) -> Result<(), RestError> {
+    let message_bytes = bid.transaction.message.serialize();
+    let all_signatures_valid = bid
+        .transaction
+        .signatures
+        .iter()
+        .zip(bid.transaction.message.static_account_keys().iter())
```
danimhr, line 1006: are the static_account_keys always sorted as same as signatures order?

auction-server/src/config.rs
m30m: this is the admin secret key. not chain specific

auction-server/src/config.rs
m30m: we can do this in a separate refactor as Dani suggested. I didn't want to break backward compatibility here.

auction-server/src/state.rs
m30m: I guess the main reason is these fields are shared across different svm chains. Hopefully we can remove this struct soon alltogether

auction-server/src/server.rs
m30m: if we don't set it here, the default is finalized which is rarely helpful. We can assume processed is the default and manually set it if we want to use something else

auction-server/src/auction.rs
m30m: yes, I copied this logic from the `verify_with_results` function

PR #140, opened 2024-09-11
Cpi return data

anihamde:
This PR refactors the express relay program CPI to return the number of bid instructions and the fees. It uses the Anchor CPI call to return data without having to use `set_return_data` and `get_return_data`.

It also renames the `router_config` -> `config_router` everywhere for consistency.


contracts/svm/programs/express_relay/README.md
```diff
@@ -19,28 +19,34 @@ To integrate with Express Relay, the integrating program needs to make the follo
 
 ## Example Integration
 
-Integrating programs can use the `check_permission` helper method defined in the Express Relay SDK:
+Integrating programs can use the `check_permission_cpi` helper method defined in the Express Relay SDK:
```
m30m, line 22: I think we can add some snippet on how config_router and express_relay_metadata look like:

```rust
#[account(seeds = [SEED_CONFIG_ROUTER, router.key().as_ref()], bump, seeds::program = express_relay.key())]
pub config_router: UncheckedAccount<'info>,
#[account(seeds = [SEED_METADATA], bump, seeds::program = express_relay.key())]
pub express_relay_metadata: Account<'info, ExpressRelayMetadata>,
```

contracts/svm/programs/express_relay/src/utils.rs
```diff
@@ -96,13 +107,49 @@ pub fn get_matching_instructions(
...
+                        &mut &**permission_info.express_relay_metadata.try_borrow_data()?;
+                    let express_relay_metadata =
+                        ExpressRelayMetadata::try_deserialize(data_express_relay_metadata)
+                            .map_err(|_| ProgramError::InvalidAccountData)?;
+                    express_relay_metadata.split_router_default
+                }
+            };
+            for ix in matching_ixs {
+                let fee = extract_bid_from_submit_bid_ix(&ix)?;
+                total_fees += fee * split_router / FEE_SPLIT_PRECISION;
```
m30m, line 146: let's put this logic in a single shared function

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -95,8 +95,8 @@ pub mod express_relay {
         // check "no reentrancy"--submit_bid instruction only used once in transaction
         // this is done to prevent an exploit where a searcher submits a transaction with multiple submit_bid instructions with different permission keys
         // that would allow the searcher to win the right to perform the transaction if they won just one of the auctions
-        let permission_count =
-            num_permissions_in_tx(ctx.accounts.sysvar_instructions.clone(), None)?;
+        let (permission_count, _) =
+            inspect_permissions_in_tx(ctx.accounts.sysvar_instructions.clone(), None)?;
```
m30m, line 99: we can use `get_matching_instructions` here and change `inspect_permissions_in_tx` to something like `calculate_total_router_fee` which gets a list of instructions and calculates the fee.
Right now, it seems that it's doing 2 separate jobs in a single function

contracts/svm/programs/express_relay/src/utils.rs
```diff
@@ -96,13 +107,49 @@ pub fn get_matching_instructions(
     Ok(matching_instructions)
 }
 
-pub fn num_permissions_in_tx(
+/// Extracts the bid paid from a SubmitBid instruction
+pub fn extract_bid_from_submit_bid_ix(submit_bid_ix: &Instruction) -> Result<u64> {
+    let submit_bid_args =
+        SubmitBidArgs::try_from_slice(&submit_bid_ix.data[8..]).map_err(|_| {
```
danimhr, line 113: instead of 8, you can use `instruction::SubmitBid::discriminator().len()..`

contracts/svm/programs/express_relay/src/utils.rs
```diff
@@ -109,47 +109,85 @@ pub fn get_matching_instructions(
...
+pub fn perform_fee_split_router(amount: u64, split_router: u64) -> Result<u64> {
+    let fee_router = amount * split_router / FEE_SPLIT_PRECISION;
+    if fee_router > amount {
+        // this error should never be reached due to fee split checks, but kept as a matter of defensive programming
+        return err!(ErrorCode::FeesHigherThanBid);
+    }
+    Ok(fee_router)
+}
+
+pub fn perform_fee_split_relayer(amount: u64, split_relayer: u64) -> Result<u64> {
```
m30m, line 130: nit: these functions being public are dangerous since one would think you input the total bid amount and the splits and it would just work. But the amount input should be after the router fee deduction. we can make:
- make them private
- document them better
- inline their logic inside perform_fee_splits and use that in the cpi too

m30m (approved):

PR #141, opened 2024-09-11
Add submission and conclusion loops

danimhr:
This PR aims to add svm submission and conclusion loops


auction-server/src/auction.rs
```diff
@@ -1066,3 +1034,353 @@ async fn simulate_bid_svm(chain_store: &ChainStoreSvm, bid: &BidSvm) -> Result<(
         None => Ok(()),
     }
 }
+
+pub trait Auction {
```
m30m, line 1038: why not `ChainStore`?

auction-server/src/auction.rs
```diff
@@ -1066,3 +1034,353 @@ async fn simulate_bid_svm(chain_store: &ChainStoreSvm, bid: &BidSvm) -> Result<(
...
+
+    const CHAIN_TYPE: models::ChainType;
+    const AUCTION_MINIMUM_LIFETIME: Duration;
+
+    fn get_ws_client(self) -> impl Future<Output = Result<Self::WsClient>> + Send;
+    fn get_block_stream<'a>(
+        client: &'a Self::WsClient,
+    ) -> impl Future<Output = Result<Self::BlockStream<'a>>>;
+    fn filter_bids(bids: Vec<SimulatedBid>) -> Vec<Self::SimulatedBid>;
+    fn get_winner_bids(
```
m30m, line 1053: I think now that we have these trait functions, it's worth documenting what each of these functions are supposed to do

auction-server/src/auction.rs
```diff
@@ -1066,3 +1034,353 @@ async fn simulate_bid_svm(chain_store: &ChainStoreSvm, bid: &BidSvm) -> Result<(
         None => Ok(()),
     }
 }
+
+pub trait Auction {
+    type Block: DebugTrait;
+    type BlockStream<'a>: Stream<Item = Self::Block> + Unpin + Send + 'a;
+    type WsClient;
+    type SimulatedBid: SimulatedBidTrait + Clone;
```
m30m, line 1042: Do we need to specifically mention Clone? SimulatedBidTrait already includes Clone

m30m (approved):

auction-server/src/auction.rs
```diff
@@ -509,34 +507,45 @@ async fn submit_auction_for_lock(
...
+        )
+        .await
+    } else {
+        submit_auction_for_bids(
+            bids.clone(),
+            bid_collection_time,
+            permission_key.clone(),
+            chain_id.clone(),
+            store.clone(),
+            chain_store.unwrap_or_else(|| panic!("Chain not found: {}", chain_id.clone())),
```
anihamde, line 544: can you match and throw an error above instead of panicing here?

auction-server/src/auction.rs
```diff
@@ -1066,3 +1034,353 @@ async fn simulate_bid_svm(chain_store: &ChainStoreSvm, bid: &BidSvm) -> Result<(
...
+
+    while !SHOULD_EXIT.load(Ordering::Acquire) {
+        tokio::select! {
+            block = stream.next() => {
+                if block.is_none() {
+                    return Err(anyhow!("Block stream ended for chain: {}", chain_id));
+                }
+
+                tracing::debug!("New block received for {} at {}: {:?}", chain_id, OffsetDateTime::now_utc(), block);
+                store.task_tracker.spawn(
```
anihamde, line 1355: the frequency of new auctions may differ for solana vs. most evm chains. e.g. we may not want to run a solana auction so frequently to give searchers time to make bids, we may decide to run one every 2 slots. what do you think about making this function defined within the Auction trait implementations?

auction-server/src/auction.rs
```diff
@@ -1066,3 +1034,353 @@ async fn simulate_bid_svm(chain_store: &ChainStoreSvm, bid: &BidSvm) -> Result<(
...
+) -> Result<()> {
+    tracing::info!(chain_id = chain_id, "Starting transaction submitter...");
+    let mut exit_check_interval = tokio::time::interval(EXIT_CHECK_INTERVAL);
+
+    let ws_client = chain_store.get_ws_client().await?;
+    let mut stream = T::get_block_stream(&ws_client).await?;
+
+    while !SHOULD_EXIT.load(Ordering::Acquire) {
+        tokio::select! {
+            block = stream.next() => {
```
anihamde, line 1349: this is using next block with transactions as a signal that a slot has passed, but slots can be skipped or have empty blocks. it may be better to get pings when a slot has passed (even when there is no block), or to just directly query the slot and update the last slot number

auction-server/src/models.rs
```diff
@@ -53,8 +32,8 @@ pub struct Auction {
     pub conclusion_time:     Option<PrimitiveDateTime>,
     pub permission_key:      Vec<u8>,
     pub chain_id:            String,
-    #[sqlx(try_from = "Option<Vec<u8>>")]
-    pub tx_hash:             TxHash,
+    pub chain_type:          ChainType,
+    pub tx_hash:             Option<Vec<u8>>,
```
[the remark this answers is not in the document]
danimhr: I left a TODO comment and we'll handle it in the other task for BidStatus

auction-server/src/auction.rs
danimhr: We just want a new trigger to run auctions. It's not importanat what create this trigger and while the conclusion and submission being triggered for every block, we are good. Actually we don't use the block value in the process.

anihamde: yes but this trigger is that the block is Some? maybe make sure that this stream will actually return Some even if the block doesn't get created, e.g. bc the slot is skipped

danimhr: It seems that it's using this rpc method: https://solana.com/docs/rpc/http/getblock
So If the block is confimed, it can't be null.

What do you think?

danimhr: Based on this link, the notification is going to be triggered when a block is confirmed or finalized.
https://solana.com/docs/rpc/websocket/blocksubscribe

auction-server/src/auction.rs
```diff
@@ -1066,3 +1016,367 @@ async fn simulate_bid_svm(chain_store: &ChainStoreSvm, bid: &BidSvm) -> Result<(
...
+    /// The conclusion result type when try to conclude the auction for the chain
+    type ConclusionResult;
+
+    /// The chain type for the chain
+    const CHAIN_TYPE: models::ChainType;
+    /// The minimum lifetime for an auction. If any bid for auction is older than this, the auction is ready to be submitted.
+    const AUCTION_MINIMUM_LIFETIME: Duration;
+
+    /// Get the ws client for the chain
+    fn get_ws_client(self) -> impl Future<Output = Result<Self::WsClient>> + Send;
```
m30m, line 1040: can be &self

auction-server/src/auction.rs
```diff
@@ -1066,3 +1016,367 @@ async fn simulate_bid_svm(chain_store: &ChainStoreSvm, bid: &BidSvm) -> Result<(
...
+    /// Get the block stream for the ws client to subscribe to new blocks
+    fn get_block_stream<'a>(
```
m30m, line 1042: no need for explicit lifetime

auction-server/src/auction.rs
```diff
@@ -1066,3 +1016,367 @@ async fn simulate_bid_svm(chain_store: &ChainStoreSvm, bid: &BidSvm) -> Result<(
...
+        client: &'a Self::WsClient,
+    ) -> impl Future<Output = Result<Self::BlockStream<'a>>>;
+    /// Filter the bids to only include the chain specific bids
+    fn filter_bids(bids: Vec<SimulatedBid>) -> Vec<Self::SimulatedBid>;
+    /// Get the winner bids for the auction
```
m30m, line 1047: this comment doesn't add anything. perhaps we should mention it may simulate bids again, sort them by bid, etc.

auction-server/src/auction.rs
```diff
@@ -1066,3 +1016,367 @@ async fn simulate_bid_svm(chain_store: &ChainStoreSvm, bid: &BidSvm) -> Result<(
...
+    const AUCTION_MINIMUM_LIFETIME: Duration;
+
+    /// Get the ws client for the chain
+    fn get_ws_client(self) -> impl Future<Output = Result<Self::WsClient>> + Send;
+    /// Get the block stream for the ws client to subscribe to new blocks
+    fn get_block_stream<'a>(
+        client: &'a Self::WsClient,
+    ) -> impl Future<Output = Result<Self::BlockStream<'a>>>;
+    /// Filter the bids to only include the chain specific bids
+    fn filter_bids(bids: Vec<SimulatedBid>) -> Vec<Self::SimulatedBid>;
```
m30m, line 1046: I think we only need this function to convert types and we are certain that all the input bids are for that specific chain. So maybe rename it to convert_bids and panic if the input is not valid?

PR #142, opened 2024-09-12
Feat/svm submission logic

m30m:


auction-server/src/auction.rs
```diff
@@ -195,6 +216,27 @@ pub async fn submit_bids(
     Ok(tx_hash)
 }
 
+#[tracing::instrument(skip_all)]
+pub async fn submit_bids_svm(
+    relayer: &Arc<Keypair>,
+    chain_store: &ChainStoreSvm,
+    bids: Vec<SimulatedBidSvm>,
```
anihamde, line 223: shouldn't this just be a single bid?

auction-server/src/auction.rs
```diff
@@ -195,6 +216,27 @@ pub async fn submit_bids(
...
+        .transaction
+        .message
+        .static_account_keys()
+        .iter()
+        .position(|p| p.eq(&relayer.pubkey()))
+        .expect("Relayer not found in static account keys");
+    bid.transaction.signatures[relayer_signature_pos] = relayer.sign_message(&serialized_message);
+    let response = chain_store.client.send_transaction(&bid.transaction).await;
+    Ok(response.unwrap())
+    // relayer.
```
anihamde, line 237: remove

auction-server/src/auction.rs
```diff
@@ -281,6 +356,53 @@ fn is_ready_for_auction(bids: Vec<SimulatedBidEvm>, bid_collection_time: OffsetD
...
+                    index:  0,
+                    result: tx_hash,
+                },
+                Err(e) => BidStatus::Lost {
+                    index:  Some(0),
+                    result: Some(tx_hash),
+                },
+            };
+            let bids: Vec<SimulatedBid> = store.bids_for_submitted_auction(auction.clone()).await;
+            // we assume that this is just a single bid and is correct
```
anihamde, line 389: would be better to check this and error if len != 1

auction-server/src/auction.rs
```diff
@@ -502,6 +624,95 @@ async fn submit_auction_for_bids<'a>(
...
+        return Ok(());
+    }
+
+    // if !is_ready_for_auction(bids.clone(), bid_collection_time) {
+    //     tracing::info!("Auction for {} is not ready yet", permission_key);
+    //     return Ok(());
+    // }
+
+    let winner_bids =
+        get_winner_bids_svm(&bids, permission_key.clone(), store.clone(), chain_store).await?;
```
anihamde, line 652: maybe `get_winner_bids_svm` should just return a single bid for now, and `submit_bids_svm` should take in a single bid as opposed to a vector of bids

PR #144, opened 2024-09-16
Feat: svm program cli

m30m:
Simple cli program in rust to interact with the deployed program.
We can use this cli after deploying our program to mainnet for initializing and changing configs.

### DIFF
--- a/contracts/svm/cli/src/main.rs
+++ b/contracts/svm/cli/src/main.rs
@@ -1,0 +1,340 @@
+use {
+    anchor_client::{
+        anchor_lang::prelude::Pubkey,
+        solana_sdk::{
+            commitment_config::CommitmentConfig,
+            signature::{
+                read_keypair_file,
+                Keypair,
+            },
+            signer::Signer,
+            system_program,
+        },
+        Client,
+        Cluster,
+    },
+    clap::{
+        Parser,
+        Subcommand,
+    },
+    express_relay::{
+        accounts,
+        instruction,
+        state::{
+            ConfigRouter,
+            ExpressRelayMetadata,
+            SEED_CONFIG_ROUTER,
+            SEED_METADATA,
+        },
+        InitializeArgs,
+        SetRouterSplitArgs,
+        SetSplitsArgs,
+    },
+    std::str::FromStr,
+};
+
+
+/// CLI utility for interacting with Express Relay SVM program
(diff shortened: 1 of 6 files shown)


contracts/svm/cli/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let args = Args::parse();
+    let program_id = args.program_id.unwrap_or(express_relay::ID);
+    let express_relay_metadata = Pubkey::find_program_address(&[SEED_METADATA], &program_id).0;
+    let cluster = Cluster::from_str(args.rpc_url.as_str())?;
+
+    match args.command {
+        Commands::Initialize(init_args) => {
+            let payer = read_keypair_file(init_args.payer)?;
```
danimhr, line 162: Not sure about this, but you can implement a function like:
```
fn send_transaction(cluster: Cluster, payer_file: String, program_id: Pubkey, accounts: impl ToAccountMetas, args: impl InstructionData) -> Result<Signature, Box<dyn std::error::Error>> {
    let payer = read_keypair_file(payer_file)?;
    let client = Client::new_with_options(cluster, &payer, CommitmentConfig::confirmed());
    let program = client.program(program_id)?;
    let result = program
        .request()
        .accounts(accounts)
        .args(args)
        .signer(&payer)
        .send()?;
    Ok(result)
}
```

and then extract the result with:

```
let result = send_transaction(
                cluster,
                init_args.payer,
                program_id,
                accounts::Initialize {
                    payer: Pubkey::default(),
                    express_relay_metadata,
                    admin: Pubkey::default(),
                    relayer_signer: Pubkey::default(),
                    fee_receiver_relayer: Pubkey::default(),
                    system_program: system_program::ID,
                },
                instruction::Initialize {
                    data: InitializeArgs {
                        split_router_default: init_args.split_router_default,
                        split_relayer:        init_args.split_relayer,
                    },
                },
            )?;
```

I think this way we can get rid of bunch of duplicate codes :-?

danimhr (approved):

contracts/svm/cli/src/main.rs
m30m: the issue is we still need the payer pubkey when creating the `accounts::Initialize` object.
I tried to make it a bit simpler

PR #148, opened 2024-09-18
Add svm bid status

danimhr:
This PR aims to add svm bid status type


auction-server/src/state.rs
```diff
@@ -885,9 +1189,17 @@ impl Store {
         match auction.tx_hash {
             Some(tx_hash) => bids
                 .into_iter()
-                .filter(|bid| match bid.get_core_fields().status {
-                    BidStatus::Submitted { result, .. } => result == tx_hash,
-                    _ => false,
+                .filter(|bid| match bid {
+                    SimulatedBid::Evm(bid) => match bid.core_fields.status {
+                        BidStatusEvm::Submitted { result, .. } => {
+                            result.as_bytes().to_vec() == tx_hash
```
m30m, line 1195: use convert_tx_hash

auction-server/src/state.rs
```diff
@@ -322,50 +503,232 @@ pub enum BidStatus {
...
+}
+
+pub trait BidStatusTrait:
+    Clone
+    + Into<models::BidStatus>
+    + std::fmt::Debug
+    + TryFrom<(models::Bid, Option<models::Auction>)>
+    + Into<BidStatus>
+{
+    type TxHash: Clone + std::fmt::Debug;
```
m30m, line 590: you can do something like this:
```
type TxHash: Clone + std::fmt::Debug + AsRef<[u8]>;
    fn convert_tx_hash(tx_hash: &Self::TxHash) -> Vec<u8> {
        tx_hash.as_ref().to_vec()
    }
```

auction-server/src/state.rs
```diff
@@ -434,32 +789,15 @@ impl SimulatedBidCoreFields {
...
-            SimulatedBid::Evm(bid) => SimulatedBid::Evm(SimulatedBidEvm { core_fields, ..bid }),
-            SimulatedBid::Svm(bid) => SimulatedBid::Svm(SimulatedBidSvm { core_fields, ..bid }),
+            SimulatedBid::Evm(bid) => bid.core_fields.id,
+            SimulatedBid::Svm(bid) => bid.core_fields.id,
         }
     }
 }
 
-impl TryFrom<(models::Bid, Option<models::Auction>)> for BidStatus {
+impl TryFrom<(models::Bid, Option<models::Auction>)> for BidStatusEvm {
```
m30m, line 800: I'd rather we use From and TryFrom traits only for conversions that are simple.
We can put these in dedicated functions inside the traits.

it's not easy to discover `let a: BidStatusEvm = (bid, auction).try_into()` is doable

auction-server/src/auction.rs
```diff
@@ -379,18 +388,22 @@ async fn broadcast_lost_bids<T: SimulatedBidTrait>(
...
-                    BidStatus::Lost {
-                        result: tx_hash,
-                        index:  None,
-                    },
-                    auction,
-                )
-                .await
-            {
-                tracing::error!("Failed to broadcast bid status: {:?} - bid: {:?}", err, bid);
+            match T::get_bid_status(models::BidStatus::Submitted, None, tx_hash) {
```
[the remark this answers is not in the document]
danimhr: Wow! good catch

auction-server/src/state.rs
danimhr: It's better to put them inside a function.
The other reason for this is you cannot get the try_from error type when implementing a try_from for a trait.

m30m (approved):

PR #153, opened 2024-09-23
refactor: Put chain specific configs in ChainStores

m30m:


danimhr (approved):
There is a submit_bids function in auction.rs which can be deleted as well :-?

PR #154, opened 2024-09-24
Refactor/opportunity module

danimhr:
This PR aims to add opportunity module with new architecture (draft) to auction server


auction-server/src/opportunity/service/make_permitted_tokens.rs
```diff
@@ -0,0 +1,53 @@
...
+    ethers::types::U256,
+};
+
+pub struct MakePermittedTokensInput {
+    pub opportunity:     OpportunityEvm,
+    pub opportunity_bid: OpportunityBid,
+}
+
+impl Service<ChainTypeEvm> {
+    pub fn make_permitted_tokens(
```
m30m, line 23: I think this separation is a bit too extreme. `make_permitted_tokens` is just some internal function that is used by another function. We shouldn't expose it publicly and I'd rather have this next to the caller. We can discuss this but maybe each file in this folder should include the logic for a complete functionality this service exposes to outside.

auction-server/src/opportunity/service/handle_opportunity_bid.rs
```diff
@@ -0,0 +1,123 @@
...
+    let erc20_decoded = erc20::ERC20Errors::decode_with_selector(revert)
+        .map(|decoded_error| format!("ERC20 Contract Revert Error: {:#?}", decoded_error));
+    apdapter_decoded.or(erc20_decoded)
+}
+
+impl Service<ChainTypeEvm> {
+    pub async fn handle_opportunity_bid(
+        &self,
+        input: HandleOpportunityBidInput,
+    ) -> anyhow::Result<Uuid, RestError> {
```
m30m, line 59: Let's use the same syntax everywhere:  `anyhow::Result<Uuid, RestError>` -> `Result<Uuid, RestError>`

auction-server/src/state.rs
```diff
@@ -845,16 +816,22 @@ pub struct ExpressRelaySvm {
...
+    pub task_tracker:     TaskTracker,
+    pub secret_key:       String,
+    pub access_tokens:    RwLock<HashMap<models::AccessTokenToken, models::Profile>>,
+    pub metrics_recorder: PrometheusHandle,
+}
+
+#[derive(Clone)]
+pub struct StoreNew {
+    pub opportunity_service_evm:
+        Arc<opportunity_service::Service<opportunity_service::ChainTypeEvm>>,
```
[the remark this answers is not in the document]
danimhr: We need to call sth in the auction module, so we need to access the store itself within the service.
When we refactor other parts, this will be fixed and we'll have only a single store with services in it.
So you are right, this isn't sth that I like to have in the final design, however we need it for now to take care of this call

auction-server/src/opportunity/service/get_spoof_info.rs
```diff
@@ -0,0 +1,54 @@
...
+
+impl Service<ChainTypeEvm> {
+    /// Find the spoof info for an ERC20 token. This includes the balance slot and the allowance slot
+    /// Returns an error if no balance or allowance slot is found
+    /// # Arguments
+    ///
+    /// * `token`: ERC20 token address
+    /// * `client`: Client to interact with the blockchain
+    #[tracing::instrument(skip_all, fields(token=%input.token))]
+    pub async fn get_spoof_info(&self, input: GetSpoofInfoInput) -> Result<SpoofInfo, RestError> {
```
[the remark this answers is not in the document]
danimhr: Using `RestError` is temporary and we should create specific errors for each module later.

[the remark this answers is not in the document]
danimhr: IMO everything in `Service` layer can be exposed

auction-server/src/opportunity/repository/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+mod models;
+mod opportunity_exists;
+mod remove_opportunity;
+
+#[derive(Debug)]
+pub struct Repository<T: Cache> {
+    pub cache: T,
+}
+
+pub trait Cache: Deref<Target = CacheCoreFields<Self::Opportunity>> {
```
m30m, line 27: Maybe rename to `InMemoryStore`? Cache misrepresents all the logic we have put inside this

auction-server/src/opportunity/service/get_opportunities.rs
```diff
@@ -0,0 +1,70 @@
...
+                    .await
+                    .iter()
+                    .map(|(_key, opportunities)| {
+                        let opportunity = opportunities
+                            .last()
+                            .expect("A permission key vector should have at least one opportunity");
+                        opportunity.clone()
+                    })
+                    .filter(|opportunity| {
+                        // let OpportunityParams::V1(params) = &params_with_id.params;
```
m30m, line 46: remove

auction-server/src/opportunity/service/make_permitted_tokens.rs
danimhr: We should make private some of the services functionalities

m30m (approved):

PR #155, opened 2024-09-25
Fix svm auction loop

danimhr:
This PR aims to fix auction loop submission for svm chains

### DIFF
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -1488,7 +1488,7 @@ async fn run_submission_loop<T: ChainStore>(
                             }
                             models::ChainType::Svm => {
                                 if let Some(chain_store) = store
-                                    .chains
+                                    .chains_svm
                                     .get(&chain_id) {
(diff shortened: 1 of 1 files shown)


auction-server/src/auction.rs
```diff
@@ -1488,7 +1488,7 @@ async fn run_submission_loop<T: ChainStore>(
                             }
                             models::ChainType::Svm => {
                                 if let Some(chain_store) = store
-                                    .chains
+                                    .chains_svm
```
m30m, line 1491: I think this is one of the places where panic is more appropriate. Otherwise, we will discover it in prod :D

m30m (approved):

auction-server/src/auction.rs
danimhr: We should find a way to make it compile time :D

m30m: `submit_auctions::<ChainStoreEvm>(store.clone(), chain_store, chain_id).await`
:D

PR #156, opened 2024-09-26
Feat/svm opportunity

danimhr:
This PR aims to add opportunity handles for svm


auction-server/src/opportunity/api.rs
```diff
@@ -102,21 +107,105 @@ pub struct OpportunityParamsV1 {
...
+    /// Token amount
+    #[schema(example = 1000)]
+    pub amount: u64,
+}
+
+// TODO descibe it
+#[serde_as]
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug)]
+pub struct OpportunityParamsV1KaminoSvm {
+    #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
```
m30m, line 144: I think the only field specific to kamino is the order and these can live on a higher level struct

auction-server/src/opportunity/api.rs
```diff
@@ -102,21 +107,105 @@ pub struct OpportunityParamsV1 {
...
+#[serde(tag = "version")]
+pub enum OpportunityParamsSvm {
     #[serde(rename = "v1")]
-    V1(OpportunityParamsV1),
+    V1(OpportunityParamsV1Svm),
+}
+
+/// Similar to OpportunityParams, but with the opportunity id included.
+#[derive(Serialize, Deserialize, ToSchema, Clone, ToResponse)]
+pub struct OpportunityParamsWithMetadataEvm {
```
m30m, line 173: can this become a generic? Since it's getting exposed to apis and axum through two concrete types from the OpportunityParamsWithMetadata enum, maybe it will work?

auction-server/src/opportunity/repository/models.rs
```diff
@@ -22,25 +34,55 @@ pub enum OpportunityRemovalReason {
     Invalid,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct OpportunityMetadataEvm {
```
m30m, line 38: On the api layer what we mean by metadata is opportunity_id, creation time and anything not representing the core opportunity. I think the meaning is reversed here. we should probably use a different word or maybe even nothing?

danimhr: Names in api layer are updated. WDYT?

auction-server/src/opportunity/api.rs
danimhr: I change types a bit, but didnt make them generic as we need these types for open api generation.

auction-server/src/opportunity/repository/models.rs
```diff
@@ -22,48 +37,67 @@ pub enum OpportunityRemovalReason {
...
+    fn get_chain_type() -> ChainType {
+        ChainType::Svm
     }
 }
+
+#[derive(Clone, FromRow, Debug)]
+pub struct Opportunity<T: OpportunityMetadata> {
+    pub id:             Uuid,
+    pub creation_time:  PrimitiveDateTime,
+    pub permission_key: Vec<u8>,
```
anihamde, line 95: the permission seems to exist in `OpportunityMetadataSvm`, it should probably only exist in 1 place

auction-server/src/opportunity/repository/models.rs
anihamde: i agree that metadata seems like a weird term to use for the core info about the opportunity; if anything, this is actually the "complement" of the metadata. i lol'd at [this](https://www.powerthesaurus.org/metadata), but maybe `OpportunityContent` would actually be a better term than metadata

auction-server/src/opportunity/service/verification.rs
```diff
@@ -279,3 +285,16 @@ impl Service<ChainTypeEvm> {
...
+impl Verification<ChainTypeSvm> for Service<ChainTypeSvm> {
+    async fn verify_opportunity(
+        &self,
+        input: VerifyOpportunityInput<entities::OpportunityCreateSvm>,
+    ) -> Result<entities::OpportunityVerificationResult, RestError> {
+        self.get_config(&input.opportunity.core_fields.chain_id)?;
+
+        // To make sure it'll be expired after a minute
+        // TODO - change this to a more realistic value
+        Ok(entities::OpportunityVerificationResult::UnableToSpoof)
```
anihamde, line 298: maybe `UnableToSpoof` should have a u256 param representing number of seconds this opp can be present? Can set then set this to different values, don't need to check staleness agaisnt `MAX_STALE_OPPORTUNITY_MICROS`

auction-server/src/opportunity/api.rs
```diff
@@ -102,148 +183,189 @@ pub struct OpportunityParamsV1 {
...
+    /// The opportunity unique id
+    #[schema(example = "obo3ee3e-58cc-4372-a567-0e02b2c3d479", value_type = String)]
+    pub opportunity_id: OpportunityId,
+    /// Creation time of the opportunity (in microseconds since the Unix epoch)
+    #[schema(example = 1_700_000_000_000_000i128, value_type = i128)]
+    pub creation_time:  UnixTimestampMicros,
+    #[serde(flatten)]
+    // expands params into component fields in the generated client schemas
+    #[schema(inline)]
+    pub params:         OpportunityCreateEvm,
```
anihamde, line 208: I think `OpportunityParamsEvm` is a more descriptive type name here? since this `Opportunity...` object may be accessed outside of just creation?

auction-server/src/opportunity/api.rs
```diff
@@ -102,148 +183,189 @@ pub struct OpportunityParamsV1 {
...
+    #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
+    #[serde_as(as = "DisplayFromStr")]
+    pub block_hash: Hash,
+
+    pub sell_tokens: Vec<TokenAmountSvm>,
+    pub buy_tokens:  Vec<TokenAmountSvm>,
+
+    #[serde(flatten)]
+    #[schema(inline)]
+    pub client_params: OpportunityCreateClientParamsV1Svm,
```
anihamde, line 270: maybe `program_params`? client feels a bit overused as a term, this field is for program-specific (e.g. `limo`) content

auction-server/src/opportunity/entities/opportunity_evm.rs
```diff
@@ -34,3 +69,158 @@ impl Deref for OpportunityEvm {
...
+            target_calldata:   val.target_calldata.clone(),
+        }
+    }
+}
+
+impl From<OpportunityEvm> for OpportunityCreateEvm {
+    fn from(val: OpportunityEvm) -> Self {
+        OpportunityCreateEvm {
+            core_fields:       OpportunityCoreFieldsCreate::<TokenAmountEvm> {
+                permission_key: val.core_fields.permission_key.clone(),
```
m30m, line 204: no cloning needed in this function

auction-server/src/opportunity/entities/opportunity_evm.rs
```diff
@@ -34,3 +69,158 @@ impl Deref for OpportunityEvm {
...
+                buy_tokens:     val.core_fields.buy_tokens.clone(),
+            },
+            target_contract:   val.target_contract,
+            target_call_value: val.target_call_value,
+            target_calldata:   val.target_calldata.clone(),
+        }
+    }
+}
+
+impl PartialEq<OpportunityCreateEvm> for OpportunityEvm {
```
m30m, line 216: with the conversions we have, we don't need this. we can convert `OpportunityEvm` into `OpportunityCreateEvm` and then compare

auction-server/src/opportunity/entities/opportunity_evm.rs
```diff
@@ -34,3 +69,158 @@ impl Deref for OpportunityEvm {
...
+}
+
+impl From<OpportunityCreateEvm> for OpportunityEvm {
+    fn from(val: OpportunityCreateEvm) -> Self {
+        let id = Uuid::new_v4();
+        let odt = OffsetDateTime::now_utc();
+        OpportunityEvm {
+            core_fields:       OpportunityCoreFields::<TokenAmountEvm> {
+                id,
+                creation_time: odt.unix_timestamp_nanos() / 1000 as UnixTimestampMicros,
```
m30m, line 187: no cloning needed in this function

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -28,3 +78,178 @@ impl Deref for OpportunitySvm {
...
+            },
+            router:      val.router,
+            permission:  val.permission,
+            block_hash:  val.block_hash,
+            client:      val.client,
+        }
+    }
+}
+
+impl PartialEq<OpportunityCreateSvm> for OpportunitySvm {
```
m30m, line 244: no need, we can just convert first and then compare

auction-server/src/opportunity/repository/get_opportunities_by_permission_key.rs
```diff
@@ -49,17 +59,19 @@ impl Repository<InMemoryStoreEvm> {
...
-                "Failed to convert opportunity to OpportunityParamsWithMetadata: {} - chain_id: {:?} - permission_key: {:?} - from_time: {:?}",
-                e,
-                chain_id,
-                permission_key,
-                from_time,
-            );
-            RestError::TemporarilyUnavailable
-        })
+
+        let opportunities: Result<Vec<T::Opportunity>, RestError> = opps.clone().into_iter().map(|opp| opp.clone().try_into().map_err(
```
m30m, line 63: no need to clone `opps`

auction-server/src/opportunity/repository/models.rs
m30m: Yeah, I still think these fields are not metadata

auction-server/src/opportunity/service/verification.rs
danimhr: I'm not sure if expiring a opportunity is the verification responsibility or not :-?

auction-server/src/opportunity/repository/models.rs
danimhr: I will update it in a separate PR

auction-server/src/opportunity/api.rs
```diff
@@ -278,76 +400,117 @@ pub async fn opportunity_bid(
...
+    let opportunities_svm = store
+        .opportunity_service_svm
         .get_opportunities(GetOpportunitiesInput {
             query_params: query_params.0,
         })
-        .await?;
-    Ok(Json(opportunities.into_iter().map(|o| o.into()).collect()))
+        .await;
+
+    if opportunities_evm.is_err() && opportunities_svm.is_err() {
```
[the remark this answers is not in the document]
danimhr: As we may later have an api for returning all opps, it's better to keep it how it is now

auction-server/src/opportunity/entities/opportunity_evm.rs
```diff
@@ -23,8 +34,44 @@ pub struct OpportunityEvm {
...
+    pub target_calldata:   Bytes,
+    pub target_call_value: U256,
+}
+
 impl Opportunity for OpportunityEvm {
     type TokenAmount = TokenAmountEvm;
+    type ModelMetadata = repository::OpportunityMetadataEvm;
+    type OpportunityCreate = OpportunityCreateEvm;
+
+    fn new_with_current_time(val: Self::OpportunityCreate) -> Self {
```
m30m, line 51: nit: we can say `fn new_with_current_time(val: OpportunityCreateEvm) -> Self` to avoid one level of indirection

m30m (approved):

PR #157, opened 2024-09-27
feat: Add limo program in tilt

m30m:
Tested it with the js sdk and had some successful bids locally

### DIFF
--- a/contracts/svm/package.json
+++ b/contracts/svm/package.json
@@ -7,13 +7,14 @@
     "@coral-xyz/anchor": "^0.30.1"
   },
   "devDependencies": {
-    "chai": "^4.3.4",
-    "mocha": "^9.0.3",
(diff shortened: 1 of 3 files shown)


contracts/svm/package.json
```diff
@@ -7,13 +7,14 @@
     "@coral-xyz/anchor": "^0.30.1"
   },
   "devDependencies": {
-    "chai": "^4.3.4",
-    "mocha": "^9.0.3",
-    "ts-mocha": "^10.0.0",
+    "@kamino-finance/limo-sdk": "^0.2.3",
```
danimhr, line 10: is it compatible with our anchor version. Not sure if it matters btw :D

danimhr (approved):

contracts/svm/package.json
m30m: I don't think it does

PR #159, opened 2024-10-01
SVM Static Analysis Improvements

anihamde:
Ran static analysis on SVM Express Relay program and incorporated some changes:
- added comments on validation of `config_router` in helper method
- added comment on rationale for `permission` as account instead of instruction data
- fixed underflow/overflow sources

### DIFF
--- a/contracts/svm/programs/express_relay/src/lib.rs
+++ b/contracts/svm/programs/express_relay/src/lib.rs
@@ -252,7 +252,7 @@ pub struct SubmitBid<'info> {
 
     pub relayer_signer: Signer<'info>,
 
-    /// CHECK: this is the permission_key. This is included as an account to save transaction size via caching in case of repeated use; account contestation is not a concern because this account should never be written to.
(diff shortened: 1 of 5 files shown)


contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -252,7 +252,7 @@ pub struct SubmitBid<'info> {
 
     pub relayer_signer: Signer<'info>,
 
-    /// CHECK: this is the permission_key
+    /// CHECK: this is the permission_key. This is included as an account to save transaction size via caching in case of repeated use; account contestation is not a concern because this account should never be written to.
```
m30m, line 255: > because this account should never be written to

This account can be mutable in the same transaction via the other programs using it, so this part may be a bit misleading (although this program never changes it).
Maybe we can say something like:
... to  save transaction size. Usually the permission_key can be described as the account storing the opportunity information, so we just reuse that account to save 32 bytes instead of putting this in instruction arguments.

m30m (approved):

contracts/svm/programs/express_relay/src/lib.rs
anihamde: you're right that other programs can set it to mutable, but that's generally true for all accounts (except programs). The comment here expresses that this account's data _should never_ be mutated; I'll make that clearer

PR #160, opened 2024-10-02
Fix open api schema

danimhr:
This PR aims to fix open api json schema generation


auction-server/src/opportunity/api.rs
```diff
@@ -60,6 +60,11 @@ use {
     std::sync::Arc,
     time::OffsetDateTime,
     utoipa::{
+        openapi::{
```
m30m, line 63: I don't think these are necessary

m30m (approved):

PR #161, opened 2024-10-07
feat: Add limo order address

m30m:
SDK requires the order address for filling them

### DIFF
--- a/auction-server/src/opportunity/repository/models.rs
+++ b/auction-server/src/opportunity/repository/models.rs
@@ -50,7 +50,9 @@ pub struct OpportunityMetadataEvm {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct OpportunityMetadataSvmProgramLimo {
     #[serde_as(as = "Base64")]
-    pub order: Vec<u8>,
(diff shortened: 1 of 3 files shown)


auction-server/src/opportunity/repository/models.rs
```diff
@@ -50,7 +50,9 @@ pub struct OpportunityMetadataEvm {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct OpportunityMetadataSvmProgramLimo {
     #[serde_as(as = "Base64")]
-    pub order: Vec<u8>,
+    pub order:         Vec<u8>,
+    #[serde_as(as = "DisplayFromStr")]
+    pub order_address: Pubkey,
```
danimhr, line 55: What will happen to old opps? Maybe create a migration to remove old svm opps?
If there is no opps yet, we are good to go

danimhr (approved):

danimhr:
@m30m Please update the version too

auction-server/src/opportunity/repository/models.rs
m30m: no opps on per-staging

PR #163, opened 2024-10-07
Add tracing for svm rpc client

danimhr:
This PR aims to add tracing for svm rpc client

### DIFF
--- a/auction-server/Cargo.toml
+++ b/auction-server/Cargo.toml
@@ -49,6 +49,7 @@ bincode = "1.3.3"
 serde_with = { version = "3.9.0", features = ["base64"] }
 anchor-lang = "0.30.1"
 express-relay = { path = "../contracts/svm/programs/express_relay" }
+solana-rpc-client = "2.0.13"
 
 # The curve25519-dalek crate is a dependency of solana-sdk.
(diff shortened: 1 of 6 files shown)


auction-server/Cargo.toml
```diff
@@ -49,6 +49,7 @@ bincode = "1.3.3"
 serde_with = { version = "3.9.0", features = ["base64"] }
 anchor-lang = "0.30.1"
 express-relay = { path = "../contracts/svm/programs/express_relay" }
+solana-rpc-client = "2.0.13"
```
m30m, line 52: I don't think we need this explicitly since `solana-client` already exposes what we want:
`solana_client::rpc_sender::RpcSender`

auction-server/src/traced_sender_svm.rs
```diff
@@ -0,0 +1,75 @@
...
+        },
+    },
+    solana_rpc_client::http_sender::HttpSender,
+    std::time::{
+        Duration,
+        Instant,
+    },
+};
+
+pub struct TracedSenderSvm {
```
m30m, line 22: I think `TracedClientSvm` is a better name since we do all RPC interactions through this and it's not just for sending.

m30m (approved):

PR #164, opened 2024-10-09
feat: Limo opp updater

m30m:
ToDo:

- [x] Integrate this with tilt

### DIFF
--- a/create_limo_profile.py
+++ b/create_limo_profile.py
@@ -1,0 +1,21 @@
+"""
+This script is used to generate a new api key for the limo profile for the limonade service.
+"""
+
+import requests
+import random
+
+def main():
+    headers = {'Authorization':'Bearer admin'}
+    random_id = random.randint(0, 10000000)
+    response = requests.post('http://localhost:9000/v1/profiles', json={'name': 'limo', 'email':f'limo{random_id}@dourolabs.com','role':'protocol'}, headers=headers)
+    profile_id = response.json()['id']
+
+    response = requests.post('http://localhost:9000/v1/profiles/access_tokens', json={'profile_id':profile_id}, headers=headers)
+    access_token = response.json()['token']
+    print(access_token)
+
+
+
+if __name__ == '__main__':
+    main()
--- a/create_limo_profile.py
+++ b/create_limo_profile.py
@@ -7,8 +7,10 @@ import random
 
 def main():
     headers = {'Authorization':'Bearer admin'}
-    random_id = random.randint(0, 10000000)
-    response = requests.post('http://localhost:9000/v1/profiles', json={'name': 'limo', 'email':f'limo{random_id}@dourolabs.com','role':'protocol'}, headers=headers)
+    limo_email='limo@dourolabs.com'
+    response = requests.post('http://localhost:9000/v1/profiles', json={'name': 'limo', 'email':limo_email,'role':'protocol'}, headers=headers)
+    if response.status_code == 400:
+        response = requests.get('http://localhost:9000/v1/profiles', headers=headers, params={'email':limo_email})
     profile_id = response.json()['id']
 
     response = requests.post('http://localhost:9000/v1/profiles/access_tokens', json={'profile_id':profile_id}, headers=headers)
--- a/scripts/limonade/Dockerfile
+++ b/scripts/limonade/Dockerfile
@@ -1,0 +1,65 @@
+# syntax=docker/dockerfile:1
+
+# Comments are provided throughout this file to help you get started.
+# If you need more help, visit the Dockerfile reference guide at
+# https://docs.docker.com/go/dockerfile-reference/
+
+# Want to help us make this template better? Share your feedback here: https://forms.gle/ybq9Krt8jtBL3iCk7
+
+ARG NODE_VERSION=18.20.2
+
+################################################################################
+# Use node image for base image for all stages.
+FROM node:${NODE_VERSION}-alpine as base
+
+# Set working directory for all build stages.
+WORKDIR /usr/src/app
+
+
+################################################################################
+# Create a stage for installing production dependecies.
+FROM base as deps
+
+# Download dependencies as a separate step to take advantage of Docker's caching.
+# Leverage a cache mount to /root/.npm to speed up subsequent builds.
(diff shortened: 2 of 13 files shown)


scripts/limonade/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+  let { blockhash: latestBlockhash } = await connection.getLatestBlockhash(
+    "confirmed"
+  );
+  const response = await connection.getProgramAccounts(limoId, {
+    commitment: "confirmed",
+    filters,
+    withContext: true,
+  });
+  const payloads = [];
+  for (const account of response.value) {
```
danimhr, line 75: Use map and filter here

scripts/limonade/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+      program: "limo",
+      blockHash: latestBlockhash,
+      chainId: argv.chainId,
+      slot: response.context.slot,
+      order: { state: order, address: account.pubkey },
+    };
+    payloads.push(payload);
+  }
+  console.log("Initial opportunities", payloads.length);
+  for (const payload of payloads) {
```
danimhr, line 90: make it concurrent with promise.all

danimhr (commented):
I think we need dockerfile for the deployment

scripts/limonade/src/index.ts
m30m: I actually don't want to submit all of these in parallel. Assuming there are a lot of orders that are still valid, it's gonna add a lot of unnecessary load to the server if we do it in parallel

scripts/limonade/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    default: "https://pyth-express-relay-mainnet.asymmetric.re/",
+  })
+  .option("rpc-endpoint", {
+    description:
+      "Solana rpc endpoint to use for subscribing to program account changes.",
+    type: "string",
+    default: "http://127.0.0.1:8899",
+  })
+  .option("global-config", {
+    description: "Global config of limo program to filter for.",
```
anihamde, line 26: nit: "Pubkey of the global config..."

scripts/limonade/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  ];
+
+  console.log("Listening for program account changes");
+  const client = new Client({ baseUrl: argv.endpoint, apiKey: argv.apiKey });
+  let { blockhash: latestBlockhash } = await connection.getLatestBlockhash(
+    "confirmed"
+  );
+  const submitExistingOpportunities = async () => {
+    const response = await connection.getProgramAccounts(limoId, {
+      commitment: "confirmed",
```
anihamde, line 71: food for thought, use "processed"? a low percentage of processed blocks don't get confirmed ([supposedly ~5%](https://solana.com/docs/advanced/confirmation)), and this could improve latency of publishing opportunities if that's a source of delay. ofc there's a tradeoff, but in worst case the searcher's trade attempt fails on-chain.

anihamde (approved):

create_limo_profile.py
```diff
@@ -0,0 +1,21 @@
...
+This script is used to generate a new api key for the limo profile for the limonade service.
+"""
+
+import requests
+import random
+
+def main():
+    headers = {'Authorization':'Bearer admin'}
+    random_id = random.randint(0, 10000000)
+    response = requests.post('http://localhost:9000/v1/profiles', json={'name': 'limo', 'email':f'limo{random_id}@dourolabs.com','role':'protocol'}, headers=headers)
```
danimhr, line 11: What will happen if we create the same profile multiple times? I think it'll raise an error which says email already exists. Should we catch this here? I think it's not gonna raise error if we create access_token for the profile twice.

scripts/limonade/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+FROM base as final
+
+# Use production node environment by default.
+ENV NODE_ENV production
+
+# Run the application as a non-root user.
+USER node
+
+# Copy package.json so that package manager commands can be used.
+COPY package.json .
```
danimhr, line 60: Should we copy .lock file as well?

scripts/limonade/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+        },
+      }))
+      .filter(
+        (opportunityCreate) =>
+          opportunityCreate.order.state.remainingInputAmount.toNumber() !== 0
+      );
+
+    console.log("Initial opportunities", payloads.length);
+    for (const payload of payloads) {
+      await client.submitOpportunity(payload);
```
danimhr, line 93: I think we don't need to await here :-?
Maybe just run sth like this:
```
payloads.forEach(payload => {
   client.submitOpportunity(payload).then((data) => console.log(`submitted opportunity successfully for ${data}`)).catch(err => console.log("failed to submit opportunity: ", err))
}
```
Or sth similar :-?

scripts/limonade/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    filters
+  );
+  const updateLatestBlockhash = async () => {
+    while (true) {
+      latestBlockhash = (await connection.getLatestBlockhash("confirmed"))
+        .blockhash;
+      await new Promise((resolve) => setTimeout(resolve, 10000));
+    }
+  };
+  await Promise.all([updateLatestBlockhash(), submitExistingOpportunities()]);
```
danimhr, line 127: You can use the then, catch pattern here for the `submitExistingOpportunities` as well.

scripts/limonade/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    description: "Global config of limo program to filter for.",
+    type: "string",
+    demandOption: true,
+  })
+  .option("chain-id", {
+    description: "Chain id to publish these orders for as opportunity",
+    type: "string",
+    demandOption: true,
+  })
+  .option("api-key", {
```
danimhr, line 35: This should be required

scripts/limonade/src/index.ts
anihamde: sorry this comment is more relevant for the filter in `onProgramAccountChange`

PR #166, opened 2024-10-09
Fix openapi docs for redoc and remove swagger

danimhr:
This PR aims to fix auction server docs issues. Additionally swagger docs is replaced with redoc as it has better ui.

### DIFF
--- a/auction-server/src/api.rs
+++ b/auction-server/src/api.rs
@@ -21,7 +21,7 @@ use {
         },
         config::RunOptions,
         models,
-        opportunity::api as opportunity_api,
+        opportunity::api as opportunity,
         server::{
             EXIT_CHECK_INTERVAL,
             SHOULD_EXIT,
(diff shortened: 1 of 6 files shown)


auction-server/src/api.rs
```diff
@@ -267,6 +266,17 @@ macro_rules! login_required {
     };
 }
 
+fn remove_discriminators(mut doc: serde_json::Value) -> serde_json::Value {
```
m30m, line 269: nit: I guess you can do this inplace:

```rs
fn remove_discriminators(doc: &mut serde_json::Value) {
    // Recursively remove all "discriminator" fields from the OpenAPI document
    if let Some(obj) = doc.as_object_mut() {
        obj.retain(|key, _| key != "discriminator");
        for value in obj.values_mut() {
            remove_discriminators(value);
        }
    }
}
```

m30m (approved):

PR #175, opened 2024-10-17
v1 program

anihamde:
changes to program to support v1

### DIFF
--- a/contracts/svm/programs/express_relay/src/lib.rs
+++ b/contracts/svm/programs/express_relay/src/lib.rs
@@ -5,7 +5,10 @@ pub mod utils;
 
 use {
     crate::{
+        cpi::accounts::CheckPermission as CheckPermissionCPI,
         error::ErrorCode,
+        program::ExpressRelay,
+        sdk::cpi::check_permission_cpi,
         state::*,
         utils::*,
     },
@@ -19,6 +22,11 @@ use {
             sysvar::instructions as sysvar_instructions,
         },
         system_program::System,
+    },
+    anchor_spl::token::{
+        Mint,
+        Token,
+        TokenAccount,
     },
 };
 
@@ -146,6 +154,63 @@ pub mod express_relay {
             &fee_receiver_admin.to_account_info(),
             amount,
         )
+    }
+
+    pub fn swap(ctx: Context<Swap>, data: SwapArgs) -> Result<()> {
+        // check permission
+        let check_permission_accounts = CheckPermissionCPI {
+            sysvar_instructions:    ctx.accounts.sysvar_instructions.to_account_info(),
+            permission:             ctx.accounts.permission.to_account_info(),
+            router:                 ctx.accounts.router.to_account_info(),
+            config_router:          ctx.accounts.config_router.to_account_info(),
+            express_relay_metadata: ctx.accounts.express_relay_metadata.to_account_info(),
+        };
+        check_permission_cpi(
+            check_permission_accounts,
+            ctx.accounts.express_relay_program.to_account_info(),
+        )?;
+
+        let fees_referral = data
+            .referral_fee_ppm
+            .checked_mul(data.amount_input)
+            .ok_or(ProgramError::ArithmeticOverflow)?
+            / 1_000_000;
+
+        if fees_referral > 0 {
+            transfer_spl(
+                &ctx.accounts.ta_input_searcher.to_account_info(),
+                &ctx.accounts.ta_input_router.to_account_info(),
+                &ctx.accounts.token_program.to_account_info(),
+                &ctx.accounts.router.to_account_info(),
+                fees_referral,
+            )?;
+        }
+
+        let amount_to_trader = data
+            .amount_input
+            .checked_sub(fees_referral)
+            .ok_or(ProgramError::ArithmeticOverflow)?;
+        if amount_to_trader > 0 {
+            transfer_spl(
+                &ctx.accounts.ta_input_searcher.to_account_info(),
+                &ctx.accounts.ta_input_trader.to_account_info(),
+                &ctx.accounts.token_program.to_account_info(),
+                &ctx.accounts.searcher.to_account_info(),
+                amount_to_trader,
+            )?;
+        }
+
+        if data.amount_output > 0 {
+            transfer_spl(
+                &ctx.accounts.ta_output_trader.to_account_info(),
+                &ctx.accounts.ta_output_searcher.to_account_info(),
+                &ctx.accounts.token_program.to_account_info(),
+                &ctx.accounts.trader.to_account_info(),
+                data.amount_output,
+            )?;
+        }
+
+        Ok(())
     }
 }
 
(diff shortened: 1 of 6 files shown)


contracts/svm/programs/express_relay/src/utils.rs
```diff
@@ -49,6 +54,50 @@ pub fn transfer_lamports_cpi<'info>(
...
+    token_program: &AccountInfo<'info>,
+    auth: &AccountInfo<'info>,
+    mint: &InterfaceAccount<'info, Mint>,
+    amount: u64,
+    fee_ppm: u64,
+) -> Result<u64> {
+    let fee = amount
+        .checked_mul(fee_ppm)
+        .ok_or(ProgramError::ArithmeticOverflow)?
+        / 1_000_000;
```
guibescos, line 92: bit of nitpick but this can overflow. For example if amount is 100M pyth tokens

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +398,119 @@ pub struct WithdrawFees<'info> {
...
+    )]
+    pub ta_output_searcher: InterfaceAccount<'info, TokenAccount>,
+
+    #[account(
+        mut,
+        token::mint = mint_input,
+        token::authority = trader,
+        token::token_program = token_program_input
+    )]
+    pub ta_input_trader: InterfaceAccount<'info, TokenAccount>,
```
guibescos, line 476: Is it possible that the trader doesn't have an input account?

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +398,119 @@ pub struct WithdrawFees<'info> {
...
+    )]
+    pub ta_input_searcher: InterfaceAccount<'info, TokenAccount>,
+
+    #[account(
+        mut,
+        token::mint = mint_output,
+        token::authority = searcher,
+        token::token_program = token_program_output
+    )]
+    pub ta_output_searcher: InterfaceAccount<'info, TokenAccount>,
```
guibescos, line 468: is it possible that the searcher doesn't have an output ata yet?

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +398,119 @@ pub struct WithdrawFees<'info> {
...
+    #[account(
+        init_if_needed,
+        payer = searcher,
+        token::mint = mint_output,
+        token::authority = router,
+        token::token_program = token_program_output
+    )]
+    pub ta_output_router: InterfaceAccount<'info, TokenAccount>,
+
+    #[account(address = ID)]
```
guibescos, line 504: nit I think this check already happens when you do `Program<'info, ExpressRelay>`

contracts/svm/programs/express_relay/src/utils.rs
anihamde: we could use bps instead of parts per million. probably sufficient precision. wdyt?

contracts/svm/programs/express_relay/src/lib.rs
anihamde: possible, although initializing this would need to be added into the logic of the method. also this wouldn't be straightforward to initialize if we wanted to keep it generic to all token accounts (not just ATAs)

contracts/svm/programs/express_relay/src/lib.rs
anihamde: ah ok, i wasn't sure if this was checking the pubkey too. good to know, can delete that

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +398,119 @@ pub struct WithdrawFees<'info> {
...
+        mint_output.key().as_ref(),
+        &data.nonce.to_le_bytes(),
+    ], bump)]
+    pub permission: AccountInfo<'info>,
+
+    /// CHECK: don't care what this looks like
+    #[account(mut)]
+    pub router: UncheckedAccount<'info>,
+
+    /// CHECK: this cannot be checked against ConfigRouter bc it may not be initialized bc anchor. we need to check this config even when unused to make sure unique fee splits don't exist
```
guibescos, line 441: this comment is confusing, suggest updating to something like
Some routers might have an initialized `ConfigRouter` at the enforced PDA address, that specified the routing fee split. If the `ConfigRouter` is unitialized we will default to the routing fee split defined in the global `ExpressRelayMetadata`. We still need to pass the account to check whether it's initialized or not.

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +398,119 @@ pub struct WithdrawFees<'info> {
...
+    #[account(mut)]
+    pub searcher: Signer<'info>,
+
+    pub trader: Signer<'info>,
+
+    pub relayer_signer: Signer<'info>,
+
+    #[account(seeds = [
+        SEED_SWAP,
+        searcher.key().as_ref(),
```
guibescos, line 429: Is it weird that this depends on the searcher? I think this means each searcher is bidding in a different auction

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +398,119 @@ pub struct WithdrawFees<'info> {
...
+
+#[derive(Accounts)]
+#[instruction(data: Box<SwapArgs>)]
+pub struct Swap<'info> {
+    #[account(mut)]
+    pub searcher: Signer<'info>,
+
+    pub trader: Signer<'info>,
+
+    pub relayer_signer: Signer<'info>,
```
guibescos, line 425: technically you don't need relayer_signer here right. Only in Submit Bid

contracts/svm/programs/express_relay/src/lib.rs
guibescos: mmmmmmmmmm, why is ok to have init_if_needed for the router token accounts but not here

contracts/svm/programs/express_relay/src/utils.rs
guibescos: this issue is pretty minor and can also be solved by doing arithmetic in u128 so if you feel like ppm is useful let's keep it like this and we can update the arithmetic later

contracts/svm/programs/express_relay/src/utils.rs
```diff
@@ -49,6 +57,55 @@ pub fn transfer_lamports_cpi<'info>(
...
+    from_ta: &AccountInfo<'info>,
+    to_ta: &AccountInfo<'info>,
+    token_program: &AccountInfo<'info>,
+    auth: &AccountInfo<'info>,
+    mint: &InterfaceAccount<'info, Mint>,
+    amount: u64,
+    fee_ppm: u64,
+) -> Result<u64> {
+    // we do the computation in u128 to avoid overflow in case of large amounts
+    let fee_u128 = (amount as u128)
```
guibescos, line 93: It's more idomatic to use `u128::from` and `u64::try_from`

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -513,9 +514,9 @@ pub struct Swap<'info> {
 
     // This account may not be initialized, and it should be set to the router's ATA to prevent the router from
     // receiving output tokens in an arbitrary token account. We perform this check within the instruction.
+    // To initialize the ATA, the transaction must contain an instruction to create the associated token account.
     #[account(
-        init_if_needed,
-        payer = searcher,
+        mut,
```
guibescos, line 519: I like init_if_needed more tbh, makes the sdk simpler

contracts/svm/testing/Cargo.toml
```diff
@@ -6,6 +6,9 @@ edition = "2021"
 
 [dependencies]
 anchor-lang = ">=0.29.0"
+anchor-spl = ">=0.29.0"
+spl-token = ">=0.7.0"
+spl-associated-token-account = ">=0.6.0"
```
guibescos, line 11: nitpick: can you make these dev dependencies instead of dependencies?

PR #176, opened 2024-10-17
feat: store svm recent blockhashes

guibescos:


auction-server/src/server.rs
```diff
@@ -93,11 +95,10 @@ use {
 async fn fault_tolerant_handler<F, Fut>(name: String, f: F)
 where
     F: Fn() -> Fut,
-    Fut: Future<Output = anyhow::Result<()>> + Send + 'static,
-    Fut::Output: Send + 'static,
+    Fut: Future<Output = anyhow::Result<()>>,
 {
     loop {
-        let res = tokio::spawn(f()).await;
+        let res = AssertUnwindSafe(f()).catch_unwind().await;
```
guibescos, line 101: This is a superior way of doing this; there's no reason to spawn a task that we await immediately.
Internally tokio uses AssertUnwindSafe too.

auction-server/src/api.rs
```diff
@@ -404,7 +404,7 @@ pub async fn start_api(run_options: RunOptions, store: Arc<StoreNew>) -> Result<
 
     // The generated OpenAPI document contains "discriminator" fields which are not generated correctly to be supported by redoc
     // We need to remove them from the document to make sure redoc can render the document correctly
-    let original_doc = serde_json::to_value(&ApiDoc::openapi())
+    let original_doc = serde_json::to_value(ApiDoc::openapi())
```
guibescos, line 407: clippy did this

auction-server/src/auction.rs
```diff
@@ -1340,6 +1355,21 @@ impl ChainStore for ChainStoreSvm {
     const CHAIN_TYPE: models::ChainType = models::ChainType::Svm;
     const AUCTION_MINIMUM_LIFETIME: Duration = Duration::from_millis(400);
 
+    async fn get_next_trigger<'a>(
```
danimhr, line 1358: is it possible to extract slot number from this block trigger?
We need to know slot number to make sure running the conclusion loop for svm every 150 slots.

auction-server/src/auction.rs
```diff
@@ -1348,8 +1378,19 @@ impl ChainStore for ChainStoreSvm {
...
-        Ok(slot_subscribe)
+        let (block_subscribe, _) = client
+            .block_subscribe(
+                solana_client::rpc_config::RpcBlockSubscribeFilter::All,
+                Some(RpcBlockSubscribeConfig {
+                    encoding:                          None,
+                    transaction_details:               Some(TransactionDetails::None),
+                    show_rewards:                      Some(false),
+                    max_supported_transaction_version: None,
+                    commitment:                        Some(CommitmentConfig::confirmed()),
```
danimhr, line 1389: is it possible to listen for `finalized` blocks?
we want to sign the transaction with this block hash and it's safer to use `finalized` for this

auction-server/src/server.rs
danimhr: Could you please describe what's the difference btw these two methods?

auction-server/src/server.rs
```diff
@@ -380,6 +381,7 @@ fn setup_svm(
             (
                 chain_id.clone(),
                 ChainStoreSvm {
+                    recent_blockhash: None.into(),
```
danimhr, line 384: We can store the initial block hash by creating a single RPC request when setting up the server. It's better to not let this value be None

auction-server/Cargo.toml
```diff
@@ -49,6 +49,7 @@ serde_with = { version = "3.9.0", features = ["base64"] }
 anchor-lang = "0.30.1"
 express-relay = { path = "../contracts/svm/programs/express_relay" }
 solana-rpc-client = "2.0.13"
+solana-transaction-status = "2.0.13"
```
danimhr, line 52: Do we need it?

auction-server/src/api.rs
danimhr, line 407: :-? what's the problem with the previous version?

PR #177, opened 2024-10-18
Feat/wallet opportunity

danimhr:
This PR aims to add svm integration for wallets

### DIFF
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -9,6 +9,13 @@ use {
             ConfigEvm,
         },
         models,
+        opportunity::service::{
+            get_live_opportunities_by_permission_key::GetOpportunitiesByPermissionKeyInput,
+            ChainType as OpportunityChainType,
+            ChainTypeEvm as OpportunityChainTypeEvm,
+            ChainTypeSvm as OpportunityChainTypeSvm,
+            Service as OpportunityService,
+        },
         server::{
             EXIT_CHECK_INTERVAL,
             SHOULD_EXIT,
@@ -30,6 +37,7 @@ use {
             SimulatedBidSvm,
             SimulatedBidTrait,
             Store,
+            StoreNew,
         },
         traced_client::TracedClient,
     },
@@ -530,27 +538,42 @@ async fn submit_auction_for_lock<T: ChainStore>(
 }
 
 #[tracing::instrument(skip_all)]
-async fn submit_auction<T: ChainStore>(
-    store: Arc<Store>,
+async fn handle_auction<T: ChainStore>(
+    store_new: Arc<StoreNew>,
     chain_store: &T,
     permission_key: Bytes,
     chain_id: String,
 ) -> Result<()> {
+    let store = store_new.store.clone();
     if !chain_store.should_submit_auction(&permission_key) {
-        return Ok(());
+        // TODO Look at get_quote in opportunity module
+        if !chain_store
+            .opportunity_exists(store_new, &permission_key)
+            .await
+        {
+            // Fetch all pending bids and mark them as lost
+            let bids = chain_store
+                .get_bids(&permission_key)
+                .await
+                .into_iter()
+                .filter(|bid| models::BidStatus::Pending == bid.status.clone().into())
+                .collect();
+            broadcast_lost_bids(store.clone(), chain_store, bids, vec![], None, None).await;
+        }
+        Ok(())
+    } else {
+        let auction_lock = chain_store.get_auction_lock(permission_key.clone()).await;
+        let result = submit_auction_for_lock(
+            store.clone(),
+            chain_store,
+            permission_key.clone(),
+            chain_id,
+            auction_lock,
+        )
+        .await;
+        chain_store.remove_auction_lock(&permission_key).await;
+        result
     }
-
-    let auction_lock = chain_store.get_auction_lock(permission_key.clone()).await;
-    let result = submit_auction_for_lock(
-        store.clone(),
-        chain_store,
-        permission_key.clone(),
-        chain_id,
-        auction_lock,
-    )
-    .await;
-    chain_store.remove_auction_lock(&permission_key).await;
-    result
 }
 
 pub fn get_express_relay_contract(
@@ -574,7 +597,11 @@ pub fn get_express_relay_contract(
     SignableExpressRelayContract::new(address, client)
 }
 
-async fn submit_auctions<T: ChainStore>(store: Arc<Store>, chain_store: &T, chain_id: String) {
+async fn handle_auctions<T: ChainStore>(
+    store_new: Arc<StoreNew>,
+    chain_store: &T,
+    chain_id: String,
+) {
     let permission_keys = chain_store.get_permission_keys_for_auction().await;
 
     tracing::info!(
@@ -583,16 +610,17 @@ async fn submit_auctions<T: ChainStore>(store: Arc<Store>, chain_store: &T, chai
         auction_len = permission_keys.len()
     );
 
+    // let store = store_new.clone().store.clone();
     for permission_key in permission_keys.iter() {
-        store.task_tracker.spawn({
-            let (store, permission_key, chain_id) =
-                (store.clone(), permission_key.clone(), chain_id.clone());
+        store_new.store.task_tracker.spawn({
+            let (store_new, permission_key, chain_id) =
+                (store_new.clone(), permission_key.clone(), chain_id.clone());
             async move {
                 let result = match T::CHAIN_TYPE {
-                    models::ChainType::Evm => match store.chains.get(&chain_id) {
+                    models::ChainType::Evm => match store_new.store.chains.get(&chain_id) {
                         Some(chain_store) => {
-                            submit_auction(
-                                store.clone(),
+                            handle_auction(
+                                store_new.clone(),
                                 chain_store,
                                 permission_key.clone(),
                                 chain_id.clone(),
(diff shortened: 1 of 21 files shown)


auction-server/src/opportunity/api.rs
```diff
@@ -235,14 +239,101 @@ pub struct TokenAmountSvm {
...
+    #[serde_as(as = "DisplayFromStr")]
+    pub input_token_mint:            Pubkey,
+    /// The token mint address of the output token.
+    #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
+    #[serde_as(as = "DisplayFromStr")]
+    pub output_token_mint:           Pubkey,
+    /// The input token amount that the user wants to swap.
+    #[schema(example = 100, value_type = u64)]
+    pub input_token_amount:          u64,
+    /// The maximum slippage percentage that the user is willing to accept.
```
anihamde, line 262: this is missing an `output_token_amount`?

auction-server/src/auction.rs
```diff
@@ -1428,10 +1546,26 @@ impl ChainStore for ChainStoreSvm {
...
+                    Ok(Some(vec![BidStatusSvm::Expired { result: tx_hash }]))
+                } else {
+                    Ok(None)
+                }
             }
         }
     }
+
+    fn should_submit_auction(&self, permission_key: &Bytes) -> bool {
+        !permission_key.starts_with(&self.phantom_router_account.to_bytes())
```
anihamde, line 1560: they could technically use different routers in the future, and for other similar style integrations, it may be annoying to add a bunch of these to the config. can we instead change the permission_keys iter to be a struct like

```
{
    permission_key: __,
    server_submit_auction: bool
}
```

auction-server/src/opportunity/service/estimate_price.rs
```diff
@@ -0,0 +1,22 @@
...
+
+pub struct EstimatePriceInput {
+    pub quote_create: entities::QuoteCreate,
+}
+
+impl Service<ChainTypeSvm> {
+    #[tracing::instrument(skip_all)]
+    pub async fn estimate_price(&self, _input: EstimatePriceInput) -> Result<u64, RestError> {
+        // TODO implement
+        return Ok(0);
```
anihamde, line 20: i think i understand, you intend for the server to grab the reference price? there are some tradeoffs here:
- if the server grabs the reference price, then all the load is at the server level. e.g. if this process is very long/slow, then the server can get backed up in api calls here
- if the wallet provides a few prices/the best price, then this method involves a lot less work and fewer external deps.

we should at the least think about some load management here to balance making any external calls here

auction-server/src/auction.rs
```diff
@@ -1428,10 +1546,26 @@ impl ChainStore for ChainStoreSvm {
             }])),
             None => {
                 // not yet confirmed
-                Ok(None)
+                if bids[0].initiation_time + BID_MAXIMUM_LIFE_TIME_SVM < OffsetDateTime::now_utc() {
```
anihamde, line 1549: this would end up not being confirmed bc it failed on-chain?

anihamde: in general, this branch is a bit confusing, and the documentation in `quote.rs` differs. it would be worth making sure these say the same thing, and also documenting the circumstances under which this branch comes into play here

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -0,0 +1,243 @@
...
+            permission_account,
+            // TODO extract latest block hash
+            block_hash: Hash::default(),
+            program: entities::OpportunitySvmProgram::Phantom(
+                entities::OpportunitySvmProgramWallet {
+                    user_wallet_address:         quote_create.user_wallet_address,
+                    maximum_slippage_percentage: quote_create.maximum_slippage_percentage,
+                },
+            ),
+            // TODO extract latest slog
```
anihamde, line 89: slot

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -0,0 +1,243 @@
...
+        sleep(BID_COLLECTION_TIME).await;
+
+        let bid_collection_time = OffsetDateTime::now_utc();
+        let mut bids = chain_store.get_bids(&opportunity.permission_key).await;
+
+        // Add metrics
+        let labels = [
+            ("chain_id", input.quote_create.chain_id.to_string()),
+            ("wallet", "phantom".to_string()),
+            ("total_bids", bids.len().to_string()),
```
m30m, line 127: I don't think this label has low cardinality

auction-server/src/auction.rs
```diff
@@ -1001,20 +1063,35 @@ pub async fn handle_bid_svm(
...
-        false => Err(RestError::BadParameters("Invalid signatures".to_string())),
+        .filter(|(signature, pubkey)| {
+            signature.verify(pubkey.as_ref(), &message_bytes) || (*pubkey).eq(relayer_pubkey)
+        })
+        .count();
+
+    let missing_signatures = bid.transaction.signatures.len() - valid_signatures_count;
+    match missing_signatures {
+        0 => Ok(()),
+        1 => {
```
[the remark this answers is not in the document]
danimhr: I didnt want to query for the opp at first :-?
But I think I was wrong and it's not a bad idea to fetch it :-?

auction-server/src/auction.rs
```diff
@@ -1520,18 +1679,75 @@ async fn run_submission_loop<T: ChainStore>(
     Ok(())
 }
 
-pub async fn run_submission_loop_evm(store: Arc<Store>, chain_id: String) -> Result<()> {
-    let chain_store = store
+pub async fn run_submission_loop_evm(store_new: Arc<StoreNew>, chain_id: String) -> Result<()> {
+    let chain_store = store_new
+        .store
```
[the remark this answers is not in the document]
danimhr: This is gonna make life easier in lots of places :D

auction-server/src/auction.rs
```diff
@@ -986,8 +1030,26 @@ pub async fn handle_bid_svm(
         submit_bid_instruction,
     )?;
 
-    verify_signatures_svm(&bid, &chain_store.express_relay_svm.relayer.pubkey())?;
+    let bytes_permission_key = Bytes::from(&permission_key.0);
```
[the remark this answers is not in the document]
danimhr: I'll add another task for this

auction-server/src/auction.rs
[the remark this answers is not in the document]
danimhr: I'll add a task for it

auction-server/src/auction.rs
```diff
@@ -986,7 +1031,15 @@ pub async fn handle_bid_svm(
...
+    verify_signatures_svm(
+        store_new,
+        chain_store,
+        &bid,
+        &chain_store.express_relay_svm.relayer.pubkey(),
+        &bytes_permission_key,
+    )
+    .await?;
```
m30m, line 1042: we should verify that the wallet bids also include another instruction to the swap program with the appropriate accounts and fields. Please add a TODO for it here since we don't have the program yet.

auction-server/src/opportunity/service/get_missing_signers.rs
```diff
@@ -0,0 +1,17 @@
...
+    super::{
+        ChainTypeSvm,
+        Service,
+    },
+    crate::opportunity::entities,
+    solana_sdk::pubkey::Pubkey,
+};
+
+impl Service<ChainTypeSvm> {
+    pub fn get_missing_signers(&self, opportunity: entities::OpportunitySvm) -> Vec<Pubkey> {
```
m30m, line 11: this function doesn't need &self. Maybe it's better to implement it inside OpportunitySvm?

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -0,0 +1,268 @@
...
+                quote_create: input.quote_create.clone(),
+            })
+            .await?;
+
+        let opportunity_create = self
+            .get_opportunity_create_for_quote(input.quote_create.clone(), output_amount)
+            .await?;
+        let opportunity = self
+            .add_opportunity(AddOpportunityInput {
+                opportunity: opportunity_create.clone(),
```
m30m, line 133: no need to clone

auction-server/src/auction.rs
```diff
@@ -1520,18 +1663,75 @@ async fn run_submission_loop<T: ChainStore>(
...
+                .to_string()]),
+            RpcTransactionLogsConfig {
+                commitment: Some(CommitmentConfig::confirmed()),
+            },
+        )
+        .await?;
+
+    while !SHOULD_EXIT.load(Ordering::Acquire) {
+        tokio::select! {
+            rpc_log = stream.next() => {
```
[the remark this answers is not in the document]
danimhr: Added a todo for it

auction-server/src/opportunity/api.rs
```diff
@@ -235,14 +239,101 @@ pub struct TokenAmountSvm {
...
+
 /// Program specific parameters for the opportunity
 #[serde_as]
 #[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug)]
 #[serde(tag = "program")]
 pub enum OpportunityCreateProgramParamsV1Svm {
-    /// Limo program specific parameters for the opportunity
-    /// It contains the Limo order to be executed, encoded in base64
-    /// SDKs will decode this order and create transaction for bidding on the opportunity
+    /// Limo program specific parameters for the opportunity.
```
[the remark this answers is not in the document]
danimhr: Added a task for it

m30m (approved):

PR #179, opened 2024-10-23
Feat/svm watcher clean

guibescos:


auction-server/src/api/ws.rs
```diff
@@ -287,6 +291,19 @@ impl Subscriber {
         Ok(())
     }
 
+    async fn handle_svm_chain_update(&mut self, svm_chain_update: SvmChainUpdate) -> Result<()> {
+        tracing::Span::current().record("name", "svm_chain_update");
```
danimhr, line 295: While we are setting the span name in the `handle_update`, we can safely remove this line.

auction-server/src/server.rs
```diff
@@ -35,6 +35,7 @@ use {
         },
         traced_client::TracedClient,
         traced_sender_svm::TracedSenderSvm,
+        watcher::run_svm_watcher_loop,
```
danimhr, line 38: move the `_svm` to the end of function name please.

auction-server/src/state.rs
```diff
@@ -1092,6 +1092,13 @@ impl Store {
         };
     }
 
+    pub fn broadcast_svm_chain_update(&self, update: SvmChainUpdate) {
+        match self.event_sender.send(UpdateEvent::SvmChainUpdate(update)) {
```
danimhr, line 1096: While we are not doing anything in the `Ok` arm, we can use `if let Err(error) = ...`

auction-server/src/watcher.rs
```diff
@@ -0,0 +1,79 @@
...
+    let ws_client = chain_store.get_ws_client().await?;
+    let (mut stream, _) = ws_client
+        .block_subscribe(
+            solana_client::rpc_config::RpcBlockSubscribeFilter::All,
+            Some(solana_client::rpc_config::RpcBlockSubscribeConfig {
+                encoding:                          None,
+                transaction_details:               Some(TransactionDetails::None), // we don't need transaction data
+                show_rewards:                      Some(false),
+                max_supported_transaction_version: None,
+                commitment:                        Some(CommitmentConfig::finalized()),
```
danimhr, line 49: Lets use `confirmed` while this is not yet fixed in quick node. Or If we have another endpoint which is gonna work with finalized config.

auction-server/src/watcher.rs
```diff
@@ -0,0 +1,79 @@
...
+    while !SHOULD_EXIT.load(Ordering::Acquire) {
+        tokio::select! {
+            update = stream.next() => {
+                if let Some(block_update) = update {
+                    let blockhash = block_update.value.block.map(|b| b.blockhash).map(|b| Hash::from_str(&b)).transpose()?;
+                    if let Some(blockhash) = blockhash {
+                        store.broadcast_svm_chain_update(SvmChainUpdate {
+                            chain_id: chain_id.clone(),
+                            blockhash,
+                            });
```
danimhr, line 63: Why the `});` is a tab forward :-?
Doesn't the formatter raise error or fix it?

auction-server/src/watcher.rs
```diff
@@ -0,0 +1,79 @@
...
+
+                            return Ok(());
```
danimhr, line 65: Is it okay to return here?

auction-server/src/watcher.rs
```diff
@@ -0,0 +1,79 @@
...
+                        }
+                    else {
+                        tracing::warn!("Blockhash not found for slot {} on chain: {}", block_update.value.slot, chain_id);
```
danimhr, line 68: it's better to use this format:

`tracing::warn!(slot = block_update.value.slot, chain = chain_id, "Blockhash not found for slot");`

danimhr (commented):
Please store the latest block data in memory as well.
Additionally it would be great if update the opportunity module as well Or let me know if we intend to do it in another PR.

auction-server/src/api/ws.rs
guibescos: done, should we do the same for `handle_new_opportunity` and `handle_bid_status_update`

auction-server/src/watcher.rs
guibescos: I just realized cargo fmt doesn't work inside macros, but fixed it manually

auction-server/src/watcher.rs
guibescos: Done

auction-server/src/watcher.rs
guibescos: oh, good catch

auction-server/src/watcher.rs
guibescos: done

auction-server/src/api/ws.rs
guibescos: fixed it for handle_new_opportunity and handle_bid_status_update too

danimhr (approved):

PR #180, opened 2024-10-24
opportunity provider script for limo

anihamde:
This PR adds a script to create orders on the limo program for testing on mainnet.


scripts/opportunity-provider-limo/index.ts
```diff
@@ -0,0 +1,206 @@
...
+
+async function createOpportunities(
+  skExecutor: Keypair,
+  limoClient: limo.LimoClient,
+  opportunitiesPath: string,
+  count: number,
+  edge: number
+) {
+  const opportunities = loadOpportunities(opportunitiesPath);
+  for (let i = 0; i < opportunities.length; i++) {
```
m30m, line 87: `for(const opportunity of opportunities)`

scripts/opportunity-provider-limo/index.ts
```diff
@@ -0,0 +1,206 @@
...
+    prices[index] = price;
+  }
+
+  const mintDecimals = await getDecimals(connection, token);
+  return prices[index] / Math.pow(10, mintDecimals);
+}
+
+async function createOpportunities(
+  skExecutor: Keypair,
+  limoClient: limo.LimoClient,
```
m30m, line 81: i'd rather get the global_config and create the limoClient inside this function.

scripts/opportunity-provider-limo/index.ts
```diff
@@ -0,0 +1,206 @@
...
+        }`
+      );
+      console.log(
+        `Output: ${outputToken.symbol}, ${
+          amountOutput / Math.pow(10, decimalsOutput)
+        }`
+      );
+
+      const signature = (
+        await limoClient.createOrderGeneric(
```
m30m, line 134: Maybe just create the limoClient here and use another variable to create the connection at the beginning of this function. It seems unnatural that we just get the connection from the client for other purposes.

scripts/opportunity-provider-limo/index.ts
```diff
@@ -0,0 +1,206 @@
...
+    description: "Path to opportunities file",
+    type: "string",
+    default: "opportunities.json",
+  })
+  .option("count", {
+    description: "Number of opportunities to create",
+    type: "number",
+    default: 10,
+  })
+  .option("edge", {
```
m30m, line 174: not sure how common this term is. Maybe just use markup?

scripts/opportunity-provider-limo/index.ts
```diff
@@ -0,0 +1,206 @@
...
+    description: "SVM RPC endpoint",
+    type: "string",
+    demandOption: true,
+  })
+  .option("opportunities", {
+    description: "Path to opportunities file",
+    type: "string",
+    default: "opportunities.json",
+  })
+  .option("count", {
```
m30m, line 169: if we want to run this in a loop, we should adjust the code. Otherwise, the deployment will be in some sort of crash loop

scripts/opportunity-provider-limo/index.ts
```diff
@@ -0,0 +1,206 @@
...
+      )[0];
+      console.log(`Created opportunity: ${signature}`);
+    }
+  }
+}
+
+const argv = yargs(hideBin(process.argv))
+  .option("sk-payer", {
+    description:
+      "Secret key of address to submit transactions with. If action is 'create', this keypair creates the order in Limo. In 64-byte base58 format",
```
m30m, line 150: what does it mean the action is 'create'?

scripts/opportunity-provider-limo/opportunities.json
```diff
@@ -0,0 +1,15 @@
+[
```
m30m, line 1: maybe rename to opportunities.sample.json

scripts/opportunity-provider-limo/index.ts
```diff
@@ -0,0 +1,225 @@
...
+    type: "string",
+    demandOption: true,
+  })
+  .option("opportunities", {
+    description: "Path to opportunities file",
+    type: "string",
+    default: "opportunities.json",
+  })
+  .option("count", {
+    description: "Number of instances to create of each opportunity",
```
m30m, line 172: can you clarify here that this is per interval

scripts/opportunity-provider-limo/index.ts
```diff
@@ -0,0 +1,225 @@
...
+  skExecutor: Keypair,
+  connection: Connection,
+  globalConfig: PublicKey,
+  opportunitiesPath: string,
+  count: number,
+  markup: number
+) {
+  const opportunities = loadOpportunities(opportunitiesPath);
+  for (const opportunity of opportunities) {
+    for (let j = 0; j < count; j++) {
```
m30m, line 89: a `j` without an `i` is very lonely :D

m30m (approved):

PR #181, opened 2024-10-24
Fix error for wher we have submitted the same svm transaction twice

danimhr:
This PR aims to fix issues for conclusion auctions with more than one bid for svm chain.

### DIFF
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -1010,6 +1010,12 @@ fn extract_bid_data_svm(
     Ok((submit_bid_data.bid_amount, PermissionKeySvm(permission_key)))
 }
 
+impl PartialEq<SimulatedBidSvm> for BidSvm {
+    fn eq(&self, other: &SimulatedBidSvm) -> bool {
+        self.transaction == other.transaction && self.chain_id == other.core_fields.chain_id
+    }
+}
+
 #[tracing::instrument(skip_all)]
 pub async fn handle_bid_svm(
     store_new: Arc<StoreNew>,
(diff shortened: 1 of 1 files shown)


auction-server/src/auction.rs
```diff
@@ -1043,6 +1049,12 @@ pub async fn handle_bid_svm(
     // TODO we should verify that the wallet bids also include another instruction to the swap program with the appropriate accounts and fields
     simulate_bid_svm(chain_store, &bid).await?;
 
+    // Check if the bid does not exists
```
anihamde, line 1052: nit: could make comment clearer, e.g. "Check if duplicate bid"?

auction-server/src/auction.rs
```diff
@@ -1589,34 +1601,36 @@ impl ChainStore for ChainStoreSvm {
             .try_into()
             .map_err(|_| anyhow!("Invalid svm signature"))?;
         if bids.len() != 1 {
```
anihamde, line 1603: a small nit: the function that gets the `bids` to pass into `get_bid_results` is `bids_for_submitted_auction`. technically, `bids` will have length greater than 1 iff multiple bids in _different_ auctions have the same tx hash. For this reason, it may be good idea to rename the `bids_for_submitted_auction` function to reflect this, e.g. maybe to `bids_matching_tx_hash`?

auction-server/src/auction.rs
```diff
@@ -1589,34 +1601,36 @@ impl ChainStore for ChainStoreSvm {
...
                     result: Some(tx_hash),
                 },
-            }])),
+            },
             None => {
                 // not yet confirmed
                 // TODO Use the correct version of the expiration algorithm, which is:
                 // the tx is not expired as long as the block hash is still recent.
                 // Assuming a certain block time, the two minute threshold is good enough but in some cases, it's not correct.
                 if bids[0].initiation_time + BID_MAXIMUM_LIFE_TIME_SVM < OffsetDateTime::now_utc() {
```
anihamde, line 1624: the expiration of bids will depend on their individual initiation time, not on another bid's initiation time. so this should actually be a loop? e.g. it's possible the first of the bids gets processed but doesn't get confirmed but a later one does get confirmed?

auction-server/src/auction.rs
danimhr: What about evm?

m30m (approved):

PR #182, opened 2024-10-24
refactor: use async traits and simplify some function call

m30m:


auction-server/src/api.rs
```diff
@@ -418,7 +418,7 @@ pub async fn start_api(run_options: RunOptions, store: Arc<StoreNew>) -> Result<
 
     // The generated OpenAPI document contains "discriminator" fields which are not generated correctly to be supported by redoc
     // We need to remove them from the document to make sure redoc can render the document correctly
-    let original_doc = serde_json::to_value(&ApiDoc::openapi())
+    let original_doc = serde_json::to_value(ApiDoc::openapi())
```
danimhr, line 421: Why we should change this line?

m30m: not sure, the linter automatically did this.

PR #183, opened 2024-10-24
feat: remove blockhash

guibescos:
I think we can merge this. Tilt still passes.
svm-linonade aka `scripts/limonade/src/index.ts` submits the opportunity with the blockhash and slot, but to fix that I need to fix `@pythnetwork/express-relay-js` in pyth-crosschain.
I will also update simpleSearcherLimo there.


danimhr (approved):

danimhr:
@guibescos I think we can keep the slot

PR #184, opened 2024-10-25
feat: DEX Router

anihamde:
Script to route limit orders to on-chain liquidity


scripts/dex-router/src/index.ts
```diff
@@ -0,0 +1,294 @@
...
+      this.connectionSvm,
+      order.state.globalConfig
+    );
+    const inputMintDecimals = await clientLimo.getOrderInputMintDecimals(order);
+    const outputMintDecimals = await clientLimo.getOrderOutputMintDecimals(
+      order
+    );
+
+    let routerOutputs: RouterOutput[] = [];
+    for (const router of this.routers) {
```
m30m, line 112: nit (since we only have a single router ATM):a better way to do this is via Promise.all + map

scripts/dex-router/src/index.ts
```diff
@@ -0,0 +1,294 @@
...
+        );
+        routerOutputs.push(routerOutput);
+      } catch (error) {
+        console.error(`Failed to route order: ${error}`);
+      }
+    }
+    if (routerOutputs.length === 0) {
+      throw new Error("No routers available to route order");
+    }
+    const routerBest = routerOutputs.reduce((prev, curr) => {
```
m30m, line 129: prev is not a good name for an accumulator. can be renamed to bestSoFar.

scripts/dex-router/src/index.ts
```diff
@@ -0,0 +1,294 @@
...
+    );
+
+    const txMsg = new TransactionMessage({
+      payerKey: this.executor.publicKey,
+      recentBlockhash: opportunity.blockHash,
+      instructions: [
+        ixSubmitBid,
+        ...ixsFlashTakeOrder.createAtaIxs,
+        ixsFlashTakeOrder.startFlashIx,
+        ...ixsRouter,
```
m30m, line 181: why not inline `...routerBest.ixs,`?

scripts/dex-router/src/index.ts
```diff
@@ -0,0 +1,294 @@
...
+          opportunity.order.state.outputMint,
+          opportunity.order.state.remainingInputAmount,
+          this.executor.publicKey
+        );
+        routerOutputs.push(routerOutput);
+      } catch (error) {
+        console.error(`Failed to route order: ${error}`);
+      }
+    }
+    if (routerOutputs.length === 0) {
```
m30m, line 126: shouldn't we check if the price is actually higher than the limit order?

scripts/dex-router/src/index.ts
```diff
@@ -0,0 +1,294 @@
...
+    const outputAmountDecimals = new Decimal(Number(routerBest.amountOut)).div(
+      new Decimal(10).pow(outputMintDecimals)
+    );
+
+    const ixsFlashTakeOrder = await clientLimo.flashTakeOrderIxs(
+      this.executor.publicKey,
+      order,
+      inputAmountDecimals,
+      outputAmountDecimals,
+      new PublicKey("PytERJFhAKuNNuaiXkApLfWzwNwSNDACpigT3LwQfou"),
```
m30m, line 146: `SVM_CONSTANTS[this.chainId].expressRelayProgram`

scripts/dex-router/src/index.ts
```diff
@@ -0,0 +1,294 @@
...
+        this.connectionSvm
+      );
+    }
+
+    const ixSubmitBid = await this.client.constructSubmitBidInstruction(
+      this.executor.publicKey,
+      router,
+      order.address,
+      bidAmount,
+      new anchor.BN(Math.round(Date.now() / 1000 + 60)),
```
m30m, line 168: nit maybe put inside a variable like `oneMinuteLater`

scripts/dex-router/src/index.ts
```diff
@@ -0,0 +1,294 @@
...
+  Keypair,
+  PublicKey,
+  TransactionMessage,
+  VersionedTransaction,
+} from "@solana/web3.js";
+import * as anchor from "@coral-xyz/anchor";
+import { Decimal } from "decimal.js";
+import * as limo from "@kamino-finance/limo-sdk";
+import { getPdaAuthority } from "@kamino-finance/limo-sdk/dist/utils";
+import { getVersionedTxSize } from "./utils/size";
```
m30m, line 24: unused

scripts/dex-router/src/types.ts
```diff
@@ -0,0 +1,19 @@
...
+export type RouterOutput = {
+  ixs: TransactionInstruction[];
+  amountIn: bigint;
+  amountOut: bigint;
+  lookupTableAddresses?: PublicKey[];
+};
+
+export interface Router {
+  route: (
+    chainId: string,
```
m30m, line 12: I would put `chainId`, `executor`, and `connection` inside the constructor of router and exclude them from this interface since they tend to not change during a single run.

scripts/dex-router/src/utils/size.ts
```diff
@@ -0,0 +1,165 @@
...
+  if (verbose) {
+    console.log("Size of header: ", sizeHeader);
+    console.log("Size of blockhash: ", sizeBlockhash);
+    console.log("Size of signatures: ", sizeSignatures);
+    console.log("Size of accounts: ", sizeAccounts);
+    console.log("Size of number of instructions: ", sizeNInstructions);
+    console.log("Size of ixs: ", ixsSize);
+    console.log("Size of lookup: ", lookupSize);
+  }
+
```
m30m, line 155: I would add an assertion that `tx.serialize().length` is equal to this sum

PR #185, opened 2024-10-25
feat: move sdks from pyth-crosschain repo

m30m:


sdk/solidity/package.json
```diff
@@ -4,15 +4,15 @@
...
+    "url": "https://github.com/pyth-network/per",
+    "directory": "sdk/solidity"
   },
   "publishConfig": {
     "access": "public"
   },
   "scripts": {
     "format": "prettier --write .",
-    "generate-abi": "generate-abis IExpressRelay IExpressRelayFeeReceiver",
+    "generate-abi": "forge inspect IExpressRelay abi > abis/IExpressRelay.json && forge inspect IExpressRelayFeeReceiver abi > abis/IExpressRelayFeeReceiver.json",
```
danimhr, line 15: Do we need `forge` to work with this command? Do we have it in dependency of solidity sdk?

examples/easy_lend/foundry.toml
```diff
@@ -3,7 +3,7 @@ src = "contracts"
 out = "out"
 libs = [
     'lib',
-    '../../../node_modules',
+    'node_modules',
     '../../../target_chains/ethereum/sdk/solidity'
```
danimhr, line 7: Update this please

examples/easy_lend/tsconfig.json
```diff
@@ -9,7 +9,9 @@
     "strict": true,
     "esModuleInterop": true,
     "resolveJsonModule": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "rootDir": "src/",
+    "outDir": "./tslib"
```
danimhr, line 14: why here the output directory is `tslib` but in `sdk/js/tsconfig.json` it is `lib`?

Tiltfile
```diff
@@ -213,3 +213,11 @@ local_resource(
     serve_cmd="npm run --prefix scripts/limonade limonade -- --global-config $(solana-keygen pubkey keypairs/limo_global_config.json)  --endpoint http://127.0.0.1:9000 --chain-id development-solana --api-key $(poetry -C per_sdk run python3 create_limo_profile.py) --rpc-endpoint %s" % rpc_url_solana,
     resource_deps=["svm-initialize-programs", "auction-server"],
 )
+
+
+local_resource(
+    "svm-searcher-py",
+    serve_cmd="poetry run python3 -m express_relay.searcher.examples.simple_searcher_svm --endpoint-express-relay http://127.0.0.1:9000 --chain-id development-solana --private-key-json-file ../../keypairs/searcher.json --endpoint-svm http://127.0.0.1:8899 --bid 10000000 --fill-rate 1",
```
danimhr, line 220: is it gonna be for limo?

danimhr (commented):
@m30m 
- Is is easy to lock the npm / node version with sth like `nvmrc` or other tools?
- Is it easy to remove per_sdk? if not, could you please add a task for it?
- Do not squash all the commits when merging the branch,. It's a good idea to squash only recent commits you made (manually) for this migration to make history clean.

examples/easy_lend/tsconfig.json
m30m: forge dependencies are by default in the lib folder. So we chose a different folder name to avoid any conflicts :D
I will add a comment for it

Tiltfile
m30m: yes

examples/easy_lend/tsconfig.json
m30m: ok, it's a json file so I can't add a comment :))

sdk/solidity/package.json
m30m: yes, it's not a npm dependency.
I need to port the workflow which runs these commands and I need to install forge in that workflow.

m30m:
> @m30m
> 
> * Is is easy to lock the npm / node version with sth like `nvmrc` or other tools?
Pushed something for it
> * Is it easy to remove per_sdk? if not, could you please add a task for it?
I tried it a bit, it's not very easy since the sdks and scripts we ported do not accept some customisation that the current scripts have. But it's definitely possible and we should do it.
> * Do not squash all the commits when merging the branch,. It's a good idea to squash only recent commits you made (manually) for this migration to make history clean.
Sure

danimhr (approved):
Please squash latest commits to a single one tnx

PR #187, opened 2024-10-25
feat: separation url for  tx send in svm

m30m:


auction-server/src/server.rs
```diff
@@ -171,37 +155,9 @@ async fn setup_chain_store(
                     let (chain_id, chain_config, wallet) =
                         (chain_id.clone(), chain_config.clone(), wallet.clone());
                     Some(async move {
```
danimhr, line 157: Do we need to use async move?
Maybe we can pass the function directly?

auction-server/src/server.rs
```diff
@@ -260,9 +216,9 @@ pub async fn start_server(run_options: RunOptions) -> Result<()> {
     let wallet = run_options.subwallet_private_key.parse::<LocalWallet>()?;
     tracing::info!("Using wallet address: {:?}", wallet.address());
 
-    let chains = setup_chain_store(config_map.clone(), wallet.clone()).await?;
+    let chains_evm = setup_chainstore_evm(config_map.clone(), wallet.clone()).await?;
```
danimhr, line 219: `chain_store` or `chainstore`?

auction-server/src/state.rs
```diff
@@ -472,16 +495,121 @@ pub struct ChainStoreEvm {
...
+
+        let express_relay_contract = get_express_relay_contract(
+            config.express_relay_contract,
+            provider.clone(),
+            relayer.clone(),
+            config.legacy_tx,
+            id,
+        );
+
+
```
danimhr, line 546: Remove this extra space

auction-server/src/state.rs
```diff
@@ -472,16 +495,121 @@ pub struct ChainStoreEvm {
...
+                "Failed to connect to chain({chain_id}) at {rpc_addr}: {:?}",
+                err,
+                chain_id = chain_id,
+                rpc_addr = chain_config.geth_rpc_addr
+            )
+        })?;
+        provider.set_interval(Duration::from_secs(chain_config.poll_interval));
+        Ok(provider)
+    }
+    pub async fn create_store(
```
danimhr, line 525: I think this function is like the `new` function inside the svm chain_store. Can we add these to `ChainStore` trait?
Maybe we can have sth like `try_new` or `create_store` for the ChainStore trait. wdyt?

auction-server/src/state.rs
```diff
@@ -472,16 +495,121 @@ pub struct ChainStoreEvm {
...
+                .parse::<usize>()
+                .expect("Failed to parse router account position"),
+        };
+
+        Self {
+            name: chain_id.clone(),
+            core_fields: Default::default(),
+            client: TracedSenderSvm::new_client(
+                chain_id.clone(),
+                config.rpc_addr.as_str(),
```
danimhr, line 587: Maybe we can rename the `rpc_addr` to make the diff clear (with the other rpc). Maybe we can use chatgpt for better naming.

auction-server/src/server.rs
m30m: filter_map output should be a bunch of `Option<impl Future>` types, so I don't think it's easy to do it, since we want to wrap it inside an Ok((chain_id.clone, ....))

auction-server/src/state.rs
m30m: I don't think it's worth the abstraction. We know if we are creating Svm version or an Evm version right now but if at some point we have a function that's creating both of them without a lot of distinguishing, we can unify them for sure.

danimhr (approved):

PR #190, opened 2024-10-27
Fix get bids result for empty bids list

danimhr:
This PR aims to fix the panic for getting empty bids list result

### DIFF
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -1420,6 +1420,10 @@ impl ChainStore for ChainStoreEvm {
         bids: Vec<Self::SimulatedBid>,
         tx_hash: Vec<u8>,
     ) -> Result<Option<Vec<<Self::SimulatedBid as SimulatedBidTrait>::StatusType>>> {
+        if bids.is_empty() {
+            return Ok(None);
+        }
+
         let tx_hash = H256::from_slice(tx_hash.as_slice());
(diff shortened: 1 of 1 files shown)


auction-server/src/auction.rs
```diff
@@ -1420,6 +1420,10 @@ impl ChainStore for ChainStoreEvm {
         bids: Vec<Self::SimulatedBid>,
         tx_hash: Vec<u8>,
     ) -> Result<Option<Vec<<Self::SimulatedBid as SimulatedBidTrait>::StatusType>>> {
+        if bids.is_empty() {
```
anihamde, line 1423: this should work, but i don't completely follow why bids would be empty here? you get the bids for a tx_hash of an auction that was submitted, which should have at least one bid that was submitted, so why would this ever be empty? maybe you could add a comment to clarify this

anihamde (approved):

auction-server/src/auction.rs
danimhr: One possible scenario:
Lets say we have two auctions with same tx hash. We try to conclude the first auction and broadcast the result for it's bids. Then the second auction is going to be concluded. The first line of `conclude_auction` if filtering the bids which are submitted. This list is gonna be empty (because the first conclusion changed the bids state). So we will call the `get_bids_result` with an empty bids list

PR #191, opened 2024-10-27
Fix concluding auctions with no bids

danimhr:
This PR aims to fix auction conclusion with no bids


auction-server/src/auction.rs
```diff
@@ -1573,7 +1569,7 @@ impl ChainStore for ChainStoreSvm {
         tx_hash: Vec<u8>,
     ) -> Result<Option<Vec<<Self::SimulatedBid as SimulatedBidTrait>::StatusType>>> {
         if bids.is_empty() {
-            return Ok(None);
+            return Ok(Some(vec![]));
```
anihamde, line 1572: good catch. the one thing is i think you could describe the situation that leads to empty bid array in a comment for clarification, bc ow it does require a lot of reasoning on the part of the code reader to comprehend, as a result of the complexity

anihamde (approved):

PR #193, opened 2024-10-29
Feat/remove opportunity

danimhr:
This PR aims to add remove opportunity functionality to auction server


m30m:
Are you going to make the necessary changes in the limonade too?

auction-server/src/opportunity/api.rs
```diff
@@ -722,10 +765,43 @@ pub async fn post_quote(
...
+    Ok(StatusCode::NO_CONTENT)
+}
+
+pub fn get_routes(store: Arc<StoreNew>) -> Router<Arc<StoreNew>> {
     Router::new()
         .route("/", post(post_opportunity))
         .route("/quote", post(post_quote))
         .route("/", get(get_opportunities))
         .route("/:opportunity_id/bids", post(opportunity_bid))
+        .route("/", login_required!(store, delete(delete_opportunities)))
```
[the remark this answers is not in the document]
danimhr: So in this case, we need to run the verification to make sure the opportunity is not valid before removing it

auction-server/src/opportunity/repository/remove_opportunities.rs
```diff
@@ -0,0 +1,43 @@
...
+        &self,
+        db: &sqlx::Pool<Postgres>,
+        permission_key: PermissionKey,
+        reason: OpportunityRemovalReason,
+    ) -> Vec<T::Opportunity> {
+        let mut write_guard = self.in_memory_store.opportunities.write().await;
+        let opportunitties = write_guard.remove(&permission_key);
+        drop(write_guard);
+
+        tokio::spawn({
```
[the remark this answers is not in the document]
danimhr: Will update it to be like remove_opportunity

auction-server/src/opportunity/repository/get_live_opportunities.rs
```diff
@@ -3,19 +3,19 @@ use {
         InMemoryStore,
         Repository,
     },
-    crate::kernel::entities::PermissionKey,
+    crate::opportunity::entities,
 };
 
 impl<T: InMemoryStore> Repository<T> {
-    pub async fn get_live_opportunities_by_permission_key(
+    pub async fn get_live_opportunities(
```
m30m, line 10: can we refactor the name of these functions. we have `get_opportunities`, `get_all_opportunities`, and `get_live_opportunities`. By just looking at the names, I can't be sure which of these is querying the db and which of the mare not.

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -292,4 +295,28 @@ impl OpportunitySvm {
             OpportunitySvmProgram::Limo(_) => vec![],
         }
     }
+
+    pub fn get_permission_key(router: Pubkey, permission_account: Pubkey) -> PermissionKey {
+        PermissionKey::from([router.to_bytes(), permission_account.to_bytes()].concat())
+    }
+}
+
+impl From<OpportunitySvm> for api::OpportunityDelete {
```
m30m, line 304: These conversions are lossy, so I prefer a specific method over using the From trait.
https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -292,4 +295,28 @@ impl OpportunitySvm {
...
+        api::OpportunityDelete::Svm(api::OpportunityDeleteSvm::V1(api::OpportunityDeleteV1Svm {
+            chain_id:           val.chain_id.clone(),
+            permission_account: val.permission_account,
+            router:             val.router,
+            program:            val.program.into(),
+        }))
+    }
+}
+
+impl From<OpportunitySvmProgram> for api::ProgramSvm {
```
m30m, line 315: Same here:
These conversions are lossy, so I prefer a specific method over using the From trait.
https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from

m30m (approved):

auction-server/src/opportunity/entities/opportunity_svm.rs
danimhr: I think this one makes more sense, but I removed the one for `OpportunityDelete`

PR #199, opened 2024-10-29
Feat/minimum bid lifetime svm

anihamde:
This PR adds a check to make sure SVM bids have minimum `deadline`s. The exact minimum deadline is dictated by the submission type of the bid.

### DIFF
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -1086,7 +1086,7 @@ impl PartialEq<SimulatedBidSvm> for BidSvm {
     }
 }
 
-const BID_MINIMUM_LIFE_TIME_SVM_SERVER: i64 = 2;
+const BID_MINIMUM_LIFE_TIME_SVM_SERVER: i64 = 5;
 const BID_MINIMUM_LIFE_TIME_SVM_OTHER: i64 = 10;
 const BID_MINIMUM_LIFE_TIME_SVM_INVALID: i64 = 0;
 
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -1047,7 +1047,7 @@ async fn extract_bid_data_svm(
     chain_store: &ChainStoreSvm,
     tx: VersionedTransaction,
     client: &RpcClient,
-) -> Result<(u64, PermissionKeySvm), RestError> {
+) -> Result<(u64, PermissionKeySvm, i64), RestError> {
     let submit_bid_instruction = verify_submit_bid_instruction_svm(
         &chain_store.config.express_relay_program_id,
         tx.clone(),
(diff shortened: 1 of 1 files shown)


auction-server/src/auction.rs
```diff
@@ -1097,10 +1105,26 @@ pub async fn handle_bid_svm(
...
+        SubmitType::SubmitByServer => BID_MINIMUM_LIFE_TIME_SVM_SERVER,
+        SubmitType::SubmitByOther => BID_MINIMUM_LIFE_TIME_SVM_OTHER,
+        SubmitType::Invalid => BID_MINIMUM_LIFE_TIME_SVM_INVALID,
+    };
+    // TODO: this uses the time at the server, which can lead to issues if Solana ever experiences clock drift
+    // using the time at the server is not ideal, but the alternative is to make an RPC call to get the Solana block time
+    // we should make this more robust, possibly by polling the current block time in the background
+    if deadline < OffsetDateTime::now_utc().unix_timestamp() + minimum_bid_life_time {
+        return Err(RestError::BadParameters(
+            "Bid deadline is too short".to_string(),
```
m30m, line 1125: we can return what's the minimum lifetime for a more helpful error message

auction-server/src/auction.rs
```diff
@@ -1097,10 +1105,26 @@ pub async fn handle_bid_svm(
...
         extract_bid_data_svm(chain_store, bid.transaction.clone(), &chain_store.client).await?;
 
     let bytes_permission_key = Bytes::from(&permission_key.0);
+    let minimum_bid_life_time = match chain_store
+        .get_submission_state(store_new.clone(), &bytes_permission_key)
+        .await
+    {
+        SubmitType::SubmitByServer => BID_MINIMUM_LIFE_TIME_SVM_SERVER,
+        SubmitType::SubmitByOther => BID_MINIMUM_LIFE_TIME_SVM_OTHER,
+        SubmitType::Invalid => BID_MINIMUM_LIFE_TIME_SVM_INVALID,
```
m30m, line 1118: this logic is only to handle SubmitByServer and SubmitByOther lifetimes. We should not check this lifetime if the SubmitType is invalid. Right now, in this branch, we return an error which is not very helpful and should be handled by `verify_signatures_svm`

auction-server/src/auction.rs
```diff
@@ -1047,7 +1047,7 @@ async fn extract_bid_data_svm(
     chain_store: &ChainStoreSvm,
     tx: VersionedTransaction,
     client: &RpcClient,
-) -> Result<(u64, PermissionKeySvm), RestError> {
+) -> Result<(u64, PermissionKeySvm, i64), RestError> {
```
m30m, line 1050: I would create a struct for (u64, PermissionKeySvm, i64)

auction-server/src/auction.rs
```diff
@@ -1082,6 +1086,10 @@ impl PartialEq<SimulatedBidSvm> for BidSvm {
     }
 }
 
+const BID_MINIMUM_LIFE_TIME_SVM_SERVER: i64 = 5;
+const BID_MINIMUM_LIFE_TIME_SVM_OTHER: i64 = 10;
+const BID_MINIMUM_LIFE_TIME_SVM_INVALID: i64 = 0;
```
m30m, line 1091: maybe a function that gets a SubmitType and returns `Option<i64>` where None means do not check for lifetime

auction-server/src/auction.rs
anihamde: this actually shouldn't return an error here, it leaves the error for `verify_signatrues_svm`, because `BID_MINIMUM_LIFE_TIME_SVM_INVALID` is 0. if you have suggestions on better ways to handle this, we can do that too, i agree that this feels a bit weird right now

m30m (approved):

PR #200, opened 2024-10-30
Add remove opportunties to sdk

danimhr:
This PR aims to add remove opportunties ws handler and js client method to er sdks

### DIFF
--- a/sdk/js/src/examples/simpleSearcherLimo.ts
+++ b/sdk/js/src/examples/simpleSearcherLimo.ts
@@ -6,7 +6,12 @@ import {
   Opportunity,
   OpportunitySvm,
 } from "../index";
-import { BidStatusUpdate, ChainId, SvmChainUpdate } from "../types";
+import {
+  BidStatusUpdate,
+  ChainId,
+  OpportunityDelete,
+  SvmChainUpdate,
+} from "../types";
 import { SVM_CONSTANTS } from "../const";
 
 import * as anchor from "@coral-xyz/anchor";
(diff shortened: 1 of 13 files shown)


sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -186,6 +192,10 @@ class SimpleSearcherLimo {
     this.recentBlockhash[update.chain_id] = update.blockhash;
   }
 
+  async removeOpportunitiesHandler(opportunityDelete: OpportunityDelete) {
+    console.log("Removing opportunities", opportunityDelete);
```
m30m, line 196: nit: Maybe add a comment that devs should implement their own removal logic or change the log message to something like `Opportunity X doesn't exist anymore`. Right now it feels like something is happening in the client but it's not

sdk/python/express_relay/searcher/examples/simple_searcher_svm.py
```diff
@@ -182,6 +185,8 @@ async def assess_opportunity(self, opp: OpportunitySvm) -> BidSvm | None:
     async def svm_chain_update_callback(self, svm_chain_update: SvmChainUpdate):
         self.recent_blockhash[svm_chain_update.chain_id] = svm_chain_update.blockhash
 
+    async def remove_opportunities_callback(self, opportunity_delete: OpportunityDelete):
+        print(f"Remove opportunities: {opportunity_delete}")
```
m30m, line 189: same nit here

m30m (approved):

PR #201, opened 2024-10-31
fix: clippy warns fail pre-commit now

m30m:
there is an on-going issue where if you run clippy with the `--fix` flag, the return code is always 0 even when it should fail.
I just removed the fix param for now to make sure we catch such issues firsthand.

### DIFF
--- a/auction-server/src/opportunity/service/mod.rs
+++ b/auction-server/src/opportunity/service/mod.rs
@@ -63,6 +63,7 @@ pub struct ConfigSvm {
     pub wallet_program_router_account: Pubkey,
 }
 
+#[allow(dead_code)]
 pub trait Config: Send + Sync {}
 
 impl Config for ConfigEvm {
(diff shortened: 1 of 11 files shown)


auction-server/src/opportunity/service/mod.rs
```diff
@@ -63,6 +63,7 @@ pub struct ConfigSvm {
     pub wallet_program_router_account: Pubkey,
 }
 
+#[allow(dead_code)]
```
anihamde, line 66: this feels kinda gross to drop in selectively, is there a CL arg we can set or something to ignore deadcode lints?

anihamde (approved):

auction-server/src/opportunity/service/mod.rs
m30m: agreed but I think it's still useful to have it and treat these as exceptions to the rule. I don't expect many more to appear in the next few months and these were sort of accumulated and ignored from the inception of the project.

PR #205, opened 2024-11-01
Feat/get all live opps

anihamde:
This PR adds options to help users get all live opportunities from the server, by paginating by time and altering the number of returned opportunities in the response.

### DIFF
--- a/auction-server/src/opportunity/api.rs
+++ b/auction-server/src/opportunity/api.rs
@@ -168,8 +168,8 @@ pub struct GetOpportunitiesQueryParams {
     #[param(example="2024-05-23T21:26:57.329954Z", value_type = Option<String>)]
     #[serde(default, with = "crate::serde::nullable_datetime")]
     pub from_time:      Option<OffsetDateTime>,
-    /// The maximum number of opportunities to return. Capped at 100.
-    #[param(example = "20", value_type = usize)]
+    /// The maximum number of opportunities to return. Capped at 100; if more than 100 requested, at most 100 will be returned.
+    #[param(example = "20", value_type = usize, maximum = 100)]
     #[serde(default = "default_limit")]
     limit:              usize,
 }
--- a/auction-server/src/opportunity/api.rs
+++ b/auction-server/src/opportunity/api.rs
@@ -736,10 +736,7 @@ pub async fn get_opportunities(
         Ok(Json(
             opportunities
                 .into_iter()
-                .take(std::cmp::min(
-                    query_params.limit,
-                    OPPORTUNITY_PAGE_SIZE_CAP as usize,
-                ))
+                .take(std::cmp::min(query_params.limit, OPPORTUNITY_PAGE_SIZE_CAP))
                 .collect(),
         ))
     }
--- a/auction-server/src/opportunity/api.rs
+++ b/auction-server/src/opportunity/api.rs
@@ -686,7 +686,7 @@ pub async fn post_opportunity(
 /// Fetch opportunities ready for execution or historical opportunities
 /// depending on the mode. You need to provide `chain_id` for historical mode.
 /// Opportunities are sorted by creation time in ascending order.
-/// Total number of opportunities returned is limited by 20.
+/// Total number of opportunities returned is capped by the server to preserve bandwidth.
 #[utoipa::path(get, path = "/v1/opportunities", responses(
 (status = 200, description = "Array of opportunities ready for bidding", body = Vec < Opportunity >),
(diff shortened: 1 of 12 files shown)


sdk/js/src/index.ts
```diff
@@ -265,12 +265,25 @@ export class Client {
...
    */
-  async getOpportunities(chainId?: string): Promise<Opportunity[]> {
+  async getOpportunities(
+    chainId?: string,
+    fromTime?: number,
+    limit?: number
+  ): Promise<Opportunity[]> {
     const client = createClient<paths>(this.clientOptions);
+    const fromTimeStr = fromTime
+      ? `${new Date(fromTime / 1000).toISOString().replace("Z", "")}${(
```
m30m, line 279: you don't need to convert to microsecond format. I suggest getting the fromTime param as a Date type similar to what you did in the python sdk

m30m (approved):

auction-server/src/opportunity/api.rs
```diff
@@ -729,7 +736,10 @@ pub async fn get_opportunities(
         Ok(Json(
             opportunities
                 .into_iter()
-                .take(repository::OPPORTUNITY_PAGE_SIZE as usize)
+                .take(std::cmp::min(
+                    query_params.limit,
+                    OPPORTUNITY_PAGE_SIZE_CAP as usize,
```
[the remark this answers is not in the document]
anihamde: i think it's better to return the max rather than error out. maybe we could add a note/warning field to the message, but i think it's bad to error here since the user can still recover some data and go on to recover more data by paginating

sdk/js/src/index.ts
anihamde: that's also an option. i did it this way to make it easier for the user to specify a microsecond time and have it query that, since `Date()` does not keep the microsecond-level granularity

auction-server/src/opportunity/repository/mod.rs
```diff
@@ -21,7 +21,7 @@ mod remove_opportunities;
 mod remove_opportunity;
 
 pub use models::*;
-pub const OPPORTUNITY_PAGE_SIZE: i32 = 20;
+pub const OPPORTUNITY_PAGE_SIZE_CAP: i32 = 100;
```
m30m, line 24: why not usize?

auction-server/src/opportunity/api.rs
```diff
@@ -678,7 +685,7 @@ pub async fn post_opportunity(
 
 /// Fetch opportunities ready for execution or historical opportunities
 /// depending on the mode. You need to provide `chain_id` for historical mode.
-/// Opportunities are sorted by creation time in ascending order in historical mode.
+/// Opportunities are sorted by creation time in ascending order.
 /// Total number of opportunities returned is limited by 20.
```
m30m, line 689: update number here

auction-server/src/opportunity/api.rs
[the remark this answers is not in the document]
m30m: people are not going to read the docs carefully and will use a large number here. Let's say a thousand. Then they will see that the endpoint is returning them 44 values and think ok, this would work later on and can return 1000 values. This would lead to a silent failure and unnoticed bugs but if we validate everything in the beginning and throw errors they will catch it in testing phase and wouldn't have any issues on prod.

sdk/js/src/index.ts
m30m: I don't think anybody is going to use such precise time formats anyway. If it's easier for user to use a Date object and do some simple day arithmetics with it, I would prefer accepting a Date object here.

auction-server/src/opportunity/api.rs
```diff
@@ -161,10 +164,14 @@ pub struct GetOpportunitiesQueryParams {
     /// The permission key to filter the opportunities by. Used only in historical mode.
     #[param(example = "0xdeadbeef", value_type = Option< String >)]
     pub permission_key: Option<Bytes>,
-    /// The time to get the opportunities from. Used only in historical mode.
+    /// The time to get the opportunities from.
     #[param(example="2024-05-23T21:26:57.329954Z", value_type = Option<String>)]
     #[serde(default, with = "crate::serde::nullable_datetime")]
     pub from_time:      Option<OffsetDateTime>,
+    /// The maximum number of opportunities to return. Capped at 100.
+    #[param(example = "20", value_type = usize)]
```
m30m, line 172: I think a better way to include the 100 parameter is using the maximum attribute in the param.
Even better we can use something like this to automate the validation:
https://github.com/Keats/validator

PR #210, opened 2024-11-04
Add remove opportunity broadcast for expired opportunities

danimhr:
This PR aims to broadcast remove opportunity message for those opportunities that are expired.

### DIFF
--- a/auction-server/src/opportunity/repository/remove_opportunity.rs
+++ b/auction-server/src/opportunity/repository/remove_opportunity.rs
@@ -4,7 +4,10 @@ use {
         InMemoryStore,
         Repository,
     },
-    crate::opportunity::entities::Opportunity,
+    crate::opportunity::entities::{
+        self,
+        Opportunity,
+    },
     sqlx::Postgres,
     time::{
         OffsetDateTime,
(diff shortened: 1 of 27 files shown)


auction-server/src/opportunity/repository/remove_opportunity.rs
```diff
@@ -17,8 +20,9 @@ impl<T: InMemoryStore> Repository<T> {
         &self,
         db: &sqlx::Pool<Postgres>,
         opportunity: &T::Opportunity,
-        reason: OpportunityRemovalReason,
+        reason: entities::OpportunityRemovalReason,
```
m30m, line 23: maybe `reason: impl Into<OpportunityRemovalReason>,` ? if you just want it to be convertable

auction-server/src/opportunity/service/remove_invalid_or_expired_opportunities.rs
```diff
@@ -59,16 +66,37 @@ where
...
+                        .remove_opportunity(&self.db, opportunity, reason)
                         .await
                     {
-                        tracing::error!("Failed to remove opportunity: {}", e);
+                        Ok(()) => {
+                            if self
+                                .repo
+                                .get_in_memory_opportunities_by_key(&opportunity.get_key())
+                                .await
+                                .is_empty()
```
m30m, line 83: I think this logic deserves a few lines of comment

m30m (commented):
I don't think we need the evm variations for now. Let's discuss this today

auction-server/src/opportunity/repository/remove_opportunity.rs
danimhr: I want to get the removal reason type of the entities :-?
We should use entities or primary types in repo functions

PR #213, opened 2024-11-05
fix: backport js changes to python 

guibescos:
Apply the changes from `@kamino-finance/limo-sdk` 0.6.0 to the Python sdk.
TODO:
- Apply rounding errors fix
- Fix WSOL as output

### DIFF
--- a/sdk/python/express_relay/svm/limo_client.py
+++ b/sdk/python/express_relay/svm/limo_client.py
@@ -193,6 +193,9 @@ class LimoClient:
         input_mint_decimals: int,
         output_mint_decimals: int,
         express_relay_program_id: Pubkey,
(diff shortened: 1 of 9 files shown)


sdk/python/express_relay/svm/limo_client.py
```diff
@@ -193,6 +193,9 @@ async def take_order_ix(
         input_mint_decimals: int,
         output_mint_decimals: int,
         express_relay_program_id: Pubkey,
+        permissionless_tip_lamports: int = 0,
```
m30m, line 196: we actually don't want to expose these to searchers. it's more for testing without going through the auction server. So we can just remove it here

m30m (approved):

PR #214, opened 2024-11-06
Add latency and margin to bid for simple searchers

danimhr:
This PR aims to add latency and bid margin for simple searchers. It separates the private key for js and python searchers to test concurrency better while using tilt

### DIFF
--- a/sdk/js/src/examples/simpleSearcherLimo.ts
+++ b/sdk/js/src/examples/simpleSearcherLimo.ts
@@ -143,13 +143,15 @@ class SimpleSearcherLimo {
       limoClient.getProgramID(),
       order.state.globalConfig
     );
-    const bidAmount = new anchor.BN(argv.bid);
+
+    let bidAmount = new anchor.BN(argv.bid);
     if (this.bidMargin !== 0) {
       const margin = new anchor.BN(
         Math.floor(Math.random() * (this.bidMargin * 2 + 1)) - this.bidMargin
       );
-      bidAmount.add(margin);
+      bidAmount = bidAmount.add(margin);
     }
+
     if (!this.expressRelayConfig) {
       this.expressRelayConfig = await this.client.getExpressRelaySvmConfig(
         this.chainId,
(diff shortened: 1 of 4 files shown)


sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -142,6 +144,12 @@ class SimpleSearcherLimo {
       order.state.globalConfig
     );
     const bidAmount = new anchor.BN(argv.bid);
+    if (this.bidMargin !== 0) {
+      const margin = new anchor.BN(
+        Math.floor(Math.random() * (this.bidMargin * 2 + 1)) - this.bidMargin
+      );
+      bidAmount.add(margin);
```
m30m, line 151: this doesn't change the bidAmount. I think you should do
`bidAmount = bidAmount.add(margin)`

sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -259,6 +273,18 @@ const argv = yargs(hideBin(process.argv))
     type: "number",
     default: 100,
   })
+  .option("with-latency", {
```
m30m, line 276: These options are mainly for internal testing and ideally we shouldn't expose them in this simple searcher.
Can we have another script and another class like `ExtendedSearcherLimo` that inherits the `SimpleSearcherLimo` and just overrides the opportunityHandler? We can add the latency there and change the bid each time before calling the generateBid function. Therefore the base script remains the same.

sdk/python/express_relay/searcher/examples/simple_searcher_svm.py
```diff
@@ -258,6 +284,7 @@ async def main():
         with open(args.private_key_json_file, "r") as f:
             searcher_keypair = Keypair.from_json(f.read())
 
+    print(args)
```
m30m, line 287: ?

sdk/python/express_relay/searcher/examples/simple_searcher_svm.py
```diff
@@ -240,6 +252,20 @@ async def main():
         default=100,
         help="How much of the initial order size to fill in percentage. Default is 100%",
     )
+    parser.add_argument(
```
m30m, line 255: same here. ideally these should be in a separate script or at least we should document that these are just for internal testing and searchers do not need to implement them

m30m (approved):

PR #216, opened 2024-11-06
fix: wsol as output

guibescos:
Replaces the TODO by some actual implementation

### DIFF
--- a/sdk/python/express_relay/svm/limo_client.py
+++ b/sdk/python/express_relay/svm/limo_client.py
@@ -1,5 +1,4 @@
 from decimal import Decimal
-import logging
 from typing import Sequence, List, TypedDict, Tuple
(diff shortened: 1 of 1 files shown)


anihamde (approved):

sdk/python/express_relay/svm/limo_client.py
```diff
@@ -143,7 +143,7 @@ async def get_init_if_needed_wsol_create_and_close_ixs(
 
         create_ixs = []
         close_ixs = []
-        if ata_info.value is None:
+        if ata_info.value is None or len(ata_info.value.data)==0:
```
guibescos, line 146: This fixes a bug where the account can exist but not be an ATA, for example if someone sends sol to the account

PR #218, opened 2024-11-07
fix: better logging on migration command

m30m:
Running the migration command printed no helpful logs to make sure it's working correctly


danimhr (approved):
Do we need to update version?

PR #219, opened 2024-11-07
Refactor/reorg limo examples

anihamde:
This PR refactors the limo examples in the SDK to be more easily human readable

### DIFF
--- a/sdk/js/src/examples/simpleSearcherLimo.ts
+++ b/sdk/js/src/examples/simpleSearcherLimo.ts
@@ -96,7 +96,7 @@ class SimpleSearcherLimo {
   }
 
   /**
-   * Generates a bid for a given opportunity by transferring assets from the searcher's wallet
+   * Generates a bid for a given opportunity. The transaction in this bid transfers assets from the searcher's wallet to fulfill the limit order.
    * @param opportunity The SVM opportunity to bid on
    * @returns The generated bid object
    */
@@ -107,7 +107,7 @@ class SimpleSearcherLimo {
       order.state.globalConfig
     );
 
-    const ixsTakeOrder = await this.formTakeOrderIxs(limoClient, order);
+    const ixsTakeOrder = await this.generateTakeOrderIxs(limoClient, order);
     const txRaw = new anchor.web3.Transaction().add(...ixsTakeOrder);
 
     const bidData = await this.getBidData(limoClient, order);
@@ -130,7 +130,7 @@ class SimpleSearcherLimo {
   }
 
   /**
-   * Fetches the router address, bid amount, and relayer addresses required to create a valid bid
+   * Gets the router address, bid amount, and relayer addresses required to create a valid bid
    * @param limoClient The Limo client
    * @param order The limit order to be fulfilled
    * @returns The fetched bid data
@@ -171,7 +171,7 @@ class SimpleSearcherLimo {
    * @param order The limit order to be fulfilled
    * @returns The Limo TakeOrder instructions used to fulfill the order
    */
-  async formTakeOrderIxs(
+  async generateTakeOrderIxs(
     limoClient: limo.LimoClient,
     order: OrderStateAndAddress
   ): Promise<TransactionInstruction[]> {
--- a/sdk/js/src/examples/simpleSearcherLimo.ts
+++ b/sdk/js/src/examples/simpleSearcherLimo.ts
@@ -8,6 +8,7 @@ import {
 } from "../index";
 import {
   BidStatusUpdate,
+  BidSvm,
   ChainId,
   OpportunityDelete,
   SvmChainUpdate,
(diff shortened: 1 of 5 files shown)


sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -80,12 +95,86 @@ class SimpleSearcherLimo {
     return decimals;
   }
 
-  async generateBid(opportunity: OpportunitySvm, recentBlockhash: Blockhash) {
+  /**
+   * Generates a bid for a given opportunity by transferring assets from the searcher's wallet
```
guibescos, line 99: This sounds like transferring the assets is required to generate the bid

sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -80,12 +95,86 @@ class SimpleSearcherLimo {
...
+   * @returns The generated bid object
+   */
+  async generateBid(opportunity: OpportunitySvm): Promise<BidSvm> {
     const order = opportunity.order;
     const limoClient = new limo.LimoClient(
       this.connectionSvm,
       order.state.globalConfig
     );
+
+    const ixsTakeOrder = await this.formTakeOrderIxs(limoClient, order);
```
guibescos, line 110: `generateTakeOrderIxs`?

sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -80,12 +95,86 @@ class SimpleSearcherLimo {
...
+      bidData.relayerFeeReceiver
+    );
+
+    bid.transaction.recentBlockhash = this.recentBlockhash[this.chainId];
+    bid.transaction.sign(this.searcher);
+    return bid;
+  }
+
+  /**
+   * Fetches the router address, bid amount, and relayer addresses required to create a valid bid
```
guibescos, line 133: nitpick bid amount doesn't get fetched but randomly generated

sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -80,12 +95,86 @@ class SimpleSearcherLimo {
...
+   */
+  async generateBid(opportunity: OpportunitySvm): Promise<BidSvm> {
     const order = opportunity.order;
     const limoClient = new limo.LimoClient(
       this.connectionSvm,
       order.state.globalConfig
     );
+
+    const ixsTakeOrder = await this.formTakeOrderIxs(limoClient, order);
+    const txRaw = new anchor.web3.Transaction().add(...ixsTakeOrder);
```
guibescos, line 111: Maybe `formTakeOrderIxs` should return this tx?

sdk/python/express_relay/searcher/examples/simple_searcher_svm.py
```diff
@@ -81,12 +89,16 @@ async def opportunity_callback(self, opp: Opportunity):
         if self.with_latency:
             await asyncio.sleep(0.5 * random.random())
 
+        if opp.chain_id not in self.recent_blockhash:
```
guibescos, line 92: put this check before `if self.with_latency` for consistency with typescript?

sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -80,12 +95,86 @@ class SimpleSearcherLimo {
...
+    order: OrderStateAndAddress
+  ): Promise<BidData> {
+    const router = getPdaAuthority(
+      limoClient.getProgramID(),
+      order.state.globalConfig
+    );
+    let bidAmount = new anchor.BN(argv.bid);
+    if (this.bidMargin !== 0) {
+      const margin = new anchor.BN(
+        Math.floor(Math.random() * (this.bidMargin * 2 + 1)) - this.bidMargin
```
guibescos, line 149: feel like the math for choosing the bid value should be in a function

sdk/python/express_relay/searcher/examples/simple_searcher_svm.py
```diff
@@ -119,7 +131,46 @@ async def get_mint_decimals(self, mint: Pubkey) -> int:
         return self.mint_decimals_cache[str(mint)]
 
     async def assess_opportunity(self, opp: OpportunitySvm) -> BidSvm | None:
```
guibescos, line 133: remove None from the return type if `assess_opportunity` can't return None anymore

sdk/python/express_relay/searcher/examples/simple_searcher_svm.py
```diff
@@ -119,7 +131,46 @@ async def get_mint_decimals(self, mint: Pubkey) -> int:
...
+        """
+        Method to assess an opportunity and return a bid if the opportunity is worth taking. This method always returns a bid for any valid opportunity.
+
+        Args:
+            opp: An object representing a single opportunity.
+        Returns:
+            A bid object if the opportunity is worth taking to be submitted to the Express Relay server, otherwise None.
```
guibescos, line 140: when can this return `None`?

guibescos (approved):
I think breaking down `generateBid` is a step in the right direction

sdk/python/express_relay/searcher/examples/simple_searcher_svm.py
anihamde: it doesn't return `None` because it always returns a bid. the typing here is just for illustration's sake and to provide an example of how one may write a bid assessment method (that ofc has the ability to return `None` and signal no bid to be made)

sdk/js/src/examples/simpleSearcherLimo.ts
anihamde: i thought this was a clearer way to construct it, simply because the function name is easier to understand. Otherwise, the output of the method would be a half-baked transaction, and the method name should change to reflect that but i can't think of a good name for it

sdk/js/src/examples/simpleSearcherLimo.ts
anihamde: this is probably going to be removed from the simple searcher class as per the task we talked about in the meeting today. so i'll leave it for now

PR #221, opened 2024-11-08
fix/db-duplicate-opp

anihamde:
This PR adds a check as to whether an opportunity has already been added to the db; if so, it does not add it to the db, so as to spare the db from having many copies of the same opportunity.


auction-server/src/opportunity/repository/add_opportunity.rs
```diff
@@ -30,28 +30,35 @@ impl<T: InMemoryStore> Repository<T> {
...
-        serde_json::to_value(metadata).expect("Failed to serialize metadata"),
-        serde_json::to_value(&opportunity.sell_tokens).unwrap(),
-        serde_json::to_value(&opportunity.buy_tokens).unwrap())
-            .execute(db)
-            .await
-            .map_err(|e| {
-                tracing::error!("DB: Failed to insert opportunity: {}", e);
-                RestError::TemporarilyUnavailable
-            })?;
+        if self
```
danimhr, line 33: I think we can have sth like this:

`get_or_add_db_opportunity`
`exists_db_opportunity`
`add_db_opportunity`

While the opportunity dont have any unique key, I think we dont need transaction for the query. So I think the above structure will be sufficient for our case.

auction-server/src/opportunity/repository/add_opportunity.rs
```diff
@@ -63,4 +70,42 @@ impl<T: InMemoryStore> Repository<T> {
...
+        match result.count {
+            Some(count) => Ok(count > 0),
+            None => {
+                tracing::error!("DB: Failed to check duplicate opportunity: count is None");
+                Err(RestError::TemporarilyUnavailable)
+            }
+        }
+    }
+
+    fn remove_slot_field(&self, mut metadata: serde_json::Value) -> serde_json::Value {
```
danimhr, line 105: Don't remove the slot. It's kind of metadata inside the db and we may need it for future tracing

auction-server/src/opportunity/repository/add_opportunity.rs
```diff
@@ -63,4 +70,42 @@ impl<T: InMemoryStore> Repository<T> {
...
+
+    async fn check_db_duplicate_opportunity(
+        &self,
+        db: &sqlx::Pool<Postgres>,
+        opportunity: &<T as InMemoryStore>::Opportunity,
+    ) -> Result<bool, RestError> {
+        let chain_type = <T::Opportunity as entities::Opportunity>::ModelMetadata::get_chain_type();
+        let metadata = opportunity.get_models_metadata();
+
+        let result = sqlx::query!("SELECT COUNT(*) FROM opportunity WHERE permission_key = $1 AND chain_id = $2 AND chain_type = $3 AND sell_tokens = $4 AND buy_tokens = $5 AND metadata #- '{slot}' = $6",
```
danimhr, line 82: I think we can ignore slot and we don't need to check the slot value.

danimhr: We remove opportunity when it's invalid, or expired.
I think if an opportunity is invalid, then we need to add a new row to the db. When the opportunity is expired, we can update the row and remove the removal time and consider this opportunity as a live opportunity again.

danimhr (commented):
Let me know if we need to chat more about the structure and the algorithm for this task :-?

auction-server/src/opportunity/repository/add_opportunity.rs
```diff
@@ -26,20 +27,100 @@ impl<T: InMemoryStore> Repository<T> {
...
+
+    async fn get_or_add_db_opportunity(
+        &self,
+        db: &sqlx::Pool<Postgres>,
+        opportunity: <T as InMemoryStore>::Opportunity,
+    ) -> Result<T::Opportunity, RestError> {
+        let chain_type = <T::Opportunity as entities::Opportunity>::ModelMetadata::get_chain_type();
         let metadata = opportunity.get_models_metadata();
+
+        let result = sqlx::query!("SELECT id FROM opportunity WHERE permission_key = $1 AND chain_id = $2 AND chain_type = $3 AND sell_tokens = $4 AND buy_tokens = $5 AND metadata #- '{slot}' = $6 AND (removal_reason IS NULL OR removal_reason = 'expired') LIMIT 1",
```
danimhr, line 51: Please pass the removal_reason as variable:

`OR removal_reason = $7`

auction-server/src/opportunity/repository/add_opportunity.rs
```diff
@@ -26,20 +27,100 @@ impl<T: InMemoryStore> Repository<T> {
...
+        opportunity.permission_key.to_vec(),
+        opportunity.chain_id,
+        chain_type as _,
+        serde_json::to_value(&opportunity.sell_tokens).unwrap(),
```
danimhr, line 55: use expect with message instead of unwrap

auction-server/src/opportunity/repository/add_opportunity.rs
```diff
@@ -26,20 +27,100 @@ impl<T: InMemoryStore> Repository<T> {
...
+        serde_json::to_value(&opportunity.buy_tokens).unwrap(),
```
danimhr, line 56: same

auction-server/src/opportunity/repository/add_opportunity.rs
```diff
@@ -26,20 +27,100 @@ impl<T: InMemoryStore> Repository<T> {
...
+            .or_insert_with(Vec::new)
+            .push(opportunity.clone());
+
+        Ok(opportunity)
+    }
+
+    async fn get_or_add_db_opportunity(
+        &self,
+        db: &sqlx::Pool<Postgres>,
+        opportunity: <T as InMemoryStore>::Opportunity,
```
danimhr, line 46: You can pass OpportunityCreate object to this function

auction-server/src/opportunity/repository/add_opportunity.rs
```diff
@@ -26,20 +27,100 @@ impl<T: InMemoryStore> Repository<T> {
...
+        id)
+            .execute(db)
+            .await
+            .map_err(|e| {
+                tracing::error!("DB: Failed to update opportunity: {}", e);
+                RestError::TemporarilyUnavailable
+            })?;
+
+        let mut modified_opportunity = opportunity;
+        modified_opportunity.id = id;
```
danimhr, line 98: maybe better function is `remove_removal_reason`. It gets an opportunity and removes it's removal reason.
In that case we dont need to pass the id and also we dont need to create the modified_opportunity object as well.

Can we use sth like this?
```
UPDATE your_table
SET column_name = new_value
WHERE condition
RETURNING *;
```

And at the end, return the updated value instead of creating the modified opp variable with id assignment?

auction-server/src/opportunity/repository/add_opportunity.rs
```diff
@@ -26,20 +27,100 @@ impl<T: InMemoryStore> Repository<T> {
...
+        Ok(modified_opportunity)
+    }
+
+    async fn add_db_opportunity(
+        &self,
+        db: &sqlx::Pool<Postgres>,
+        opportunity: <T as InMemoryStore>::Opportunity,
```
danimhr, line 105: Pass create_opportunity to this function

danimhr (commented):
I think we should consider do the following as well:
- Add the `last_creation_time` to the opportunity object inside the `models.rs`file.
- Update the remove opportunity loop logic. We should take a look at last_creation_time to mark opps as expired. Currently we are looking at the creation_time.
- Create a new file for each of the new functionality inside the repo module.
- Update the `add_opportunity.rs` file name to `get_or_add_opportunity`.

danimhr (approved):

PR #223, opened 2024-11-08
fix(python): add bid expired to the enum

guibescos:


auction-server/src/auction.rs
```diff
@@ -1583,12 +1583,12 @@ impl ChainStore for ChainStoreEvm {
         tx_hash: Vec<u8>,
     ) -> Result<Option<Vec<<Self::SimulatedBid as SimulatedBidTrait>::StatusType>>> {
         let tx_hash = H256::from_slice(tx_hash.as_slice());
-        let reciept = self
+        let receipt = self
```
guibescos, line 1586: drive-by typo 😉

sdk/python/express_relay/models/base.py
```diff
@@ -19,6 +19,7 @@ class BidStatus(Enum):
     SUBMITTED = "submitted"
     LOST = "lost"
     WON = "won"
+    EXPIRED = "expired"
```
danimhr, line 22: This one is only for svm bids.

PR #226, opened 2024-11-11
Fix get bid status for svm expired

danimhr:
This PR aims to resolve the bug of get_bid_status api for svm bids with expired status


auction-server/src/state.rs
```diff
@@ -945,6 +945,8 @@ impl BidStatusTrait for BidStatusSvm {
                     Ok(BidStatusSvm::Won { result })
                 } else if bid.status == models::BidStatus::Submitted {
                     Ok(BidStatusSvm::Submitted { result })
+                } else if bid.status == models::BidStatus::Expired {
```
m30m, line 948: can we use match without a catch all here? to avoid this in the future?

m30m (approved):

PR #227, opened 2024-11-11
fix: Check svm tx size before simulation

m30m:


auction-server/src/auction.rs
```diff
@@ -1188,6 +1192,19 @@ pub async fn handle_bid_svm(
     Ok(core_fields.id)
 }
 
+fn check_tx_size(transaction: &VersionedTransaction) -> Result<(), RestError> {
+    let tx_serialized = serialize(&transaction)
```
anihamde, line 1196: this is probably fine, but you could also use the built-in serialization: https://docs.rs/solana-program/2.1.0/src/solana_program/message/versions/v0/mod.rs.html#292

anihamde (approved):

auction-server/src/auction.rs
m30m: This would not include the signatures and probably some other headers.

PR #231, opened 2024-11-12
feat: retry logic for svm sending transactions

m30m:
Use tokio broadcast to have multiple subscribers to rpc logs. One for concluding the auctions and the other for stopping resubmission of transactions.

### DIFF
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -119,10 +119,6 @@ use {
             pubsub_client::PubsubClient,
             rpc_client::RpcClient,
         },
-        rpc_config::{
-            RpcTransactionLogsConfig,
-            RpcTransactionLogsFilter,
-        },
         rpc_response::SlotInfo,
     },
     solana_sdk::{
@@ -158,6 +154,7 @@ use {
         Mutex,
         MutexGuard,
     },
+    tokio_util::task::TaskTracker,
     utoipa::ToSchema,
     uuid::Uuid,
 };
@@ -482,7 +479,11 @@ async fn submit_auction_for_bids<'a, T: ChainStore>(
     );
 
     match chain_store
-        .submit_bids(permission_key.clone(), winner_bids.clone())
+        .submit_bids(
+            permission_key.clone(),
+            winner_bids.clone(),
+            store.task_tracker.clone(),
+        )
         .await
     {
         Ok(tx_hash) => {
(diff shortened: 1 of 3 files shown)


auction-server/src/state.rs
```diff
@@ -578,11 +610,86 @@ impl ChainStoreSvm {
...
         &self,
         tx: &VersionedTransaction,
+        task_tracker: TaskTracker,
     ) -> solana_client::client_error::Result<Signature> {
-        self.tx_broadcaster_client.send_transaction(tx).await
+        let config = RpcSendTransactionConfig {
+            skip_preflight:       true,
+            preflight_commitment: None,
+            encoding:             None,
+            max_retries:          None,
```
guibescos, line 662: should we set this to 0 according to the guidelines?

guibescos: I'd just copy this one from Kamino
https://github.com/Kamino-Finance/orbit-link/blob/master/src/async_client/rpc_client.rs#L41

guibescos: finalized commitment is fine since we get the blockhash finalized as per https://docs.triton.one/chains/solana/sending-txs

auction-server/src/state.rs
```diff
@@ -553,6 +582,7 @@ impl ChainStoreSvm {
                 .parse::<usize>()
                 .expect("Failed to parse router account position"),
         };
+        let (tx, rx) = broadcast::channel(1000);
```
guibescos, line 585: I think this doesn't work. If you don't receive the messages in _dummy_log_receiver  they will stay in the channel forever.

`When a value is sent, all [Receiver] handles are notified and will receive the value. The value is stored once inside the channel and cloned on demand for each receiver. Once all receivers have received a clone of the value, the value is released from the channel.`

auction-server/src/auction.rs
```diff
@@ -1758,6 +1762,8 @@ impl ChainStore for ChainStoreSvm {
             tracing::warn!(tx_hash = ?tx_hash, bids = ?bids, "multiple bids found for transaction hash");
         }
 
+
+        //TODO: this can be optimized out if triggered by websocket events
```
guibescos, line 1766: mmm I feel like the `get_signature_status_with_commitment` method is more reliable

guibescos (approved):

PR #233, opened 2024-11-12
fix: type

guibescos:
I think this a type error? Does tsc not type check these files?


m30m:
I also noticed this today. Created a task to fix this

m30m (approved):

PR #234, opened 2024-11-13
feat: better signature checking

m30m:
- Better error message when we know which signature is not valid
- Check that relayer pubkey is actually in the tx as a signer
- Better error if no opportunity was found

### DIFF
--- a/auction-server/src/auction.rs
+++ b/auction-server/src/auction.rs
@@ -1216,7 +1216,7 @@ fn all_signature_exists_svm(
         .any(|account| account.eq(relayer_pubkey));
     if !relayer_exists {
         return Err(RestError::BadParameters(format!(
-            "Relayer account {:?} is missing",
+            "Relayer account {} is not a signer in the transaction",
             relayer_pubkey
         )));
     }
@@ -1226,7 +1226,7 @@ fn all_signature_exists_svm(
         }
         if !signature.verify(pubkey.as_ref(), message_bytes) {
             return Err(RestError::BadParameters(format!(
-                "Signature for account {:?} is invalid",
+                "Signature for account {} is invalid",
                 pubkey
             )));
         }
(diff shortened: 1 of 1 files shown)


auction-server/src/auction.rs
```diff
@@ -1249,21 +1265,20 @@ async fn verify_signatures_svm(
                     ),
                 })
                 .await;
-            opportunities.into_iter().any(|opportunity| {
-                let mut missing_signers = opportunity.get_missing_signers();
-                missing_signers.push(*relayer_pubkey);
-                all_signature_exists_svm(&message_bytes, accounts, &signatures, &missing_signers)
-            })
+            let opportunity = opportunities
+                .first()
```
anihamde, line 1269: i assume you're trying to check the most recent opportunity added? just make sure that first() returns the most recent and not the oldest

auction-server/src/auction.rs
```diff
@@ -1249,21 +1265,20 @@ async fn verify_signatures_svm(
                     ),
                 })
                 .await;
-            opportunities.into_iter().any(|opportunity| {
-                let mut missing_signers = opportunity.get_missing_signers();
-                missing_signers.push(*relayer_pubkey);
-                all_signature_exists_svm(&message_bytes, accounts, &signatures, &missing_signers)
-            })
+            let opportunity = opportunities
```
anihamde, line 1268: just a note: why do we check against existing opportunities? there are some tradeoffs to just simulating the tx and not checking that an opp exists for this PK, but that's how we handle on evm side, i don't recall why we decided on checking against existing opps for svm

anihamde (approved):

auction-server/src/auction.rs
m30m: main reason here is to ignore the other necessary signature from wallet which comes at the end of the flow

auction-server/src/auction.rs
m30m: not really. There should be only one opportunity but the APIs return `Vec`.
@danimhr please correct me if I'm wrong

guibescos: mmm I'm not sure https://github.com/pyth-network/per/blob/main/auction-server/src/opportunity/repository/add_opportunity.rs#L41

m30m: I added some comments here

PR #236, opened 2024-11-13
Feat Dex Router: Filter out too large txs

anihamde:
This PR filters out too large transactions from being submitted via the DEX Router. It also adds a cli arg for specifying multiple max accounts Jupiter queries to run, to generate multiple transactions, each with likely different tx size. This will allow for multiple Jupiter txs to be created and the best one under the max tx size to be submitted.


scripts/dex-router/src/index.ts
```diff
@@ -178,48 +198,54 @@ export class DexRouter {
...
         })
       )
-    ).filter((routerOutput) => routerOutput !== undefined);
-    if (routerOutputs.length === 0) {
+    )
+      .filter((routerInfo) => {
+        if (routerInfo === undefined) {
+          return false;
+        }
+        return routerInfo.tx.serialize().length <= 1232;
```
m30m, line 239: define a constant for this

scripts/dex-router/src/router/jupiter.ts
```diff
@@ -5,17 +5,18 @@ import {
...
+const MAX_SLIPPAGE_BPS = 50;
 
 export class JupiterRouter implements Router {
   private chainId: string;
   private executor: PublicKey;
+  private maxAccounts: number;
   private jupiterClient = createJupiterApiClient();
 
-  constructor(chainId: string, executor: PublicKey) {
+  constructor(chainId: string, executor: PublicKey, maxAccounts: number = 20) {
```
m30m, line 16: I think this is an important variable and should not have a default value so that we force the user to decide about it.

scripts/dex-router/src/index.ts
```diff
@@ -178,48 +198,54 @@ export class DexRouter {
...
+      })
+      .filter((routerInfo) => routerInfo !== undefined);
```
m30m, line 241: do we need this?

scripts/dex-router/src/index.ts
m30m: I think this is not a good place to handle this size limit. Maybe `createRouterTransaction` should throw an exception if it can not a create tansaction within the size limit or the output of it can be possibly null to signify it wasn't successful.

scripts/dex-router/src/index.ts
```diff
@@ -418,6 +444,12 @@ const argv = yargs(hideBin(process.argv))
     type: "array",
     demandOption: false,
   })
+  .option("options-max-accounts-jupiter", {
+    description:
+      "Options for the max number of accounts to include in Jupiter instructions",
+    type: "array",
+    default: [10, 20, 50],
```
m30m, line 451: is 50 realistic? does it help?

scripts/dex-router/src/index.ts
anihamde: yeah for typechecking unfortunately

scripts/dex-router/src/router/jupiter.ts
anihamde: this is fair. 20 is a reasonable value i often found to work, so that's why i wanted to leave it as a default as a hint to the user. otherwise, without context there's no way for a user to know what's reasonable without a bunch of trial and error

m30m: good point

PR #243, opened 2024-11-18
feat: Use priority fees on dex-router

m30m:


scripts/dex-router/package.json
```diff
@@ -14,7 +14,7 @@
     "@coral-xyz/anchor": "^0.30.1",
     "@jup-ag/api": "^6.0.29",
     "@kamino-finance/limo-sdk": "^0.7.0",
-    "@pythnetwork/express-relay-js": "^0.14.0",
+    "@pythnetwork/express-relay-js": "^0.14.1",
```
anihamde, line 17: this should be 0.14.2 now with the other bump?

scripts/dex-router/package.json
```diff
@@ -14,7 +14,7 @@
     "@coral-xyz/anchor": "^0.30.1",
```
anihamde, line 14: also let's start bumping the versions on this. we should start to stick to some standard, so if a PR is a backward incompatible feature addition, we should increment the minor version pre-release. here i think it's not, so can probs just patch bump

scripts/dex-router/src/index.ts
```diff
@@ -321,18 +322,24 @@ export class DexRouter {
...
+    if (!this.latestChainUpdate[this.chainId]) {
       console.log(
         `No recent blockhash for chain ${this.chainId}, getting manually`
       );
-      this.recentBlockhash[this.chainId] = (
+      this.latestChainUpdate[this.chainId] = (
         await this.connectionSvm.getLatestBlockhash("confirmed")
       ).blockhash;
+    } else {
+      const feeInstruction = ComputeBudgetProgram.setComputeUnitPrice({
```
anihamde, line 333: we are handling the compute budget instruction somewhere else in the code via what jupiter gives us and with the filterComputeBudget call, pls remove the use of the jupiter compute budget ix and modify the filter ix

anihamde: i think it's fine to keep the filter ix as a way to cap against server sending outrageous priority fees. you can also add this fee instruction construction to the file `computeBudget.ts` since that already exists and would be a better place for it to live

anihamde: another thought: you could get the jupiter compute budget ixs and filter out the set price ix for this one. that way, you could still have a set compute unit limit ix. not necessary, feel free to ignore this

PR #246, opened 2024-11-19
fix: better rpc health monitoring

anihamde:
This PR adds a check to exclude alerts when a block doesn't exist for a slot.

### DIFF
--- a/scripts/dex-router/package-lock.json
+++ b/scripts/dex-router/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dex-router",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "dex-router",
-      "version": "0.0.2",
+      "version": "0.0.3",
       "license": "Apache-2.0",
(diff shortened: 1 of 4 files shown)


scripts/dex-router/src/utils/health.ts
```diff
@@ -22,7 +22,13 @@ export async function checkRpcHealth(
         );
       }
     } catch (e) {
-      console.error("Health Error (RPC endpoint), failure to fetch: ", e);
+      if (e instanceof Error) {
+        if (!e.message.includes("Block not available for slot")) {
+          console.error("Health Error (RPC endpoint), failure to fetch: ", e);
```
m30m, line 27: duplicate logic, I guess you can combine the ifs here

scripts/limonade/src/index.ts
```diff
@@ -182,7 +182,13 @@ async function run() {
           );
         }
       } catch (e) {
-        console.error("Health Error (RPC endpoint), failure to fetch: ", e);
+        if (e instanceof Error) {
```
m30m, line 185: ditto

m30m (approved):

PR #254, opened 2024-11-21
Fix/bump pkg lock dex router

anihamde:


package-lock.json
```diff
@@ -14,7 +14,7 @@
       },
       "engines": {
         "node": "^18.20.2",
-        "npm": "^10.9.0"
+        "npm": "^10.7.0"
```
m30m, line 17: let's not commit this or also update package.json to represent the same version

m30m (approved):

package-lock.json
anihamde: an earlier PR made the versions in package-lock and package different, this is correcting them

PR #257, opened 2024-11-22
bump: Bump sdks

m30m:
- Increase `SEND_TRANSACTION_RETRY_COUNT_SVM` to 30
- Use Kamino sdk 0.7.1
- Bump sdk versions to include the new bid status

### DIFF
--- a/sdk/js/package.json
+++ b/sdk/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/express-relay-js",
-  "version": "0.14.3",
+  "version": "0.15.0",
   "description": "Utilities for interacting with the express relay protocol",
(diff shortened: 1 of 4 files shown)


sdk/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/express-relay-js",
-  "version": "0.14.3",
+  "version": "0.15.0",
```
danimhr, line 3: I'm not sure about best practice here
Every time limo update, are we gonna increase the version like this? :-?

danimhr (approved):

sdk/js/package.json
m30m: I did this because of the introduction of new bid status and not because of limo

guibescos (approved):

PR #259, opened 2024-11-25
fix: add another condition for filled orders

m30m:


scripts/limonade/src/index.ts
```diff
@@ -50,6 +50,13 @@ const argv = yargs(hideBin(process.argv))
   .alias("help", "h")
   .parseSync();
 
+function isCompletelyFilled(order: Order): boolean {
```
anihamde, line 53: in this case, should we not expose this as an opportunity? what about the example where I create an order with X input amount and Y min output amount, and someone fills me for Y but only takes Z << X input tokens, leaving X-Z >> 0 in the order. Should I still get filled for more, or should I be returned my remaining input tokens?

anihamde: in any case, this seems like a fix on the limo program side. this order is still live as of the current program rules (the input tokens haven't been returned/the order account hasn't been closed), so i don't think we should filter this out from valid opportunity set.

anihamde: The only case where I see the validity of this is when there's just dust remaining in this account and no one is going to bite on it. But I think that should be a different filter (e.g. if someone makes a dust order in the first place, we shoudl filter that out to prevent spamming opps)

PR #260, opened 2024-11-25
Fix/improve dex router

anihamde:
This PR updates the DEX Router and Limonade scripts to the most recent versions of the Express Relay and Limo SDKs. It also improves in a couple ways the DEX Router:

- Fixes an issue with Jupiter's endpoint returning flat microLamportsPrice, uses priority fees from the server
- Makes the Jupiter endpoint specifiable via CLI
- Adds documentation on the return types of methods
- Turns latency to 0 to mitigate chances of landing tx with failure


scripts/dex-router/src/const.ts
```diff
@@ -1,6 +1,6 @@
 // Time to wait in milliseconds before submitting a router bid on opportunity
 // We introduce this wait time, since this router bid is a backup to searchers normally fulfilling
-export const OPPORTUNITY_WAIT_TIME_MS = 800;
+export const OPPORTUNITY_WAIT_TIME_MS = 0;
```
m30m, line 3: let's remove this logic completely. We can always look in git commit history and put it back if needed

scripts/dex-router/src/router/jupiter.ts
```diff
@@ -2,21 +2,36 @@ import { Router, RouterOutput } from "../types";
...
+    executor: PublicKey,
+    maxAccounts: number,
+    apiKey?: string
+  ) {
     this.chainId = chainId;
     this.executor = executor;
     this.maxAccounts = maxAccounts;
+    if (apiKey) {
+      this.jupiterClient = createJupiterApiClient({
+        basePath: BASE_PATH_QUICKNODE,
```
m30m, line 29: this is a bit inflexible in case we want to use another provider or quicknode changes the domain format. Maybe we can use `URL.parse` to separate the basePath and apiKey?

PR #261, opened 2024-11-26
Add more fields to tracing

danimhr:
This PR aims to add more fields for tracing

### DIFF
--- a/auction-server/src/auction/entities/bid.rs
+++ b/auction-server/src/auction/entities/bid.rs
@@ -170,7 +170,7 @@ pub type PermissionKey<T> = <<T as ChainTrait>::BidChainDataType as BidChainData
 pub type TxHash<T> = <<T as ChainTrait>::BidStatusType as BidStatus>::TxHash;
 
 pub trait BidChainData: Send + Sync + Clone + Debug + PartialEq {
-    type PermissionKey: Send + Sync + Debug + Hash + Eq + Clone;
+    type PermissionKey: Send + Sync + Debug + Hash + Eq + Clone + Debug;
 
     fn get_permission_key(&self) -> Self::PermissionKey;
 }
--- a/auction-server/src/auction/entities/bid.rs
+++ b/auction-server/src/auction/entities/bid.rs
@@ -32,7 +32,11 @@ use {
         transaction::VersionedTransaction,
     },
     std::{
-        fmt::Debug,
+        fmt::{
+            Debug,
+            Display,
+            Formatter,
+        },
         hash::Hash,
     },
     time::OffsetDateTime,
@@ -267,5 +271,19 @@ impl From<(Bid<Evm>, bool)> for MulticallData {
             gas_limit: bid.chain_data.gas_limit,
             revert_on_failure,
         }
+    }
+}
+
+pub struct BidContainerTracing<'a, T: ChainTrait>(pub &'a [Bid<T>]);
+impl<T: ChainTrait> Display for BidContainerTracing<'_, T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{:?}",
+            self.0
+                .iter()
+                .map(|x| x.id.to_string())
+                .collect::<Vec<String>>()
+        )
     }
 }
--- a/auction-server/src/auction/repository/submit_auction.rs
+++ b/auction-server/src/auction/repository/submit_auction.rs
@@ -14,12 +14,15 @@ use {
 };
 
 impl<T: ChainTrait> Repository<T> {
-    #[tracing::instrument(skip_all)]
+    #[tracing::instrument(skip_all, name = "submit_auction_repo", fields(auction_id, tx_hash))]
     pub async fn submit_auction(
         &self,
         auction: entities::Auction<T>,
         transaction_hash: entities::TxHash<T>,
     ) -> anyhow::Result<entities::Auction<T>> {
+        tracing::Span::current().record("auction_id", auction.id.to_string());
+        tracing::Span::current().record("tx_hash", format!("{:?}", transaction_hash));
+
         let mut auction = auction.clone();
         let now = OffsetDateTime::now_utc();
         auction.tx_hash = Some(transaction_hash.clone());
--- a/auction-server/src/auction/service/auction_manager.rs
+++ b/auction-server/src/auction/service/auction_manager.rs
@@ -48,10 +48,7 @@ use {
             Signature,
             Signer,
         },
-        transaction::{
-            TransactionError,
-            VersionedTransaction,
-        },
+        transaction::TransactionError,
     },
     std::{
         fmt::Debug,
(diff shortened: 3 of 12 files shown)


auction-server/src/auction/repository/submit_auction.rs
```diff
@@ -14,12 +14,15 @@ use {
...
 impl<T: ChainTrait> Repository<T> {
-    #[tracing::instrument(skip_all)]
+    #[tracing::instrument(skip_all, name = "submit_auction_repo", fields(auction_id, tx_hash))]
     pub async fn submit_auction(
         &self,
         auction: entities::Auction<T>,
         transaction_hash: entities::TxHash<T>,
     ) -> anyhow::Result<entities::Auction<T>> {
+        tracing::Span::current().record("auction_id", auction.id.to_string());
+        tracing::Span::current().record("tx_hash", format!("{:?}", transaction_hash));
```
m30m, line 24: is this a reasonable format? From what I remember, it outputs an array of u8 which is hard to parse

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -180,6 +188,8 @@ impl AuctionManager<Evm> for Service<Evm> {
             )
             .await?;
 
+        tracing::Span::current().record("simulation_result", format!("{:?}", simulation_result));
```
m30m, line 191: it seems that the default span attribute value length limit is around 2048 bytes. Maybe we should use a more compact representation here?

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -211,24 +221,40 @@ impl AuctionManager<Evm> for Service<Evm> {
...
             .chain_config
             .provider
             .get_transaction_receipt(bid_status_auction.tx_hash)
             .await
             .map_err(|e| anyhow::anyhow!("Failed to get transaction receipt: {:?}", e))?;
+
         match receipt {
             Some(receipt) => {
                 let decoded_logs = Self::decode_logs_for_receipt(&receipt);
+                tracing::Span::current().record("result", format!("{:?}", decoded_logs));
```
m30m, line 257: it seems that the default span attribute value length limit is around 2048 bytes. Maybe we should use a more compact representation here? Maybe just a true or false on tx failure?

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -370,20 +410,26 @@ impl AuctionManager<Svm> for Service<Svm> {
...
                 // Either simulation was successful or we can't simulate at this moment
-                _ => return Ok(vec![bid.clone()]),
+                _ => {
+                    tracing::Span::current().record("simulation_result", format!("{:?}", result));
+                    return Ok(vec![bid.clone()]);
+                }
             }
         }
+
+        tracing::Span::current().record("simulation_result", format!("{:?}", results));
```
m30m, line 428: it seems that the default span attribute value length limit is around 2048 bytes. Maybe we should use a more compact representation here?

auction-server/src/auction/service/handle_auction.rs
```diff
@@ -24,18 +24,42 @@ impl<T: ChainTrait> Service<T>
...
+    #[tracing::instrument(skip_all, fields(auction_id, bid_ids, winner_bid_ids))]
     async fn submit_auction<'a>(
         &self,
         auction: entities::Auction<T>,
         _auction_mutex_gaurd: MutexGuard<'a, ()>,
     ) -> anyhow::Result<()> {
+        tracing::Span::current().record("auction_id", auction.id.to_string());
+        tracing::Span::current().record(
+            "bid_ids",
+            format!(
```
m30m, line 36: this seems like a lot of work to do each time. How about something like this:
```rs
tracing::Span::current().record(
            "bid_ids",
            tracing::field::display(BidContainerTracing(&bids))
        );

/// put these in bid probably
        
pub struct BidContainerTracing<'a, T: ChainTrait>(pub &'a [Bid<T>]);

impl<T: ChainTrait> Display for BidContainerTracing<'_, T> {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        std::fmt::Display::fmt(&self.0.iter().map(|x| x.id.to_string()).collect::<Vec<String>>(), f)
    }
}
```

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -522,10 +585,47 @@ impl Service<Svm> {
...
+    async fn blocking_send_transaction(&self, bid: entities::Bid<Svm>, signature: Signature) {
+        tracing::Span::current().record("bid_id", bid.id.to_string());
+        tracing::Span::current().record("tx_hash", signature.to_string());
+        let config = RpcSendTransactionConfig {
+            skip_preflight: true,
+            max_retries: Some(0),
+            ..RpcSendTransactionConfig::default()
+        };
+        let mut receiver = self.config.chain_config.log_sender.subscribe();
+        let signature_bs58 = bs58::encode(signature).into_string();
```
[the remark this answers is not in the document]
danimhr: Yes, it's doing the same thing :))

auction-server/src/auction/service/auction_manager.rs
danimhr: I tested it with a string with 1million chars. It worked :-?

auction-server/src/auction/service/auction_manager.rs
danimhr: I tested it with a string with 1million chars. It worked :-?

auction-server/src/auction/repository/submit_auction.rs
danimhr: Maybe this issue was with permission_key?

auction-server/src/auction/service/auction_manager.rs
m30m: this is specific for tempo which we use in staging and prod.
https://grafana.com/docs/tempo/latest/configuration/#ingestion-limits
`max_label_value_length`

I don't think it's good practice to store big values here in general.

danimhr: I think 2048 bytes is more than enough for these logs :-?

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -522,10 +562,46 @@ impl Service<Svm> {
...
+        };
+        let mut receiver = self.config.chain_config.log_sender.subscribe();
+        for retry_count in 0..SEND_TRANSACTION_RETRY_COUNT_SVM {
+            tokio::time::sleep(Duration::from_secs(2)).await;
+
+            // Do not wait for the logs to be received
+            // just check if the transaction is in the logs already
+            while let Ok(log) = receiver.try_recv() {
+                if log.value.signature.eq(&signature.to_string()) {
+                    tracing::Span::current().record("total_tries", retry_count + 1);
```
m30m, line 582: This is called `total_tires` but the field on top of the function is named `total_retries`.

Can we have this field in case of complete failure too? I know that we can assume if the span doesn't have this field, it means it retried the max amount but I guess it's just easier the other way.

m30m (approved):

auction-server/src/auction/service/auction_manager.rs
danimhr: Good catch

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -522,10 +562,48 @@ impl Service<Svm> {
             relayer.sign_message(&serialized_message);
     }
 
+    #[tracing::instrument(skip_all, fields(bid_id, total_tries, tx_hash))]
+    async fn blocking_send_transaction(&self, bid: entities::Bid<Svm>, signature: Signature) {
+        tracing::Span::current().record("bid_id", bid.id.to_string());
+        tracing::Span::current().record("tx_hash", signature.to_string());
+        let config = RpcSendTransactionConfig {
```
guibescos, line 569: duplicating `RpcSendTransactionConfig` doesn't seem great

auction-server/src/auction/entities/bid.rs
```diff
@@ -170,7 +174,7 @@ pub type PermissionKey<T> = <<T as ChainTrait>::BidChainDataType as BidChainData
 pub type TxHash<T> = <<T as ChainTrait>::BidStatusType as BidStatus>::TxHash;
 
 pub trait BidChainData: Send + Sync + Clone + Debug + PartialEq {
-    type PermissionKey: Send + Sync + Debug + Hash + Eq + Clone;
+    type PermissionKey: Send + Sync + Debug + Hash + Eq + Clone + Debug;
```
guibescos, line 177: debug is twice here

guibescos (commented):
found some nits

m30m:
fixed these in my PR

PR #262, opened 2024-11-26
feat: better error logs

guibescos:
- bubble up status codes so we can filter on 5xx in the logs
- add ping/pong to the websocket to ensure we notice it the connection gets interrupted

### DIFF
--- a/sdk/js/src/index.ts
+++ b/sdk/js/src/index.ts
@@ -37,8 +37,8 @@ export * from "./types";
 export * from "./const";
 
 export class ClientError extends Error {
-  static newHttpError(error: string, status: number) {
-    const message = `Auction server http error ${status} - ${error}`;
+  static newHttpError(error: string, status?: number) {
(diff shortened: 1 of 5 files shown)


sdk/js/src/index.ts
```diff
@@ -48,10 +58,12 @@ export interface WsOptions {
    * Max time to wait for a response from the server in milliseconds
    */
   response_timeout: number;
+  ping_interval: number;
 }
 
 const DEFAULT_WS_OPTIONS: WsOptions = {
   response_timeout: 10000,
+  ping_interval: 30000,
```
m30m, line 66: I don't think you need to actively ping from client side. Server pings every subscriber, so you can make sure you are receiving a ping from server side every 30 seconds. You can look at swap-beacon for an example implementation

guibescos: I am embarrassed I didn't see this

m30m (approved):

PR #264, opened 2024-11-28
limonade: bump express-relay sdk and remove direct http checks

guibescos:
- remove direct `/live` checks
- send all `submitOpportunity` calls concurrently
- filter out duplicate opportunity errors to not pollute stderr
- handle rejects in the callback inside `onProgramAccountChange`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,5 +2,11 @@ module.exports = {
   root: true,
   parser: "@typescript-eslint/parser",
   plugins: ["@typescript-eslint"],
+  parserOptions: {
+    project: "./tsconfig.json",
+  },
+  rules: {
+    "@typescript-eslint/no-misused-promises": "error",
+  },
   extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
 };
--- a/scripts/limonade/src/index.ts
+++ b/scripts/limonade/src/index.ts
@@ -101,20 +101,22 @@ async function run() {
       );
 
     console.log("Resubmitting opportunities", payloads.length);
-    for (const payload of payloads) {
-      try {
-        await client.submitOpportunity(payload);
-      } catch (e) {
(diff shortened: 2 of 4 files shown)


.eslintrc.js
```diff
@@ -2,5 +2,11 @@ module.exports = {
   root: true,
   parser: "@typescript-eslint/parser",
   plugins: ["@typescript-eslint"],
+  parserOptions: {
+    project: "./tsconfig.json",
+  },
+  rules: {
+    "@typescript-eslint/no-misused-promises": "error",
```
guibescos, line 9: The promise inside `onProgramAccounts` wasn't handled, and this rules would've caught

scripts/limonade/src/index.ts
```diff
@@ -100,55 +101,67 @@ async function run() {
...
+    await Promise.all(
+      payloads.map(async (payload) => {
+        try {
+          await client.submitOpportunity(payload);
+        } catch (e) {
+          if (
+            e instanceof ClientError &&
+            e.message.includes("Same opportunity is submitted recently")
+          ) {
+            console.log(e); // We don't want to pollute stderr with this
```
m30m, line 113: I'm even ok with not polluting the stdout :D

scripts/limonade/src/index.ts
```diff
@@ -100,55 +101,67 @@ async function run() {
...
 
     console.log("Resubmitting opportunities", payloads.length);
-    for (const payload of payloads) {
-      try {
-        await client.submitOpportunity(payload);
-      } catch (e) {
-        console.error(e);
-      }
-    }
+    await Promise.all(
```
m30m, line 104: there should be some batching mechanism here. It's not practical to send 1000 requests at the same time + we will probably get rate-limited on prod

m30m (approved):

PR #270, opened 2024-11-29
Add pnpm to per

danimhr:
This PR aims to add pnpm to per


.github/workflows/ci-pre-commit.yml
```diff
@@ -39,8 +39,12 @@ jobs:
...
         working-directory: contracts/evm
         run: npm ci
-      - name: Install npm dependencies
-        run: npm ci
+      - uses: pnpm/action-setup@v4
+        name: Install pnpm
+        with:
+          run_install: true
+      - name: Install dependencies
+        run: pnpm install
```
cprussin, line 47: This isn't necessary, it's covered by the `run_install` on line 45 above

.github/workflows/publish-js-sdks.yml
```diff
@@ -17,20 +17,26 @@ jobs:
...
 
+      - uses: pnpm/action-setup@v4
+        name: Install pnpm
+        with:
+          run_install: true
+
       - name: Install dependencies
-        run: npm install
-        working-directory: sdk/js/
+        run: pnpm --filter @pythnetwork/express-relay-js install
```
cprussin, line 26: Again this is redundant with line 23 above (I generally recommend against a scoped install, it might seem like it would be an optimization but it actually usually just makes it harder to do caching well in CI and doesn't usually buy you very much even in the best case)

.github/workflows/publish-js-sdks.yml
```diff
@@ -17,20 +17,26 @@ jobs:
...
 
       - name: Run tests
-        run: npm test # Ensure your tests pass before publishing
-        working-directory: sdk/js/
+        run: pnpm --filter @pythnetwork/express-relay-js test # Ensure your tests pass before publishing
```
cprussin, line 29: I'd get rid of the `--filter` and use `-r` instead; that way pnpm will just run `test` on any package it finds in the workspace which seems more sementically intentional

.github/workflows/publish-js-sdks.yml
```diff
@@ -17,20 +17,26 @@ jobs:
...
+
+      - name: Set publishing config
+        run: pnpm config set '//registry.npmjs.org/:_authToken' "${NODE_AUTH_TOKEN}"
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
       - name: Build package
-        run: npm run build # If your package requires a build step
-        working-directory: sdk/js/
+        run: pnpm --filter @pythnetwork/express-relay-js build # If your package requires a build step
```
cprussin, line 37: Same thing, I'd just run `pnpm -r build` here and `pnpm -r publish`, so you can have new packages hook into this CI workflow just by defining the expected lifecycle hooks

.github/workflows/publish-solidity-sdk.yml
```diff
@@ -17,25 +17,31 @@ jobs:
           node-version-file: "package.json"
           registry-url: "https://registry.npmjs.org"
```
cprussin, line 19: Unless I'm misunderstanding, if you follow my suggestions above, you can get rid of this workflow entirely.

.github/workflows/publish-solidity-sdk.yml
```diff
@@ -17,25 +17,31 @@ jobs:
...
 
       - name: Install dependencies
-        run: npm install
-        working-directory: sdk/solidity/
+        run: pnpm --filter @pythnetwork/express-relay-sdk-solidity install
 
       - name: Generate ABIs
-        run: npm run generate-abi
-        working-directory: sdk/solidity/
+        run: pnpm generate-abi
```
cprussin, line 34: For this kind of thing I generally make the package that requires it have scripts that look like this:

```
"scripts": {
  "build": "pnpm build:abi && pnpm build:tsc",
  "build:abi": "...",
  "build:tsc": "tsc"
}
```

Of course in the crosschain repo turborepo takes care of the `build` script.

This way you don't have to configure CI workflows to be aware of the steps required to build individual packages.

examples/easy_lend/package.json
```diff
@@ -21,21 +21,20 @@
...
+    "@pythnetwork/express-relay-js": "workspace:*",
+    "@pythnetwork/express-relay-sdk-solidity": "workspace:*",
     "@pythnetwork/price-service-client": "^1.9.0",
     "@pythnetwork/pyth-evm-js": "^1.73.0",
     "@pythnetwork/pyth-sdk-solidity": "^4.0.0",
     "ts-node": "^10.9.1",
-    "typescript": "^5.3.3",
-    "viem": "^2.7.6"
+    "typescript": "catalog:",
+    "viem": "catalog:"
```
cprussin, line 31: :tada:

package.json
```diff
@@ -1,14 +1,17 @@
...
     "prettier": "^3.1.1",
-    "prettier-plugin-solidity": "^1.3.1",
-    "typescript": "^5.5.4",
-    "@typescript-eslint/eslint-plugin": "^6.21.0",
-    "@typescript-eslint/parser": "^6.21.0",
-    "eslint": "^8.57.0"
-  }
+    "eslint": "^8.57.1",
+    "@typescript-eslint/eslint-plugin": "^8.16.0",
+    "@typescript-eslint/parser": "^8.16.0"
```
cprussin, line 14: I generally recommend managing tooling like prettier and eslint on each package and not at the root; this way when you have some package that has some dependency that prevents you from upgrading you can hold that package back without affecting the whole repo.

Pnpm catalogs are super useful for this kind of thing.

scripts/dex-router/Dockerfile
```diff
@@ -56,10 +46,7 @@ ENV NODE_ENV production
...
 
-# Copy package.json so that package manager commands can be used.
-COPY package.json .
-
 # Copy the production dependencies from the deps stage and also
 # the built application from the build stage into the image.
-COPY --from=deps /usr/src/app/node_modules ./node_modules
-COPY --from=build /usr/src/app/dist ./dist
+COPY --from=build /prod/dex-router/node_modules ./node_modules
+COPY --from=build /usr/src/app/scripts/dex-router/dist ./dist
```
cprussin, line 52: I did a few things in `Dockerfile.node` to optimize stuff (https://github.com/pyth-network/pyth-crosschain/blob/main/Dockerfile.node) that you might want to duplicate:

1. Base image shas are locked, preventing issues where some tag floated and you upgrade dependencies unintentionally (these can be very painful to debug)
2. The build image and the runner images are separated; a large image is used for the build which prevents some issues with some packages that Alpine has with building some gyp-based dependencies, these can again be very painful to debug
3. Corepack is used to manage the pnpm version, so you don't need to worry about changing the PNPM_VERSION to match what you're using in dev

Honestly you can probably just use the exact same dockerfile if you wanted

scripts/limonade/Dockerfile
```diff
@@ -56,10 +46,7 @@ ENV NODE_ENV production
...
 
-# Copy package.json so that package manager commands can be used.
-COPY package.json .
-
 # Copy the production dependencies from the deps stage and also
 # the built application from the build stage into the image.
-COPY --from=deps /usr/src/app/node_modules ./node_modules
-COPY --from=build /usr/src/app/dist ./dist
+COPY --from=build /prod/limonade/node_modules ./node_modules
+COPY --from=build /usr/src/app/scripts/limonade/dist ./dist
```
cprussin, line 52: If you create a parameterized dockerfile then you wouldn't have to duplicate this, again see https://github.com/pyth-network/pyth-crosschain/blob/main/Dockerfile.node and how I run it with e.g. https://github.com/pyth-network/pyth-crosschain/blob/main/.github/workflows/push-xc-admin-images.yml#L31

cprussin (approved):

.github/workflows/publish-js-sdks.yml
danimhr: I want to exactly publish the `@pythnetwork/express-relay-js` pacakge. Why should I build and publish all of them then?

.github/workflows/publish-solidity-sdk.yml
danimhr: We aim to to have different tags and different workflows for publishing separate packages. Also we need to publish only solidity sdk and js sdk. But there are bunch of other packages here as well. How to achieve this?

package.json
danimhr: I just put this here to make sure every internal package in the per repo, follows same eslint rules. I agree that later it can become harder to have same eslint rules for all packages.

scripts/dex-router/Dockerfile
danimhr: As I understand, I think the difference here is that you are copying the whole monorepo to build packages. I'm not sure if I wanna do same thing for building the packages :-?

scripts/limonade/Dockerfile
danimhr: Tnx for this comment. Now using a shared dockerfile for node services

.github/workflows/publish-solidity-sdk.yml
danimhr: For now I just built all packages using `pnpm -r build` :D

.github/workflows/publish-solidity-sdk.yml
cprussin: This makes more sense!

Eventually I'll move the monorepo to a proper release management tool, e.g. I've used https://github.com/changesets/changesets in the past and I find it quite nice and there are a few other similar things out there.  For now I don't have any better suggestions than what you're doing if you want to release packages independently.

scripts/dex-router/Dockerfile
cprussin: ahh yeah that part isn't well optimized, I was more focused on optimizing the output image size and avoiding build issues from Node dependencies that can be painful to deal with than I was focused on improving the build time.  I think you could probably combine some of the optimizations above with also only copying specific packages into the build image...

package.json
cprussin: makes sense -- if you decide to eject what I usually do is have a common eslint config so each package can just reference that config, and use catalogs to manage the eslint & config dependencies.  Then you achieve all packages having shared eslint config, but can still easily exit out for specific places when you have a good reason to do so.

I follow this same pattern for all tools I use fwiw, including prettier, typescript, eslint, jest, etc.

examples/easy_lend/package.json
```diff
@@ -21,21 +21,20 @@
...
   },
   "devDependencies": {
-    "@types/yargs": "^17.0.10",
-    "eslint": "^8.56.0",
+    "@types/yargs": "catalog:",
     "prettier": "^2.6.2",
     "typedoc": "^0.25.7",
-    "typescript": "^5.1",
-    "yargs": "^17.4.1"
+    "typescript": "cataglog:",
```
m30m, line 37: typo?

m30m (approved):

PR #272, opened 2024-12-02
Fix/node build

danimhr:
This PR aims to fix the build for node services

### DIFF
--- a/.github/workflows/push-services.yaml
+++ b/.github/workflows/push-services.yaml
@@ -5,7 +5,6 @@ on:
       - v*
     branches:
       - main
-      - fix/node-build
   workflow_dispatch:
     inputs:
       dispatch_description:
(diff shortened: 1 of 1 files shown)


.github/workflows/push-services.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
           push: true
           tags: ${{ steps.meta_monitor.outputs.tags }}
           labels: ${{ steps.meta_monitor.outputs.labels }}
-  build-and-push-node-services:
```
anihamde, line 73: what was the issue with this before?

anihamde (approved):

.github/workflows/push-services.yaml
danimhr: the quotations

PR #275, opened 2024-12-02
do it

guibescos:


sdk/js/src/index.ts
```diff
@@ -148,6 +151,11 @@ export class Client {
     this.websocketBidStatusCallback = bidStatusCallback;
     this.websocketSvmChainUpdateCallback = svmChainUpdateCallback;
     this.websocketRemoveOpportunitiesCallback = removeOpportunitiesCallback;
+    this.websocketCloseCallback =
+      websocketCloseCallback ??
```
m30m, line 155: why not make it optional similar to other callbacks?

guibescos: It's optional but defaults to throwing an error

sdk/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/express-relay-js",
-  "version": "0.15.3",
```
m30m, line 3: seems minor to me.

m30m (approved):

sdk/js/package.json
guibescos: done

PR #276, opened 2024-12-02
feat: simple searcher for testing the auction server configs

m30m:
another simple client just to test that the auction server is working correctly and submitting bids, without limo or any other programs involved.


guibescos (approved):
works for me

PR #278, opened 2024-12-02
chore: add websocket timeout to limonade

guibescos:
We might want to make limonade crash if the websocket is not working (since that's the lowest latency path for opportunities to be published).
This PR adds a timeout for the websocket leveraging the onSlotChange subscription. If there are not new slots for 30 seconds, we will assume the websocket doesn't work and crash.

### DIFF
--- a/scripts/limonade/src/index.ts
+++ b/scripts/limonade/src/index.ts
@@ -52,6 +52,11 @@ const argv = yargs(hideBin(process.argv))
     type: "number",
     default: 100,
   })
+  .option("solana-websocket-timeout", {
+    description: "Solana websocket timeout (milliseconds)",
+    type: "number",
+    default: 10 * 1000,
+  })
   .help()
   .alias("help", "h")
   .parseSync();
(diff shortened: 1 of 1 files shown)


scripts/limonade/src/index.ts
```diff
@@ -180,6 +187,16 @@ async function run() {
       filters,
     }
   );
+  connection.onSlotChange(() => {
+    if (solanaConnectionTimeout !== undefined) {
+      clearTimeout(solanaConnectionTimeout);
+    }
+
+    solanaConnectionTimeout = setTimeout(() => {
+      throw new Error("Solana websocket timeout");
```
anihamde, line 196: maybe don't crash, just create a log for alerts? idk what the best practice is, but if it's just a momentary issue, maybe it's better to keep it alive? maybe this is bad hygeine

anihamde (approved):

scripts/limonade/src/index.ts
guibescos: yeah I'm on kind of on the fence too

guibescos: I think it's good because on restart it will into `submitExistingOpportunities` with the updates that we missed for 30 seconds

PR #281, opened 2024-12-04
More profiling + small bugfix

m30m:


guibescos (approved):
lgtm

auction-server/src/auction/service/handle_bid.rs
```diff
@@ -21,16 +21,25 @@ impl<T: ChainTrait> Service<T>
...
+        if let Some(profile) = &input.bid_create.profile {
+            tracing::Span::current().record("profile_name", &profile.name);
+        }
+        let verification = self
             .verify_bid(VerifyBidInput {
                 bid_create: input.bid_create.clone(),
             })
-            .await?;
+            .await;
+        if let Err(RestError::SimulationError { result: _, reason }) = &verification {
```
guibescos, line 37: Interesting that you can get the `Err` without consuming the `Result`

PR #283, opened 2024-12-05
feat: reduce rpc calls in python sdk

m30m:


sdk/python/express_relay/svm/limo_client.py
```diff
@@ -111,82 +107,72 @@ def create_associated_token_account_idempotent(
...
     ) -> WSOLInstructions:
         """
         Returns necessary instructions to create, fill and close a wrapped SOL account.
-        If the account already exists:
-            it makes sure the balance is at least `amount_to_deposit_lamports` with no create or close instructions.
-        If the account does not exist:
-            it creates the account, fills it with `amount_to_deposit_lamports` lamports and close it in the end.
+        Creation instruction is idempotent.
+        Filling instruction doesn't take into account the current WSOL balance.
+        Closing instruction unwraps all the WSOL back to the owner, even if it was deposited by another transaction.
```
anihamde, line 130: nit: "even if it was deposited by another transaction" is a bit confusing, i think you can just say the WSOL token account is always closed and the SOL is unwrapped back to the owner

sdk/python/express_relay/svm/limo_client.py
```diff
@@ -111,82 +107,72 @@ def create_associated_token_account_idempotent(
...
         payer: Pubkey,
     ) -> Tuple[Pubkey, Sequence[Instruction]]:
         ata = self.get_ata(owner, token_mint_address, token_program_id)
-        if not await self.account_exists(ata):
-            ix = self.create_associated_token_account_idempotent(
-                payer, owner, token_mint_address, token_program_id
-            )
-            return ata, [ix]
-        return ata, []
+        ix = self.create_associated_token_account_idempotent(
```
anihamde, line 118: i think this is fine, just double checking that limo sdk does idempotent for general ATAs too?

anihamde (approved):

sdk/python/express_relay/svm/limo_client.py
m30m: it does, I cross checked

PR #288, opened 2024-12-06
fix: docstrings

guibescos:
Update python simple searcher docstring and function signature to be consistent with JS


sdk/python/express_relay/searcher/examples/simple_searcher_svm.py
```diff
@@ -127,14 +127,15 @@ async def get_mint_decimals(self, mint: Pubkey) -> int:
             ] = await self.limo_client.get_mint_decimals(mint)
         return self.mint_decimals_cache[str(mint)]
 
-    async def assess_opportunity(self, opp: OpportunitySvm) -> BidSvm | None:
+    async def generate_bid(self, opp: OpportunitySvm) -> BidSvm:
```
guibescos, line 130: rename to `generate_bid` since it doesn't assess the opportunity really

anihamde (approved):

PR #289, opened 2024-12-06
nit: make owner pay always for wsol ata creation

guibescos:
`get_init_if_needed_wsol_create_and_close_ixs` always gets called with `owner == payer` and it doesn't make a lot of sense to call it with `owner != payer` so I propose merging the two args

nit from https://github.com/pyth-network/per/pull/283

### DIFF
--- a/sdk/python/express_relay/svm/limo_client.py
+++ b/sdk/python/express_relay/svm/limo_client.py
@@ -196,7 +196,7 @@ class LimoClient:
         taker_input_ata: Pubkey
         if order["state"].input_mint == WRAPPED_SOL_MINT:
             instructions = self.get_init_if_needed_wsol_create_and_close_ixs(
-                owner=taker, payer=taker, amount_to_deposit_lamports=0
+                owner=taker, amount_to_deposit_lamports=0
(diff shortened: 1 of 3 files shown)


sdk/python/express_relay/svm/limo_client.py
```diff
@@ -348,7 +347,10 @@ def get_event_authority(program_id: Pubkey) -> Pubkey:
...
 
     @staticmethod
-    def get_intermediary_token_account_pda(program_id: Pubkey, order_address: Pubkey) -> Pubkey:
+    def get_intermediary_token_account_pda(
+        program_id: Pubkey, order_address: Pubkey
+    ) -> Pubkey:
         return Pubkey.find_program_address(
-            seeds=[INTERMEDIARY_OUTPUT_TOKEN_ACCOUNT_SEED, bytes(order_address)], program_id=program_id
+            seeds=[INTERMEDIARY_OUTPUT_TOKEN_ACCOUNT_SEED, bytes(order_address)],
+            program_id=program_id,
```
guibescos, line 355: black did this

PR #290, opened 2024-12-11
feat: make limonade smart

guibescos:
This pull request adds some logic to limonade to ignore orders that are out-of-money. With a 5% threshold, if currently reduces the amount of orders to submit from 200 to 50. 

Features:
- If no `price-config` is provided it will just revert to the old behavior of submitting everything
- If `price-config` is provided it will crash if the hermes streaming is interrupted or times out. This is so kubernetes restarts limonade.
- If no price is available for `quote` or `base` (because one of the two is not listed in `price-config.yaml`  or because the price from Hermes is stale), it will revert to the old behavior of submitting all opportunities

### DIFF
--- a/scripts/limonade/price-config.sample.yaml
+++ b/scripts/limonade/price-config.sample.yaml
@@ -1,0 +1,6 @@
+- alias: SOL
+  mint: So11111111111111111111111111111111111111112
+  id: ef0d8b6fda2ceba41da15d4095d1da392a0d2f8ed0c6c7bc0f4cfac8c280b56d
+- alias: USDC
+  mint: EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v
+  id: eaa020c61cc479712813461ce153894a96a6c00b21ed0cfc2798d1f9a9e9c94a
--- a/scripts/limonade/src/index.ts
+++ b/scripts/limonade/src/index.ts
@@ -122,6 +122,7 @@ async function run() {
       filters,
       withContext: true,
     });
+    console.log("Found ", response.value.length, " orders");
 
     const payloads: OpportunityCreate[] = response.value
       .filter(
@@ -142,11 +143,6 @@ async function run() {
         (opportunityCreate) =>
           opportunityCreate.order.state.remainingInputAmount.toNumber() !== 0
       )
-      // .filter((opportunityCreate) =>
-      //   opportunityCreate.order.address.equals(
-      //     new PublicKey("FcrkTHX99fPokp5CtJzHp5MmvujFnnip57C78TLiaryj")
-      //   )
-      // )
       .filter((opportunityCreate) => !isOffMarket(opportunityCreate.order));
 
     console.log("Resubmitting opportunities", payloads.length);
--- a/scripts/limonade/src/index.ts
+++ b/scripts/limonade/src/index.ts
@@ -74,11 +74,11 @@ const argv = yargs(hideBin(process.argv))
     type: "string",
     default: "https://hermes.pyth.network/",
   })
-  .option("off-market-threshold", {
+  .option("active-opportunity-price-band", {
     description:
-      "Threshold of price ratio to consider an opportunity off-market",
+      "Number of basis points an opportunity's implied price needs to be within the price from Hermes for the opportunity to be considered active",
     type: "number",
-    default: 1.05,
+    default: 500,
   })
   .option("price-staleness-threshold", {
     description:
@@ -199,7 +199,7 @@ async function run() {
             priceInputMint.mintDecimals -
             priceOutputMint.mintDecimals);
 
-      if (ratio > argv.offMarketThreshold) {
+      if (ratio > 1 + argv.activeOpportunityPriceBand / 10000) {
         return true;
       }
 
--- a/scripts/limonade/src/index.ts
+++ b/scripts/limonade/src/index.ts
@@ -8,12 +8,14 @@ import bs58 from "bs58";
 import { hideBin } from "yargs/helpers";
 import yargs from "yargs";
 import {
-  ChainType,
   Client,
   ClientError,
   OpportunityCreate,
 } from "@pythnetwork/express-relay-js";
 import { getPdaAuthority } from "@kamino-finance/limo-sdk/dist/utils";
+import { HermesClient, PriceUpdate } from "@pythnetwork/hermes-client";
+import { PriceConfig, readPriceConfigFile } from "./price-config";
+import { Pubkey } from "@kamino-finance/limo-sdk/dist/rpc_client/types/UpdateGlobalConfigValue";
 
 const lastChange: Record<string, number> = {};
 
@@ -45,7 +47,6 @@ const argv = yargs(hideBin(process.argv))
     description:
       "API key to authenticate with the express relay server for publishing opportunities.",
     type: "string",
-    demandOption: true,
   })
   .option("number-of-concurrent-submissions", {
     description: "Number of concurrent submissions to the express relay server",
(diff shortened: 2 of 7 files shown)


scripts/limonade/src/index.ts
```diff
@@ -197,6 +267,71 @@ async function run() {
...
+
+        const data: PriceUpdate = JSON.parse(event.data);
+        const now = Date.now();
+        if (data.parsed) {
+          for (const parsedUpdate of data.parsed) {
+            const priceConfig = priceConfigs.find(
+              (priceConfig) => priceConfig.pythFeedId === parsedUpdate.id
+            );
+            if (priceConfig) {
+              if (
```
anihamde, line 303: i think you should store the price no matter how stale it is, as long as it's fresher than what is currently stored. i think you can handle processing whether to use the prices in the price evaluation

scripts/limonade/src/index.ts
```diff
@@ -138,6 +177,36 @@ async function run() {
     }
   };
 
+  const isOffMarket = (order: { state: Order; address: PublicKey }) => {
+    const priceInputMint = priceStore[order.state.inputMint.toString()];
+    const priceOutputMint = priceStore[order.state.outputMint.toString()];
+
+    if (!priceInputMint || !priceOutputMint) {
```
anihamde, line 184: i think you should check the price staleness here? otherwise, it's possible a price gets added but then doesn't update and then is stale later on but is still stored?

scripts/limonade/src/index.ts
```diff
@@ -197,6 +267,71 @@ async function run() {
...
+
+    eventSource.onerror = (event: Event) => {
+      console.error("Hermes streaming error", event);
+    };
+
+    /// Await for the first message before continuing
+    await new Promise<void>((resolve, reject) => {
+      setTimeout(() => {
+        reject(new Error("Hermes streaming timeout"));
+      }, argv.hermesStreamingTimeout);
```
anihamde, line 290: there should be a way to write this with a single `setTimeout`, since the error message is not distinct between the two cases (and the timeout durations are the same). any reason you wanted to distinguish between these two?

scripts/limonade/src/index.ts
```diff
@@ -57,16 +59,52 @@ const argv = yargs(hideBin(process.argv))
...
+  })
+  .option("off-market-threshold", {
+    description:
+      "Threshold of price ratio to consider an opportunity off-market",
+    type: "number",
+    default: 1.05,
+  })
+  .option("price-staleness-threshold", {
+    description:
+      "Threshold of price staleness (seconds), if price is stale, we will not use it",
```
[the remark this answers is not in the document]
anihamde: nit, make this milliseconds?

scripts/limonade/src/index.ts
```diff
@@ -138,6 +177,36 @@ async function run() {
...
+    if (!priceInputMint || !priceOutputMint) {
+      // If we don't have price info, we will not consider it off-market
+      return false;
+    } else {
+      const inputAmount = order.state.remainingInputAmount;
+      const outputAmount = order.state.expectedOutputAmount.sub(
+        order.state.filledOutputAmount
+      );
+
+      const ratio =
```
anihamde, line 193: nit, this ratio is probably going to be recomputed many times for a given pair, maybe it makes sense to cache? for now it's probably not going to be too big a deal, so will leave it up to you

scripts/limonade/src/index.ts
```diff
@@ -138,6 +177,36 @@ async function run() {
...
 
+  const isOffMarket = (order: { state: Order; address: PublicKey }) => {
+    const priceInputMint = priceStore[order.state.inputMint.toString()];
+    const priceOutputMint = priceStore[order.state.outputMint.toString()];
+
+    if (!priceInputMint || !priceOutputMint) {
+      // If we don't have price info, we will not consider it off-market
+      return false;
+    } else {
+      const inputAmount = order.state.remainingInputAmount;
```
m30m, line 188: inputAmount can be `initialInputAmount` and outputAmount can be expectedOutputAmount without the subtraction.
This way, we don't need to assume the inputAmount is positive

scripts/limonade/src/index.ts
```diff
@@ -197,6 +267,71 @@ async function run() {
...
+        parsed: true,
+        ignoreInvalidPriceIds: true,
+      }
+    );
+
+    eventSource.onerror = (event: Event) => {
+      console.error("Hermes streaming error", event);
+    };
+
+    /// Await for the first message before continuing
```
m30m, line 286: I think we use double slash in most of the places

scripts/limonade/src/index.ts
```diff
@@ -197,6 +267,71 @@ async function run() {
...
+        const now = Date.now();
+        if (data.parsed) {
+          for (const parsedUpdate of data.parsed) {
+            const priceConfig = priceConfigs.find(
+              (priceConfig) => priceConfig.pythFeedId === parsedUpdate.id
+            );
+            if (priceConfig) {
+              if (
+                parsedUpdate.price.publish_time * 1000 <
+                now - argv.priceStalenessThreshold * 1000
```
m30m, line 305: definition of now can be closer to the usage or you can just make it inline

scripts/limonade/src/index.ts
```diff
@@ -57,16 +59,52 @@ const argv = yargs(hideBin(process.argv))
...
+  .option("hermes-endpoint", {
+    description: "Hermes endpoint to use for fetching price updates",
+    type: "string",
+    default: "https://hermes.pyth.network/",
+  })
+  .option("off-market-threshold", {
+    description:
+      "Threshold of price ratio to consider an opportunity off-market",
+    type: "number",
+    default: 1.05,
```
m30m, line 81: can you add a parameter for how often we try to resubmit opportunities?
This parameter and the frequency go hand in hand. We can deploy multiple limonade instances and tune these. For example, one that resubmits stuff faster if they are closer to the current price

scripts/limonade/src/index.ts
```diff
@@ -57,16 +59,52 @@ const argv = yargs(hideBin(process.argv))
...
     default: 30 * 1000,
   })
+  .option("hermes-streaming-timeout", {
+    description: "Hermes streaming timeout (milliseconds)",
+    type: "number",
+    default: 5 * 1000,
+  })
+  .option("price-config", {
+    description:
+      "Path to the price config file, if not provided, we will not get price info from Hermes",
```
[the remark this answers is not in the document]
guibescos: ooops, forgot

scripts/limonade/src/index.ts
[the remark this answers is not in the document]
guibescos: added clarification in the description

scripts/limonade/src/index.ts
guibescos: done

scripts/limonade/src/index.ts
guibescos: done

scripts/limonade/src/index.ts
guibescos: I like it this way since it reflects more the current state of the order. 
Besides just before filtering by `isWithinMarketPriceBand` with filter by `remainingInputAmount != 0` so we know we won't divide by 0 here.

scripts/limonade/src/index.ts
guibescos: done

scripts/limonade/src/index.ts
guibescos: done

scripts/limonade/src/index.ts
guibescos: I don't see a way to write this with a single timeout.
The first one is handling the case where we never receive any messages.
The second one handles if we've already received messages in the past but we stop receiving them.

scripts/limonade/src/index.ts
guibescos: seems too early to optimize this

scripts/limonade/price-config.sample.yaml
```diff
@@ -0,0 +1,6 @@
+- alias: SOL
```
anihamde, line 1: can you include some notes in the README about cloning this? and also any details about where to grab heremes prices, the new args here, etc?

anihamde (approved):

scripts/limonade/price-config.sample.yaml
guibescos: updated readme

PR #291, opened 2024-12-12
Feat: min slot bid svm

m30m:
add optional parameter for svm bids to sepcify the minimum slot where the bid is applicable.
If simulation fails for any reason and the simulation context has a lower slot compared to the bid, we will retry hoping that the RPC will catch up soon.


auction-server/src/auction/service/verification.rs
```diff
@@ -578,25 +578,43 @@ impl Service<Svm> {
...
-            }
-            Ok(_) => Ok(()),
+        const RETRY_LIMIT: usize = 5;
+        let mut retry_count = 0;
+        loop {
+            let response = self
+                .config
+                .chain_config
+                .simulator
+                .simulate_transaction(&bid.chain_data.transaction)
```
anihamde, line 588: i think the retry should happen within the fetching? before the simulator actually runs? otherwise it's a weird interface where we're asking for a minimum slot to simulate against and then ignoring that as long as it passes simulation.

anihamde: we can also add some checks to make sure searchers aren't submitting slots that are more than some reasonable number off of where the rpc currently stands

auction-server/src/auction/service/verification.rs
```diff
@@ -578,25 +578,43 @@ impl Service<Svm> {
     }
 
     pub async fn simulate_bid(&self, bid: &entities::BidCreate<Svm>) -> Result<(), RestError> {
```
danimhr, line 580: I think calling this function recursively will make the code more readable (instead of having loop)

auction-server/src/auction/service/verification.rs
m30m: I consider this some internal behaviour which is subject to change. Interface only adds the option for a searcher to guarantee successful simulation from the specified slot. How we verify this internally is not part of the interface.

I classify that as a DoS attack and don't think it's a priority

auction-server/src/auction/service/simulator.rs
```diff
@@ -190,11 +191,15 @@ impl Simulator {
         for chunk_result in chunk_results {
             let chunk_result = chunk_result?;
             result.extend(chunk_result.value);
```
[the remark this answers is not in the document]
m30m: agreed, this can become problematic with AMM accounts. I added a comment

auction-server/src/auction/service/verification.rs
m30m: I also considered that but that was also a bit weird. Refactored a bit to make it more readable

auction-server/src/auction/service/verification.rs
```diff
@@ -613,6 +627,8 @@ impl Service<Svm> {
                         reason: msgs.join("\n"),
                     })
                 }
+                // Not important to check if bid slot is less than simulation slot if simulation is successful
```
anihamde, line 630: this comment is a bit confusing, i'm not sure what you're trying to clarify

auction-server/src/auction/service/simulator.rs
```diff
@@ -168,6 +168,8 @@ impl Simulator {
         Ok(result.value)
     }
 
+    /// Fetches multiple accounts from the RPC in chunks
+    /// There is no guarantee that all the accounts will be fetched with the same slot
```
anihamde, line 172: i think this is still weird, but i don't have a great solution for it atm. maybe highlight this as a TODO to think about downstream simulation issues

anihamde: if the max rpc account limit is 100, should be fine though? max of 64 accounts per transaction, where did you get the max limit of 100 from?

auction-server/src/auction/service/verification.rs
anihamde: in that case, maybe it's worth a comment baking that into the interface on the SDK side, or in the `BidChainDataCreateSvm` object in `auction-server/src/auction/entities/bid.rs`

anihamde (approved):

auction-server/src/auction/service/simulator.rs
m30m: Based on the docs:
https://solana.com/docs/rpc/http/getmultipleaccounts

We try to fetch all the accounts needed for multiple transactions, so we might need chunking.

auction-server/src/auction/service/verification.rs
m30m: trying to clarify why we are not checking the bid slot against result.context.slot in all cases

auction-server/src/auction/service/verification.rs
m30m: there is a comment already there. do you think we should explain more?

PR #292, opened 2024-12-13
Fix svm slow conclusion

danimhr:
This PR aims to fix the svm slow conslusion issue

Sometimes the websocket hears about confirmation before the RPC client. In the previous version the conclude auction was dependent on RPC client also seeing the confirmation transaction. Now in the conclusion log listener worker, the code only use the log to conclude the auction.

### DIFF
--- a/auction-server/src/auction/service/conclude_auction.rs
+++ b/auction-server/src/auction/service/conclude_auction.rs
@@ -17,7 +17,7 @@ pub struct ConcludeAuctionInput<T: ChainTrait> {
 
 pub struct ConcludeAuctionWithStatusesInput<T: ChainTrait> {
     pub auction:      entities::Auction<T>,
-    pub bid_statuses: Vec<Option<T::BidStatusType>>,
+    pub bid_statuses: Vec<(T::BidStatusType, entities::Bid<T>)>,
 }
 
 impl<T: ChainTrait> Service<T>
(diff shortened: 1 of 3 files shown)


auction-server/src/auction/service/conclude_auction.rs
```diff
@@ -15,13 +15,55 @@ pub struct ConcludeAuctionInput<T: ChainTrait> {
...
+                .conclude_auction(&mut auction)
+                .await
+                .map_err(|e| anyhow::anyhow!("Failed to conclude auction: {:?}", e))?;
+            self.repo.remove_in_memory_submitted_auction(auction).await;
+        }
+
+        Ok(())
+    }
+
+    /// This one concludes an auction by getting the auction transaction status from the chain.
```
m30m, line 63: "This one" is unnecessary

auction-server/src/auction/service/conclude_auction.rs
```diff
@@ -45,35 +87,17 @@ where
                 )
                 .await?;
 
-            join_all(
-                bid_statuses
+            self.conclude_auction_with_statuses(ConcludeAuctionWithStatusesInput {
+                auction,
+                bid_statuses: bid_statuses
                     .iter()
```
m30m, line 93: nit: into_iter and remove clone
```
                    .into_iter()
                    .zip(bids)
                    .filter_map(|(status, bid)| {
                        status.map(|status| (status, bid))
                    })
```

auction-server/src/auction/service/conclude_auction.rs
```diff
@@ -15,13 +15,55 @@ pub struct ConcludeAuctionInput<T: ChainTrait> {
...
+                .await
+                .map_err(|e| anyhow::anyhow!("Failed to conclude auction: {:?}", e))?;
+            self.repo.remove_in_memory_submitted_auction(auction).await;
+        }
+
+        Ok(())
+    }
+
+    /// This one concludes an auction by getting the auction transaction status from the chain.
+    #[tracing::instrument(skip_all, fields(auction_id, tx_hash, bid_ids))]
```
m30m, line 64: I suggest removing this instrumentation.
tx_hash is a bit misleading now since auction can have multiple transactions.
The rest is traced inside the `conclude_auction_with_statuses` call.

auction-server/src/auction/service/workers.rs
```diff
@@ -132,6 +137,53 @@ impl Service<Evm> {
 const GET_LATEST_BLOCKHASH_INTERVAL_SVM: Duration = Duration::from_secs(5);
 
 impl Service<Svm> {
+    pub async fn conclude_auction_for_log(
```
m30m, line 140: remove `//TODO: this can be optimized out if triggered by websocket events` from auction_manager.rs since you implemented this :D

m30m (commented):
this is not the fix we had in mind. what changed?

auction-server/src/auction/service/conclude_auction.rs
danimhr: I removed the fields, but kept the instument itself.

m30m (approved):

PR #298, opened 2024-12-17
Feat: add market (auction paradigm 2) to server

anihamde:
This PR adapts the server code to welcome market orders/swaps that are in Auction Paradigm 2 (bidding in one of the tokens swapped, rather than in native token).

This PR does so in the minimal way--there are some outstanding changes that will improve the code quality later on in future PRs. Some of these are:

- [ ] Remove `wallet_program_router_account`, use a different way to check the bid belongs to Auction Paradigm 2. i.e. use parsed instruction data, fact that AP2 will include swap ix.
- [ ] Refactor db longer-term. Maybe think about separate OpportunityMarket flag (rather than relying on filtering on the router account)
- [ ] Think about good ways to tilt test this, once program is updated


auction-server/api-types/src/opportunity.rs
```diff
@@ -466,40 +476,45 @@ pub struct OpportunityBidEvm {
...
+    #[schema(example = 50)]
+    pub maximum_slippage_bps: u16,
     /// The chain id for creating the quote.
     #[schema(example = "solana", value_type = String)]
-    pub chain_id:                    ChainId,
+    pub chain_id:             ChainId,
+}
+
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug)]
+pub enum QuoteTokenAmount {
```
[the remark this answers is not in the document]
danimhr: We can also rename this to sth else :-? Maybe like QouteType?

auction-server/src/auction/service/verification.rs
```diff
@@ -463,42 +519,95 @@ impl Service<Svm> {
...
-                        "Invalid deadline: {:?} {:?}",
-                        submit_bid_data.deadline, e
-                    ))
-                },
-            )?,
-        })
+        let submit_bid_instruction_result = self.verify_submit_bid_instruction(transaction.clone());
+        let swap_instruction_result = self.verify_swap_instruction(transaction.clone());
+
+        assert!(
```
danimhr, line 525: This will panic. I think we should not panic here

auction-server/src/auction/service/verification.rs
```diff
@@ -463,42 +519,95 @@ impl Service<Svm> {
...
+            submit_bid_instruction_result.is_err() != swap_instruction_result.is_err(),
+            "Either submit_bid or swap must be present, but not both"
+        );
+
+        if submit_bid_instruction_result.is_ok() {
```
danimhr, line 530: I think we can use match here:

match ((submit_bid_instruction_result, swap_instruction_result)) {
    Err(_), Ok(a) => {},
   Ok(a), Err(_) => {},
   _ => anyhow!(...)
}

auction-server/src/auction/service/verification.rs
```diff
@@ -463,42 +519,95 @@ impl Service<Svm> {
...
+                router,
+                deadline: OffsetDateTime::from_unix_timestamp(submit_bid_data.deadline).map_err(
+                    |e| {
+                        RestError::BadParameters(format!(
+                            "Invalid deadline: {:?} {:?}",
+                            submit_bid_data.deadline, e
+                        ))
+                    },
+                )?,
+                bid_type: BidType::SubmitBid,
```
danimhr, line 566: We can use SubmitType here instead of creating a new struct `BidType`

auction-server/src/auction/service/verification.rs
```diff
@@ -673,11 +783,16 @@ impl Verification<Svm> for Service<Svm> {
             transaction:        bid.chain_data.transaction.clone(),
         };
         let permission_key = bid_chain_data.get_permission_key();
+        let submit_type = self.get_submission_state(&permission_key).await;
+        assert!(
```
danimhr, line 787: Dont use assert

auction-server/src/auction/service/verification.rs
```diff
@@ -673,11 +783,16 @@ impl Verification<Svm> for Service<Svm> {
             transaction:        bid.chain_data.transaction.clone(),
         };
         let permission_key = bid_chain_data.get_permission_key();
+        let submit_type = self.get_submission_state(&permission_key).await;
```
danimhr, line 786: The simulation takes like 100ms and during this time, the state for the submit_type may change (It can become invalid). Leave a comment here to take care of this latency later.

danimhr, line 786: Basically we want to make sure the swap_instruction only exists for the SubmitType::ByOther. Isn't it better to include this logic inside the `extract_bid_data`?

auction-server/src/auction/service/verification.rs
```diff
@@ -673,11 +783,16 @@ impl Verification<Svm> for Service<Svm> {
...
+        assert!(
+            bid_data.bid_type == BidType::SubmitBid && submit_type == entities::SubmitType::ByOther,
+            "Invalid bid format for swap, should not use submit_bid instruction"
+        );
         tracing::Span::current().record("permission_key", bid_data.permission_account.to_string());
         self.check_deadline(&permission_key, bid_data.deadline)
             .await?;
-        self.verify_signatures(&bid, &bid_chain_data).await?;
-        // TODO we should verify that the wallet bids also include another instruction to the swap program with the appropriate accounts and fields
+        self.verify_signatures(&bid, &bid_chain_data, submit_type)
```
danimhr, line 794: I suggested to pass submit_type to this function. But I think it's not necessary as the opportunity exists check, only query the in_memory opportunity and it's fast enough. So I think it's not important to pre compute this value.

auction-server/src/config.rs
```diff
@@ -164,7 +164,7 @@ pub struct ConfigSvm {
     /// Timeout for RPC requests in seconds.
     #[serde(default = "default_rpc_timeout_svm")]
     pub rpc_timeout:                   u64,
-    /// The router account for Phantom.
+    /// The router account for swap program. // TODO: this should be replaced with a different check
```
danimhr, line 167: Isn't this comment for `get_submission_state` function?

auction-server/src/opportunity/api.rs
```diff
@@ -211,7 +211,7 @@ pub async fn post_quote(
     State(store): State<Arc<StoreNew>>,
     Json(params): Json<QuoteCreate>,
 ) -> Result<Json<Quote>, RestError> {
-    if get_program(&auth)? != ProgramSvm::Phantom {
+    if get_program(&auth)? != ProgramSvm::Swap {
```
danimhr, line 214: I'm not sure about changing this part. We wanna create a different api_key for each of our routers. One for each client.

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -33,14 +33,14 @@ pub struct OpportunitySvmProgramLimo {
...
-    pub maximum_slippage_percentage: f64,
+    pub user_wallet_address:  Pubkey,
+    pub maximum_slippage_bps: u16,
 }
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum OpportunitySvmProgram {
     Limo(OpportunitySvmProgramLimo),
-    Phantom(OpportunitySvmProgramWallet),
+    Swap(OpportunitySvmProgramWallet),
```
danimhr, line 43: I think we need a way to extract the program for each of these requests. Are we going to use router account for it? :-?

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -170,29 +170,45 @@ impl From<OpportunitySvm> for api::OpportunitySvm {
...
+                let buy_token_zero = buy_token.amount == 0;
+                let sell_token = val
+                    .sell_tokens
+                    .first()
+                    .ok_or(anyhow::anyhow!(
+                        "Failed to get sell token from opportunity svm"
+                    ))
+                    .expect("Failed to get sell token from opportunity svm");
+                let sell_token_zero = sell_token.amount == 0;
+                assert!(
```
danimhr, line 190: No assert

auction-server/src/opportunity/repository/models.rs
```diff
@@ -59,15 +59,15 @@ pub struct OpportunityMetadataSvmProgramLimo {
...
+    pub user_wallet_address:  Pubkey,
+    pub maximum_slippage_bps: u16,
 }
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(tag = "program", rename_all = "lowercase")]
 pub enum OpportunityMetadataSvmProgram {
     Limo(OpportunityMetadataSvmProgramLimo),
-    Phantom(OpportunityMetadataSvmProgramWallet),
+    Swap(OpportunityMetadataSvmProgramWallet),
```
danimhr, line 70: We should make sure that there is no phantom opportunity inside the DB. Maybe add a task to backlog for it?

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -93,22 +103,22 @@ impl Service<ChainTypeSvm> {
...
         let opportunity_create = self
-            .get_opportunity_create_for_quote(input.quote_create.clone(), output_amount)
+            .get_opportunity_create_for_quote(input.quote_create.clone())
             .await?;
         let opportunity = self
             .add_opportunity(AddOpportunityInput {
                 opportunity: opportunity_create,
             })
             .await?;
+        let input_specified = matches!(
```
danimhr, line 116: remove this var and use match later

danimhr:
There are some errors when running cargo check which we can resolve

auction-server/src/config.rs
anihamde: yes but we should be able to remove this field from the config once we fix that method

auction-server/src/opportunity/repository/models.rs
anihamde: do you mean that we don't currently have any opportunities of this sort saved in the db?

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -135,7 +186,7 @@ impl Service<ChainTypeSvm> {
         // Add metrics
         let labels = [
             ("chain_id", input.quote_create.chain_id.to_string()),
-            ("wallet", "phantom".to_string()),
+            ("router", input.quote_create.router.to_string()),
```
anihamde, line 189: we can also pass in the auth name from `post_quote` if we prefer

auction-server/api-types/src/opportunity.rs
```diff
@@ -206,18 +205,18 @@ pub enum OpportunityCreateProgramParamsV1Svm {
...
         #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
         #[serde_as(as = "DisplayFromStr")]
         user_wallet_address: Pubkey,
 
-        /// The maximum slippage percentage that the user is willing to accept.
-        #[schema(example = 0.5, value_type = f64)]
-        maximum_slippage_percentage: f64,
+        /// The maximum slippage in basis points that the user is willing to accept.
+        #[schema(example = 50, value_type = u16)]
+        maximum_slippage_bps: u16,
```
guibescos, line 219: I think slippage doesn't make sense anymore in this context.

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -50,25 +44,82 @@ pub struct GetQuoteInput {
...
+        entities::QuoteTokens::OutputTokenSpecified {
+            input_token,
+            output_token,
+        } => {
+            let input_token_mint = input_token.as_ref();
+            let output_token_mint = output_token.token.as_ref();
+            output_token_amount = output_token.amount.to_le_bytes();
+            [
+                user_wallet_address.as_ref(),
+                input_token_mint,
```
guibescos, line 78: this seems weird. I think I'd rather have a 1 bytes discriminator between InputTokenSpecified and OutputTokenSpecified

auction-server/api-types/src/opportunity.rs
guibescos: Unless there's a ref price

anihamde: i agree, i kept it in here for legacy reasons, since wallets will always pass this number in their requests. we don't have to consume it, but we could also get rid of it.

auction-server/src/opportunity/service/get_quote.rs
anihamde: yeah that's fair, i guess the main difference here is the placement of the `u64`. i can add a string discriminator at the start to be explicit like "input_token_specified" or "output_token_specified", although these should be distinct already

auction-server/src/auction/service/verification.rs
```diff
@@ -463,42 +514,95 @@ impl Service<Svm> {
...
-            )?,
-        })
+        let submit_bid_instruction_result = self.verify_submit_bid_instruction(transaction.clone());
+        let swap_instruction_result = self.verify_swap_instruction(transaction.clone());
+
+        match (
+            submit_bid_instruction_result.clone(),
+            swap_instruction_result.clone(),
+        ) {
+            (Ok(_), Err(_)) => {
```
danimhr, line 524: You can use Ok(submit_bid_instruction), Err(_) instead of extracting it in the next line.

auction-server/src/auction/service/verification.rs
```diff
@@ -463,42 +514,95 @@ impl Service<Svm> {
...
+                        .map_err(|e| {
+                            RestError::BadParameters(format!(
+                                "Invalid deadline: {:?} {:?}",
+                                submit_bid_data.deadline, e
+                            ))
+                        })?,
+                    submit_type: SubmitType::ByServer,
+                })
+            }
+            (Err(_), Ok(_)) => {
```
danimhr, line 568: Same here

auction-server/src/auction/service/verification.rs
```diff
@@ -463,42 +514,95 @@ impl Service<Svm> {
...
+                //     router: swap_data.router,
+                //     deadline: OffsetDateTime::from_unix_timestamp(swap_data.deadline).map_err(
+                //         |e| {
+                //             RestError::BadParameters(format!(
+                //                 "Invalid deadline: {:?} {:?}",
+                //                 swap_data.deadline, e
+                //             ))
+                //         },
+                //     )?,
+                //     submit_type: SubmitType::ByServer,
```
danimhr, line 599: I think the submit_type here should be by_other. Additionally, maybe it's a better idea to remove this field :-?

auction-server/src/opportunity/api.rs
danimhr, line 214: I think the program here should be `KaminoSwap`

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -336,7 +349,7 @@ impl From<OpportunitySvmProgram> for api::ProgramSvm {
     fn from(val: OpportunitySvmProgram) -> Self {
         match val {
             OpportunitySvmProgram::Limo(_) => api::ProgramSvm::Limo,
-            OpportunitySvmProgram::Phantom(_) => api::ProgramSvm::Phantom,
+            OpportunitySvmProgram::Swap(_) => api::ProgramSvm::Swap,
```
danimhr, line 352: Isnt the program name `KaminoMarket` or sth like this?
Maybe we can talk on this part a bit

auction-server/src/opportunity/service/mod.rs
```diff
@@ -198,12 +194,11 @@ impl ConfigSvm {
     ) -> anyhow::Result<HashMap<ChainId, Self>> {
         Ok(chains
             .iter()
-            .map(|(chain_id, config)| {
+            .map(|(chain_id, _)| {
                 (
                     chain_id.clone(),
                     Self {
-                        wallet_program_router_account: config.wallet_program_router_account,
```
danimhr, line 205: When we receive a new opportunity for `swap`, we need to check if the router is equal to `wallet_program_router_account`. If we don't check it, we may have some opps for swap with a different router key, then someone can submit a `submit_bid` instruction for this router and someone can submit a `swap`. Then we have two different bid type for a single `permission_key` which will lead to errors.
To resolve this issue, for now we should only accept the swap opps which the `wallet_program_router_account` is equal to the router.

PR #300, opened 2024-12-17
fix: don't submit order with negative output

guibescos:


sdk/js/src/index.ts
```diff
@@ -379,10 +380,12 @@ export class Client {
         encoded_order,
         Order.discriminator.length
       );
-      const remainingOutputAmount =
+      const remainingOutputAmount = BN.max(
```
m30m, line 383: I prefer using `anchor.BN` to remain consistent with the rest of the file or change everything to this new format.
The main reason is to explicitly mention this is the BN exposed by anchor and not the BN library itself.

m30m (approved):

sdk/js/src/index.ts
guibescos: fixed

PR #302, opened 2024-12-17
fix: sdk to not swallow errors

guibescos:
This PR aims at fixing the sdk swallowing some auction server error messages like the 422 we saw on Tuesday last week:
before: 
`ClientError: Auction server http error 422 - undefined`
after:
`ClientError: Auction server http error 422 - "Failed to deserialize the JSON body into the target type: data did not match any variant of untagged enum OpportunityCreate"`

### DIFF
--- a/sdk/js/src/index.ts
+++ b/sdk/js/src/index.ts
@@ -434,7 +434,7 @@ export class Client {
     });
     if (response.error) {
       throw ClientError.newHttpError(
-        response.error.error,
+        JSON.stringify(response.error),
         response.response.status
       );
     }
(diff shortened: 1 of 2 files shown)


sdk/js/src/index.ts
```diff
@@ -467,7 +467,7 @@ export class Client {
     });
     if (response.error) {
       throw ClientError.newHttpError(
-        response.error.error,
```
anihamde, line 470: there are still a few other places where the `ClientError.newHttpError(response.error.error,...)` pattern is observed. any particular reason you didn't choose to modify those?

anihamde: e.g. in `submitOpportunity`, `getBids`, etc.

guibescos: fixed

anihamde (approved):

PR #304, opened 2024-12-18
feat: swap program v2

guibescos:
This requires https://github.com/pyth-network/per/pull/305 to be merged

### DIFF
--- a/contracts/svm/programs/express_relay/src/lib.rs
+++ b/contracts/svm/programs/express_relay/src/lib.rs
@@ -1,12 +1,16 @@
 pub mod error;
 pub mod sdk;
 pub mod state;
+pub mod swap;
+pub mod token;
 pub mod utils;
 
 use {
     crate::{
         error::ErrorCode,
         state::*,
+        swap::PostFeeSwapArgs,
+        token::transfer_token_if_needed,
         utils::*,
     },
     anchor_lang::{
@@ -19,6 +23,11 @@ use {
             sysvar::instructions as sysvar_instructions,
         },
         system_program::System,
+    },
+    anchor_spl::token_interface::{
+        Mint,
+        TokenAccount,
+        TokenInterface,
     },
 };
 
@@ -147,6 +156,35 @@ pub mod express_relay {
             amount,
         )
     }
+
+    pub fn swap(ctx: Context<Swap>, data: SwapArgs) -> Result<()> {
+        let PostFeeSwapArgs {
+            input_after_fees,
+            output_after_fees,
+        } = ctx.accounts.transfer_swap_fees(&data)?;
+
+        // Transfer tokens
+        transfer_token_if_needed(
+            &ctx.accounts.searcher_input_ta,
+            &ctx.accounts.trader_input_ata,
+            &ctx.accounts.token_program_input,
+            &ctx.accounts.searcher,
+            &ctx.accounts.mint_input,
+            input_after_fees,
+        )?;
+
+        transfer_token_if_needed(
+            &ctx.accounts.trader_output_ata,
+            &ctx.accounts.searcher_output_ta,
+            &ctx.accounts.token_program_output,
+            &ctx.accounts.trader,
+            &ctx.accounts.mint_output,
+            output_after_fees,
+        )?;
+
+
+        Ok(())
+    }
 }
 
 #[derive(AnchorSerialize, AnchorDeserialize, Eq, PartialEq, Clone, Copy, Debug)]
@@ -259,7 +297,7 @@ pub struct SubmitBid<'info> {
     #[account(mut)]
     pub router: UncheckedAccount<'info>,
 
-    /// CHECK: this cannot be checked against ConfigRouter bc it may not be initialized bc anchor. we need to check this config even when unused to make sure unique fee splits don't exist.
+    /// CHECK: Some routers might have an initialized ConfigRouter at the enforced PDA address which specifies a custom routing fee split. If the ConfigRouter is unitialized we will default to the routing fee split defined in the global ExpressRelayMetadata. We need to pass this account to check whether it exists and therefore there is a custom fee split.
     #[account(seeds = [SEED_CONFIG_ROUTER, router.key().as_ref()], bump)]
     pub config_router: UncheckedAccount<'info>,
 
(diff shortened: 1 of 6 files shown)


contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +463,94 @@ pub struct WithdrawFees<'info> {
...
+
+    #[account(mut,
+        token::authority = express_relay_metadata.relayer_signer,
+    )]
+    pub relayer_fee_receiver_ata: InterfaceAccount<'info, TokenAccount>,
+
+    #[account(mut,
+        token::authority = express_relay_metadata.key(),
+    )]
+    pub protocol_fee_receiver_ata: InterfaceAccount<'info, TokenAccount>,
```
anihamde, line 546: `protocol` is potentially confusing, since we use `protocol` to mean the "router"/partner in other contexts. can we use a clearer term, e.g. `express_relay_fee_receiver_ata`?

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +463,94 @@ pub struct WithdrawFees<'info> {
...
+    #[account(
+        mut,
+        token::mint = mint_output,
+        token::authority = trader,
+        token::token_program = token_program_output
+    )]
+    pub trader_output_ata: InterfaceAccount<'info, TokenAccount>,
+
+    // Accounts for the fee split
+    // TODO can we trust the Associated Token Program or should we check mint and token_program
```
anihamde, line 533: it should be fine if we are checking the Associated Token Program seeds and validating ATA manually. but i don't mind it as an extra level of defensive programming either. will leave it up to you

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +463,94 @@ pub struct WithdrawFees<'info> {
...
+    // Token in which the fees will be paid
+    pub fee_token:     FeeToken,
+}
+
+#[derive(Accounts)]
+#[instruction(data: Box<SwapArgs>)]
+pub struct Swap<'info> {
+    /// Express relay configuration
+    #[account(seeds = [SEED_METADATA], bump)]
+    pub express_relay_metadata: Account<'info, ExpressRelayMetadata>,
```
anihamde, line 492: slight nit, can we keep a similar order to what we have in other instructions? signers usually first, followed by other accounts, `express_relay_metadata` usually near the end iirc

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +463,94 @@ pub struct WithdrawFees<'info> {
...
+
+/// For all swap instructions and contexts, input and output are defined with respect to the searcher
+/// So mint_input refers to the token that the searcher provides to the trader
+/// mint_output refers to the token that the searcher receives from the trader
+/// This choice is made to minimize confusion for the searchers, who are more likely to parse the program
+#[derive(AnchorSerialize, AnchorDeserialize, Eq, PartialEq, Clone, Copy, Debug)]
+pub struct SwapArgs {
+    pub amount_input:  u64,
+    pub amount_output: u64,
+    // The referral fee is specified in parts per million (e.g. 100 = 1 bp)
```
anihamde, line 481: nit, i'd prefer to keep the label in the variable name as well, just to avoid any confusion and keep the expectation very explicit

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -147,6 +152,155 @@ pub mod express_relay {
...
+
+                validate_ata(
+                    &ctx.accounts.protocol_fee_receiver_ata.key(),
+                    &ctx.accounts.express_relay_metadata.key(),
+                    &ctx.accounts.mint_input.key(),
+                )?;
+
+                let (remaining_amount, fee_protocol) = perform_fee_split(
+                    data.amount_input,
+                    ctx.accounts.express_relay_metadata.split_protocol * 100,
```
anihamde, line 190: we can't use these splits here, they're denominated in terms of the surplus/bid amounts. so they're rather large, e.g. split_router is 50%.

i think those splits in the metadata account should be marked as specifically for Auction Paradigm 1 (i.e. native token surplus bids). For this Auction Paradigm 2 (token transfer fees), we should probably add a new field to Express Relay metadata called like `token_fee`. 

We take `token_fee` bps and split it between the relayer and the express relay (we can probably continue to use the splits we have right now). And then we also take a referral fee that is specified in this instruction and send that to the router. This way, we can add a field to Express Relay Metadata called `token_fee`, and it's fine if it's just set to 0 for now. Wdyt?

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +463,94 @@ pub struct WithdrawFees<'info> {
...
+    pub trader_output_ata: InterfaceAccount<'info, TokenAccount>,
+
+    // Accounts for the fee split
+    // TODO can we trust the Associated Token Program or should we check mint and token_program
+    /// CHECK: this is just used to check router_fee_receiver_ata is a valid ATA
+    pub router:                  UncheckedAccount<'info>,
+    pub router_fee_receiver_ata: InterfaceAccount<'info, TokenAccount>,
+
+    #[account(mut,
+        token::authority = express_relay_metadata.relayer_signer,
```
anihamde, line 539: this should actually be the relayer_fee_receiver, which is a separate wallet

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -147,6 +152,155 @@ pub mod express_relay {
             amount,
         )
     }
+
+    pub fn swap(ctx: Context<Swap>, data: SwapArgs) -> Result<()> {
+        validate_ata(
+            &ctx.accounts.trader_input_ata.key(),
```
anihamde, line 158: the trader_input token account doesn't need to be an ATA? it is possible that a trader wants to use tokens from a non-canonical token account and the program should let them?

anihamde: maybe there are some niche hiccups associated with allowing the searcher to specify any trader TA they want? not 100% sure yet, would be good to think about this before committing to one

contracts/svm/programs/express_relay/src/utils.rs
```diff
@@ -49,6 +57,29 @@ pub fn transfer_lamports_cpi<'info>(
...
+    amount: u64,
+) -> Result<()> {
+    let cpi_accounts = SplTransfer {
+        from:      from_ta.clone(),
+        to:        to_ta.clone(),
+        mint:      mint.to_account_info(),
+        authority: authority.clone(),
+    };
+
+    token_interface::transfer_checked(
```
anihamde, line 75: nit, we can save some gas here by checking `amount > 0` first

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +463,94 @@ pub struct WithdrawFees<'info> {
...
+    )]
+    pub trader_input_ata: InterfaceAccount<'info, TokenAccount>,
+
+    #[account(
+        mut,
+        token::mint = mint_output,
+        token::authority = trader,
+        token::token_program = token_program_output
+    )]
+    pub trader_output_ata: InterfaceAccount<'info, TokenAccount>,
```
anihamde, line 530: btw, any reason we don't check in the [accounts struct](https://github.com/solana-developers/anchor-examples/blob/a2fe1fe9a1a009d9e6a88391633b1b788b743d37/account-constraints/associated_token/programs/example/src/lib.rs#L26-L32) and instead validate manually? i realize for some of the accounts where the mint is unclear without the data specifying whether it's input or output token we need to do it manually, but for `trader_output_ata`, could do it here

anihamde: but i also think it's fine to do it manually; doing it in the Accounts struct often has some consequences due to stack consumption due to Anchor inefficiencies

contracts/svm/programs/dummy/Cargo.toml
```diff
@@ -14,10 +14,11 @@ cpi = ["no-entrypoint"]
 no-entrypoint = []
 no-idl = []
 no-log-ix-name = []
-idl-build = ["anchor-lang/idl-build"]
+idl-build = ["anchor-lang/idl-build", "anchor-spl/idl-build"]
```
guibescos, line 17: I had to add this here for `anchor build` to build the idl correctly

contracts/svm/programs/express_relay/src/utils.rs
guibescos: done

contracts/svm/programs/express_relay/src/lib.rs
guibescos: done

contracts/svm/programs/express_relay/src/lib.rs
guibescos: moved `express_relay_metadata` to the end and both signers to the top

contracts/svm/programs/express_relay/src/lib.rs
guibescos: done for trader_output and trader_input, thanks for the remark

contracts/svm/programs/express_relay/src/lib.rs
guibescos: I added these manual checks

guibescos: I added these manual checks

contracts/svm/programs/express_relay/src/lib.rs
guibescos: done, also updated it to bps to be consistent with the other fee ratios that we have

contracts/svm/programs/express_relay/src/lib.rs
guibescos: fixed!

contracts/svm/programs/express_relay/src/lib.rs
guibescos: made it ATA for now, we can change later

guibescos: made it ATA for now, we can change later

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -309,3 +351,97 @@ pub struct WithdrawFees<'info> {
...
+    #[account(
+        mut,
+        associated_token::mint = mint_output,
+        associated_token::authority = trader,
+        associated_token::token_program = token_program_output
+    )]
+    pub trader_output_ata: InterfaceAccount<'info, TokenAccount>,
+
+    // Fee receivers, we can't do any mint checks here because the mint depends on the fee token
+    /// Router fee receiver token account: the referrer can provide an arbitrary receiver for the router fee
```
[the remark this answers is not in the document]
anihamde: i think if you make this an ata, you can just run all the ATA checks in the context by specifying a fee mint account. otherwise, maybe there's some way to catch the mint based on conditional logic using the instruction args that Anchor exposes?

i'm fine with the way it is rn, doing the ata validation in the code, but i know you wanted to look for ways to do it in the context

anihamde (approved):
besides some small comments, lgtm

contracts/svm/programs/express_relay/src/lib.rs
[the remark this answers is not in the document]
guibescos: I moved the validation to the context, it only adds two bytes to the transaction size since it's just 2 extra account indexes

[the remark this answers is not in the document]
guibescos: I think `router_fee_receiver_ta` as an arbitrary token account works because it gives more flexibility to the receiver

contracts/svm/programs/express_relay/src/swap.rs
```diff
@@ -0,0 +1,231 @@
...
+
+pub struct SwapFees {
+    pub router_fee:        u64,
+    pub relayer_fee:       u64,
+    pub express_relay_fee: u64,
+    pub remaining_amount:  u64,
+}
+impl ExpressRelayMetadata {
+    pub fn compute_swap_fees(&self, referral_fee_bps: u64, amount: u64) -> Result<SwapFees> {
+        let total_fee = amount
```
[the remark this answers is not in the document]
guibescos: done

contracts/svm/programs/express_relay/src/state.rs
```diff
@@ -15,6 +15,8 @@ pub struct ExpressRelayMetadata {
     pub split_router_default: u64,
     // the portion of the remaining bid (after router fees) that goes to the relayer, in bps
     pub split_relayer:        u64,
+    // the portion of the swap referral fee that goes to the platform (relayer + express relay), in bps
+    pub swap_platform_fee:    u64,
```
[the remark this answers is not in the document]
guibescos: done

contracts/svm/programs/express_relay/src/lib.rs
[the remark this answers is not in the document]
guibescos: `router_fee_receiver_ta` is specified by the referrer when they call the quote api and it should be checked by the server

PR #308, opened 2025-01-04
fix: svm deps

guibescos:
This PR adds `anchor-spl` to `express-relay` and fixes all the resulting dependency conflicts.


Cargo.toml
```diff
@@ -47,3 +49,11 @@ rand = "0.8.5"
 [patch.crates-io.curve25519-dalek]
 git = "https://github.com/solana-labs/curve25519-dalek.git"
 rev = "b500cdc2a920cd5bff9e2dd974d7b97349d61464"
+
+# Analoguous to the zeroize issue above, the aes-gcm-siv crate has dependencies
+# on the crates zerioze and subtle that are too strict.
+# Both issues were solved as of the commit we are using for this patch.
+
+[patch.crates-io.aes-gcm-siv]
```
anihamde, line 57: do you mind documenting the issue/fix a bit more? doesn't have to be immediate, but it might be good to just leave some notes in at least this PR for posterity

anihamde (approved):

Cargo.toml
guibescos: improved the comment

PR #311, opened 2025-01-07
feat: server swap

anihamde:
This PR adapts the server code to welcome market orders/swaps that are in Auction Paradigm 2 (bidding in one of the tokens swapped, rather than in native token).

This PR does so in the minimal way--there are some outstanding changes that will improve the code quality later on in future PRs. Some of these are:

- [x] Completely remove `wallet_program_router_account`, should not be necessary with the current method of parsing the transaction for the bid payment instruction type.
- [ ] Refactor db longer-term. Maybe think about separate OpportunityMarket flag (rather than relying on filtering on the router account)
- [x] Rename the API exposed `OpportunityCreateProgramParamsV1Svm` enum option to something more generic
- [x] Add deadline parsing once that is added to the program
- [ ] Incorporate fees of the output token in the returned quote amount


auction-server/src/auction/service/verification.rs
```diff
@@ -400,12 +411,30 @@ impl Service<Svm> {
...
-    pub fn verify_submit_bid_instruction(
+    pub fn extract_swap_data(
+        instruction: &CompiledInstruction,
+    ) -> Result<express_relay_svm::SwapArgs, RestError> {
+        let discriminator = express_relay_svm::instruction::Swap::DISCRIMINATOR;
+        express_relay_svm::SwapArgs::try_from_slice(
+            &instruction.data.as_slice()[discriminator.len()..],
+        )
+        .map_err(|e| {
+            RestError::BadParameters(format!("Invalid submit_bid instruction data: {}", e))
```
danimhr, line 422: Invalid "swap" instruction data

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -50,25 +45,82 @@ pub struct GetQuoteInput {
...
 impl Service<ChainTypeSvm> {
     async fn get_opportunity_create_for_quote(
         &self,
         quote_create: entities::QuoteCreate,
-        output_amount: u64,
     ) -> Result<entities::OpportunityCreateSvm, RestError> {
-        let chain_config = self.get_config(&quote_create.chain_id)?;
-        let router = chain_config.wallet_program_router_account;
-        let permission_account = Pubkey::new_from_array(rand::thread_rng().gen());
+        let router = quote_create.router;
```
danimhr, line 95: We should check if the router matches the defined router in the config. If not we should raise error.

danimhr (commented):
We need to upgrade and publish the api-types version as well

PR #312, opened 2025-01-07
feat: create quote response submit method, typescript

anihamde:
This PR adds a method to submit a quote request from the client. It only adds this to the TS sdk, and it is not complete--it is meant just to serve as a quick reference for integrating routers.

### DIFF
--- a/sdk/js/src/index.ts
+++ b/sdk/js/src/index.ts
@@ -492,6 +492,7 @@ export class Client {
       user_wallet_address: quoteRequest.userWallet.toBase58(),
       version: "v1" as const,
     };
+    // TODO: we may want to wrap all the GET/POST calls in a try/catch block to handle errors
     const response = await client.POST("/v1/opportunities/quote", {
       body: body,
     });
--- a/sdk/js/src/index.ts
+++ b/sdk/js/src/index.ts
@@ -642,7 +642,7 @@ export class Client {
   ): QuoteResponse {
     return {
       chainId: quoteResponse.chain_id,
-      expirationTime: BigInt(quoteResponse.expiration_time),
+      expirationTime: new Date(quoteResponse.expiration_time * 1000),
       inputToken: {
         token: new PublicKey(quoteResponse.input_token.token),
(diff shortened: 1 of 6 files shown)


sdk/js/src/index.ts
```diff
@@ -473,6 +476,34 @@ export class Client {
...
+    const body = {
+      chain_id: quoteRequest.chainId,
+      input_token_mint: quoteRequest.inputTokenMint.toBase58(),
+      output_token_mint: quoteRequest.outputTokenMint.toBase58(),
+      router: quoteRequest.router.toBase58(),
+      specified_token_amount: quoteRequest.specifiedTokenAmount,
+      user_wallet_address: quoteRequest.userWallet.toBase58(),
+      version: "v1" as const,
+    };
+    const response = await client.POST("/v1/opportunities/quote", {
```
danimhr, line 495: wrap it in try?

sdk/js/src/index.ts
```diff
@@ -601,6 +632,29 @@ export class Client {
...
+   * Converts a quote response from the server to the client format
+   * @param quoteResponse
+   * @returns Quote response in the converted client format
+   */
+  public convertQuoteResponse(
+    quoteResponse: components["schemas"]["QuoteSvm"]
+  ): QuoteResponse {
+    return {
+      chainId: quoteResponse.chain_id,
+      expirationTime: BigInt(quoteResponse.expiration_time),
```
danimhr, line 645: isn't it better to convert it to a DateTime instead of BigInt?

sdk/js/src/index.ts
danimhr: The await should be wrapped in try, or it cause the whole process to be killed

sdk/js/src/index.ts
danimhr: We can assume the time is utc?

danimhr: While we are converting the type, it's better to make the types as clear as possible

danimhr (approved):

PR #314, opened 2025-01-08
feat: update swap platform fee + update tilt scripts to initialize swap platform fee

guibescos:


contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -248,9 +259,13 @@ pub struct SetSplitsArgs {
...
 }
 
+#[derive(AnchorSerialize, AnchorDeserialize, Eq, PartialEq, Clone, Copy, Debug)]
+pub struct SetSwapPlatformFeeArgs {
+    pub swap_platform_fee_bps: u64,
+}
+
 #[derive(Accounts)]
 pub struct SetSplits<'info> {
-    #[account(mut)]
```
guibescos, line 253: this doesn't need to be mutable imo

anihamde: i think you can remove the constraint from other places too, like `SetAdmin`, `WithdrawFees`, etc.

sdk/python/express_relay/svm/generated/express_relay/instructions/set_splits.py
```diff
@@ -26,7 +26,7 @@ def set_splits(
     remaining_accounts: typing.Optional[typing.List[AccountMeta]] = None,
 ) -> Instruction:
     keys: list[AccountMeta] = [
-        AccountMeta(pubkey=accounts["admin"], is_signer=True, is_writable=True),
+        AccountMeta(pubkey=accounts["admin"], is_signer=True, is_writable=False),
```
anihamde, line 29: ditto here, can set to False for all the other instructions where you remove `mut`

anihamde (approved):

sdk/python/express_relay/svm/generated/express_relay/instructions/set_splits.py
guibescos: done

PR #318, opened 2025-01-09
Add submit bid instruction

danimhr:
This PR aims to add svm functionalities for rust sdk

### DIFF
--- a/contracts/svm/programs/express_relay/src/sdk/helpers.rs
+++ b/contracts/svm/programs/express_relay/src/sdk/helpers.rs
@@ -6,7 +6,6 @@ use {
         FeeToken,
         SubmitBidArgs,
         SwapArgs,
-        ID as EXPRESS_RELAY_PID,
         SEED_CONFIG_ROUTER,
         SEED_METADATA,
     },
@@ -19,13 +18,14 @@ use {
         system_program,
         InstructionData,
     },
-    anchor_spl::token::spl_token,
+    anchor_spl::associated_token::get_associated_token_address_with_program_id,
 };
 
 /// Creates and adds to the provided instructions a `SubmitBid` instruction.
 #[allow(clippy::too_many_arguments)]
-pub fn add_express_relay_submit_bid_instruction(
+pub fn add_submit_bid_instruction(
     ixs: &mut Vec<Instruction>,
+    express_relay_pid: Pubkey,
     searcher: Pubkey,
     relayer_signer: Pubkey,
     fee_receiver_relayer: Pubkey,
@@ -35,6 +35,7 @@ pub fn add_express_relay_submit_bid_instruction(
     deadline: i64,
 ) -> Vec<Instruction> {
     let ix_submit_bid = create_submit_bid_instruction(
+        express_relay_pid,
         searcher,
         relayer_signer,
         fee_receiver_relayer,
@@ -48,7 +49,9 @@ pub fn add_express_relay_submit_bid_instruction(
     ixs.to_vec()
 }
 
+#[allow(clippy::too_many_arguments)]
 pub fn create_submit_bid_instruction(
+    express_relay_pid: Pubkey,
     searcher: Pubkey,
     relayer_signer: Pubkey,
     fee_receiver_relayer: Pubkey,
@@ -58,9 +61,9 @@ pub fn create_submit_bid_instruction(
     bid_amount: u64,
 ) -> Instruction {
     let config_router =
-        Pubkey::find_program_address(&[SEED_CONFIG_ROUTER, router.as_ref()], &EXPRESS_RELAY_PID).0;
+        Pubkey::find_program_address(&[SEED_CONFIG_ROUTER, router.as_ref()], &express_relay_pid).0;
     let express_relay_metadata =
-        Pubkey::find_program_address(&[SEED_METADATA], &EXPRESS_RELAY_PID).0;
+        Pubkey::find_program_address(&[SEED_METADATA], &express_relay_pid).0;
 
     let accounts_submit_bid = accounts::SubmitBid {
         searcher,
@@ -83,7 +86,7 @@ pub fn create_submit_bid_instruction(
     .data();
 
     Instruction {
-        program_id: EXPRESS_RELAY_PID,
+        program_id: express_relay_pid,
         accounts:   accounts_submit_bid,
         data:       data_submit_bid,
     }
@@ -91,6 +94,7 @@ pub fn create_submit_bid_instruction(
 
 #[allow(clippy::too_many_arguments)]
 pub fn create_swap_instruction(
+    express_relay_pid: Pubkey,
     searcher: Pubkey,
     trader: Pubkey,
     searcher_input_ta: Option<Pubkey>,
@@ -99,12 +103,12 @@ pub fn create_swap_instruction(
     fee_receiver_relayer: Pubkey,
     mint_input: Pubkey,
     mint_output: Pubkey,
-    token_program_input: Option<Pubkey>,
-    token_program_output: Option<Pubkey>,
+    token_program_input: Pubkey,
+    token_program_output: Pubkey,
     swap_args: SwapArgs,
 ) -> Instruction {
     let express_relay_metadata =
-        Pubkey::find_program_address(&[SEED_METADATA], &EXPRESS_RELAY_PID).0;
+        Pubkey::find_program_address(&[SEED_METADATA], &express_relay_pid).0;
 
     let mint_fee = match swap_args.fee_token {
         FeeToken::Input => mint_input,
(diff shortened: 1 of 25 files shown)


sdk/rust/src/lib.rs
```diff
@@ -660,49 +673,177 @@ pub trait Biddable {
...
+                ));
+                instructions.push(create_associated_token_account_idempotent(
+                    &params.payer,
+                    &params.fee_receiver_relayer,
+                    &fee_token,
+                    &fee_token_program,
+                ));
+                instructions.push(create_associated_token_account_idempotent(
+                    &params.payer,
+                    &Pubkey::find_program_address(
```
m30m, line 818: let's make this a separate function

sdk/rust/src/lib.rs
```diff
@@ -660,49 +673,177 @@ pub trait Biddable {
...
+                    &fee_token_program,
+                ));
+                instructions.push(svm::Svm::get_swap_instruction(NewSwapInstructionParams {
+                    opportunity_params:   opportunity.params,
+                    bid_amount:           params.amount,
+                    deadline:             params.deadline,
+                    searcher:             params.searcher,
+                    fee_receiver_relayer: params.fee_receiver_relayer,
+                })?);
+                let mut transaction =
```
m30m, line 833: this final logic is duplicate and we can do it outside match

contracts/svm/programs/express_relay/src/sdk/helpers.rs
```diff
@@ -0,0 +1,170 @@
...
+    fee_receiver_relayer: Pubkey,
+    permission: Pubkey,
+    router: Pubkey,
+    deadline: i64,
+    bid_amount: u64,
+) -> Instruction {
+    let config_router =
+        Pubkey::find_program_address(&[SEED_CONFIG_ROUTER, router.as_ref()], &EXPRESS_RELAY_PID).0;
+    let express_relay_metadata =
+        Pubkey::find_program_address(&[SEED_METADATA], &EXPRESS_RELAY_PID).0;
```
m30m, line 63: EXPRESS_RELAY_PID should not be hardcoded unfortunately because now we have separate IDs for staging and prod. I think it should be a function input here but for the sdk, we should get the chain_id and decide based on that (similar to python and js)

sdk/rust/src/svm.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+pub struct ProgramParamsSwap {}
+
+pub enum ProgramParams {
+    Limo(ProgramParamsLimo),
+    Swap(ProgramParamsSwap),
+}
+
+pub struct NewBidParams {
```
m30m, line 49: I'm not sure why some of these structs have the `New` prefix. For example shouldn't `get_submit_bid_instruction` accept a `GetSubmitBidInstructionParams` type?

sdk/rust/src/lib.rs
```diff
@@ -660,49 +673,177 @@ pub trait Biddable {
...
+                            input_token_program,
+                            output_token_program,
+                        ),
+                    };
+                let (fee_token, fee_token_program) = match fee_token {
+                    FeeToken::InputToken => (input_token, input_token_program),
+                    FeeToken::OutputToken => (output_token, output_token_program),
+                };
+                let mut instructions = params.instructions;
+                instructions.push(create_associated_token_account_idempotent(
```
m30m, line 798: I think it's nice to have a function that gets all these 5 instructions without the need to pass some localwallet.
Maybe another utility function in svm.rs that wraps `get_swap_instruction` and add these as well?

sdk/rust/src/svm.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub searcher:             Pubkey,
+    pub permission:           Pubkey,
+    pub router:               Pubkey,
+    pub relayer_signer:       Pubkey,
+    pub fee_receiver_relayer: Pubkey,
+}
+
+pub struct NewSwapInstructionParams {
+    pub searcher:             Pubkey,
+    pub opportunity_params:   OpportunityParamsSvm,
```
m30m, line 74: Maybe we can OpportunityParamsV1ProgramSvm::Swap variant data a separate structure (right now, it's unnamed). Then we can only accept that structure here instead of the general params.

sdk/rust/src/lib.rs
```diff
@@ -660,49 +673,177 @@ pub trait Biddable {
...
+                    &params.payer,
+                    &user_wallet_address,
+                    &output_token,
+                    &output_token_program,
+                ));
+                instructions.push(create_associated_token_account_idempotent(
```
m30m, line 804: let's only do this if referral_fee_bps > 0

sdk/rust/src/lib.rs
danimhr: The bid creation part is not the same.

sdk/rust/testing-searcher/src/main.rs
```diff
@@ -21,8 +22,18 @@ async fn main() {
         env::var("ADAPTER_BYTECODE_HASH").expect("ADAPTER_BYTECODE_HASH is not set");
     let permit2 = env::var("PERMIT2").expect("PERMIT2 is not set");
     let chain_id_num = env::var("CHAIN_ID_NUM").expect("CHAIN_ID_NUM is not set");
+    let svm_private_key_file =
+        env::var("SVM_PRIVATE_KEY_FILE").expect("SVM_PRIVATE_KEY_FILE is not set");
+    let svm_rpc_url = "http://127.0.0.1:8899";
     let server_url = "http://127.0.0.1:9000";
 
+
+    let svm_private_key_file_content = std::fs::read_to_string(svm_private_key_file.clone())
```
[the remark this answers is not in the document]
danimhr: I need to clone the SimpleSearcher. That's why I'm using `String` for now

sdk/rust/src/svm.rs
danimhr: What about versioning?

sdk/rust/src/svm.rs
```diff
@@ -0,0 +1,317 @@
...
+                ClientError::SvmError("Failed to fetch express relay metadata".to_string())
+            })?;
+        deserialize_metadata(data).map_err(|_| {
+            ClientError::SvmError("Failed to deserialize express relay metadata".to_string())
+        })
+    }
+
+    pub fn get_express_relay_pid(chain_id: String) -> Pubkey {
+        if chain_id == "development-solana" {
+            Pubkey::from_str("stag1NN9voD7436oFvKmy1kvRZYLLW8drKocSCt2W79")
```
m30m, line 134: Ani mentioned another method called new_from_array that is a const function, so you don't need the error handling anymore but you have to input the byte array which is ugly

sdk/rust/src/svm.rs
```diff
@@ -0,0 +1,317 @@
...
+                .expect("Failed to parse express relay pubkey")
+        } else {
+            express_relay::ID.to_bytes().into()
```
m30m, line 137: Let's be explicit here and mention the program name again and not rely on the actual program declared id.

sdk/rust/src/svm.rs
```diff
@@ -0,0 +1,317 @@
...
+impl Svm {
+    pub fn new(rpc_url: String) -> Self {
+        Self {
+            client: RpcClient::new(rpc_url),
+        }
+    }
+
+    pub async fn get_express_relay_metadata(&self) -> Result<ExpressRelayMetadata, ClientError> {
+        let express_relay_metadata =
+            Pubkey::find_program_address(&[SEED_METADATA], &express_relay::ID.to_bytes().into()).0;
```
m30m, line 118: I think this should also use `get_express_relay_pid`

sdk/rust/src/svm.rs
```diff
@@ -0,0 +1,317 @@
...
+        ));
+        instructions.push(create_associated_token_account_idempotent(
+            &params.payer,
+            &params.fee_receiver_relayer,
+            &params.fee_token,
+            &params.fee_token_program,
+        ));
+        instructions.push(create_associated_token_account_idempotent(
+            &params.payer,
+            &Pubkey::find_program_address(&[SEED_METADATA], &express_relay::ID.to_bytes().into()).0,
```
m30m, line 212: same here

contracts/svm/programs/express_relay/src/sdk/helpers.rs
```diff
@@ -0,0 +1,186 @@
...
+    mint_input: Pubkey,
+    mint_output: Pubkey,
+    token_program_input: Pubkey,
+    token_program_output: Pubkey,
+    swap_args: SwapArgs,
+) -> Instruction {
+    let express_relay_metadata =
+        Pubkey::find_program_address(&[SEED_METADATA], &express_relay_pid).0;
+
+    let mint_fee = match swap_args.fee_token {
```
m30m, line 113: maybe merge these 2 matches

contracts/svm/programs/express_relay/src/sdk/helpers.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+
+#[allow(clippy::too_many_arguments)]
+pub fn create_swap_instruction(
+    express_relay_pid: Pubkey,
+    searcher: Pubkey,
+    trader: Pubkey,
+    searcher_input_ta: Option<Pubkey>,
+    searcher_output_ta: Option<Pubkey>,
+    router_fee_receiver_ta: Pubkey,
```
m30m, line 102: why not calculate this here similar to the others?

sdk/rust/src/svm.rs
danimhr: I prefer this version for readability

sdk/rust/src/svm.rs
danimhr: Good catch

contracts/svm/programs/express_relay/src/sdk/helpers.rs
danimhr: This is in the contract sdk. As the variable is TA, I prefer to get it as input. If we make it optional, then we need to get the `router_account` pubkey. I think the contract should not know about the router and I feel it's better to get the TA itself :-?

PR #320, opened 2025-01-10
feat: Swap js integration

m30m:


Tiltfile
```diff
@@ -211,14 +211,21 @@ local_resource(
...
+    serve_cmd="poetry run python3 -m express_relay.searcher.examples.testing_searcher_svm --endpoint-express-relay http://127.0.0.1:9000 --chain-id development-solana --private-key-json-file ../../keypairs/searcher_py.json --endpoint-svm http://127.0.0.1:8899 --bid 10000000 --fill-rate 4 --bid-margin 100 --with-latency",
     serve_dir="sdk/python",
     resource_deps=["svm-initialize-programs", "auction-server"],
 )
 
 local_resource(
     "svm-searcher-js",
-    serve_cmd="pnpm run testing-searcher-limo --endpoint-express-relay http://127.0.0.1:9000 --chain-id development-solana --private-key-json-file ../../keypairs/searcher_py.json --endpoint-svm http://127.0.0.1:8899 --bid 10000000 --fill-rate 4 --bid-margin 100 --with-latency",
+    serve_cmd="pnpm run testing-searcher-limo --endpoint-express-relay http://127.0.0.1:9000 --chain-id development-solana --private-key-json-file ../../keypairs/searcher_js.json --endpoint-svm http://127.0.0.1:8899 --bid 1000 --fill-rate 4 --bid-margin 100",
```
m30m, line 221: mint more tokens to everyone and revert bid number to the higher value

.pre-commit-config.yaml
```diff
@@ -25,14 +25,14 @@ repos:
         language: "rust"
         entry: cargo +nightly-2024-12-03 fmt --manifest-path ./Cargo.toml --all -- --config-path rustfmt.toml
         pass_filenames: false
-        types_or: ["rust", "cargo", "cargo-lock"]
```
m30m, line 28: fix local pre-commit

sdk/js/src/const.ts
```diff
@@ -28,10 +28,12 @@ export const SVM_CONSTANTS: Record<string, SvmConstantsConfig> = {
     expressRelayProgram: new PublicKey(
       "PytERJFhAKuNNuaiXkApLfWzwNwSNDACpigT3LwQfou"
     ),
+    walletRouter: new PublicKey("3hv8L8UeBbyM3M25dF3h2C5p8yA4FptD7FFZu4Z1jCMn"),
   },
   solana: {
     expressRelayProgram: new PublicKey(
       "PytERJFhAKuNNuaiXkApLfWzwNwSNDACpigT3LwQfou"
     ),
+    walletRouter: new PublicKey("3hv8L8UeBbyM3M25dF3h2C5p8yA4FptD7FFZu4Z1jCMn"),
```
m30m, line 37: add todo to change later

sdk/js/src/svm.ts
```diff
@@ -82,6 +88,175 @@ export async function constructSubmitBidInstruction(
...
+  relayerSigner: PublicKey
+): Promise<BidSvm> {
+  const expressRelay = new Program<ExpressRelay>(
+    expressRelayIdl as ExpressRelay,
+    {} as AnchorProvider
+  );
+  const expressRelayMetadata = getExpressRelayMetadataPda(chainId);
+  const svmConstants = SVM_CONSTANTS[chainId];
+
+  const tokenProgramInput = TOKEN_PROGRAM_ID;
```
m30m, line 141: get from server

auction-server/api-types/src/opportunity.rs
```diff
@@ -61,13 +61,9 @@ pub struct OpportunityBidResult {
...
 
 #[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug, Display)]
-#[serde(rename_all = "lowercase")]
+#[serde(rename_all = "snake_case")]
 pub enum ProgramSvm {
-    #[serde(rename = "swap_kamino")]
-    #[strum(serialize = "swap_kamino")]
     SwapKamino,
-    #[serde(rename = "limo")]
-    #[strum(serialize = "limo")]
```
danimhr, line 70: strum and serde are doing diff stuff. Are you sure it's safe to remove the strum?

auction-server/src/auction/service/verification.rs
```diff
@@ -616,23 +601,71 @@ impl Service<Svm> {
...
+                let associated_token_program = Pubkey::from_str(
+                    "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL",
+                )
+                .map_err(|e| {
+                    RestError::BadParameters(format!(
+                        "Invalid associated token program address: {:?}",
+                        e
+                    ))
+                })?;
+                let expected_router_fee_receiver_ta = Pubkey::find_program_address(
```
danimhr, line 636: We should use this address for creating the permission key

auction-server/src/auction/service/verification.rs
```diff
@@ -616,23 +601,71 @@ impl Service<Svm> {
...
-                    deadline: OffsetDateTime::now_utc(),
-                    // deadline: OffsetDateTime::from_unix_timestamp(swap_data.deadline).map_err(
-                    //     |e| {
-                    //         RestError::BadParameters(format!(
-                    //             "Invalid deadline: {:?} {:?}",
-                    //             swap_data.deadline, e
-                    //         ))
-                    //     },
-                    // )?,
+                    router: self.config.chain_config.wallet_program_router_account,
```
danimhr, line 660: if we use the `router_fee_receiver_ta` to create the permission key, we don't need to do these checks. actually the permission key will be invalid if the receiver_ta account does not match with what we expect.

auction-server/src/auction/service/verification.rs
```diff
@@ -714,7 +753,13 @@ impl Service<Svm> {
                 )
             }
             SubmitType::ByServer => {
-                self.all_signatures_exists(&message_bytes, accounts, &signatures, &[])
+                self.relayer_signer_exists(accounts, &signatures)?;
```
danimhr, line 756: Does the relayer exist at this point?

auction-server/src/opportunity/api.rs
```diff
@@ -207,20 +207,14 @@ pub async fn get_opportunities(
     (status = 404, description = "No quote available right now", body = ErrorBodyResponse),
 ),)]
 pub async fn post_quote(
-    auth: Auth,
     State(store): State<Arc<StoreNew>>,
     Json(params): Json<QuoteCreate>,
 ) -> Result<Json<Quote>, RestError> {
-    let program = get_program(&auth)?;
```
danimhr, line 214: I think it's better to make it permissioned at the beginnig

danimhr (commented):
I reviewed the server parts. You may need to merge with main. There is a bunch of other changes merged recently.

.pre-commit-config.yaml
anihamde: same here, this issue popped up for me out of the blue yesterday. updating pre commit resolved it

sdk/js/src/svm.ts
```diff
@@ -82,6 +88,175 @@ export async function constructSubmitBidInstruction(
...
+      tokenProgramInput,
+      mintInput,
+      traderOutputAta: getAssociatedTokenAddress(
+        trader,
+        mintOutput,
+        tokenProgramOutput
+      ),
+      tokenProgramOutput,
+      mintOutput,
+      routerFeeReceiverTa: getAssociatedTokenAddress(
```
danimhr, line 199: Maybe it's better to make it Ata inside the program?

sdk/js/src/types.ts
```diff
@@ -124,9 +122,35 @@ export type OpportunitySvm = {
   opportunityId: string;
 };
 
+export type OpportunitySvmLimo = {
+  order: OrderStateAndAddress;
+  program: "limo";
+} & OpportunitySvmMetadata;
+
+export type OpportunitySvmSwap = {
```
anihamde, line 130: this should have the `referral_fee_bps` as well

sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -133,6 +136,53 @@ export class SimpleSearcherLimo {
     return bid;
   }
 
+  /**
+   * Generates a bid for a given limo opportunity.
```
anihamde, line 140: typo

sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -133,6 +136,53 @@ export class SimpleSearcherLimo {
...
+   * The transaction in this bid transfers assets from the searcher's wallet to fulfill the limit order.
```
anihamde, line 141: update docstring

sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -133,6 +136,53 @@ export class SimpleSearcherLimo {
...
+   * @param opportunity The SVM opportunity to bid on.
+   * @returns The generated bid object.
+   */
+  async generateBidSwap(opportunity: OpportunitySvmSwap): Promise<BidSvm> {
```
anihamde, line 145: i think this should live in a separate `simpleSearcherSwap.ts` file

sdk/js/src/index.ts
```diff
@@ -627,9 +627,39 @@ export class Client {
...
+        program: "swap",
+        permissionAccount: new PublicKey(opportunity.permission_account),
+        routerAccount: new PublicKey(opportunity.router_account),
+        userWalletAddress: new PublicKey(opportunity.user_wallet_address),
+        tokens,
+      };
     } else {
-      console.warn("Cannot handle wallet opportunities");
-      return undefined;
+      console.warn("Unsupported opportunity", opportunity);
```
anihamde, line 662: we have `constructSvmBid` in the client, it may be good to have `constructSvmSwap` as part of the client for consistency

sdk/js/src/index.ts
```diff
@@ -615,7 +615,7 @@ export class Client {
         buyTokens: opportunity.buy_tokens.map(checkTokenQty),
```
anihamde, line 615: i think for `submitOpportunity` we should fail if opportunity.program == "swap"

anihamde: this would help make it clearer (along with some docs) that swap opportunities are not meant to be created via post opportunity but rather through getQuote

sdk/js/src/svm.ts
```diff
@@ -82,6 +88,175 @@ export async function constructSubmitBidInstruction(
...
 
+export function getAssociatedTokenAddress(
+  owner: PublicKey,
+  tokenMintAddress: PublicKey,
+  tokenProgram: PublicKey
+): PublicKey {
+  return getAssociatedTokenAddressSync(
+    tokenMintAddress,
+    owner,
+    true,
```
anihamde, line 99: makes sense why you defined this, but maybe a comment noting that this allows off-curve owners?

sdk/js/src/svm.ts
```diff
@@ -82,6 +88,175 @@ export async function constructSubmitBidInstruction(
...
+}
+
+export function createAssociatedTokenAccountIdempotentInstruction(
+  owner: PublicKey,
+  mint: PublicKey,
+  payer: PublicKey = owner,
+  tokenProgram: PublicKey
+): [PublicKey, TransactionInstruction] {
+  const ataAddress = getAssociatedTokenAddress(owner, mint, tokenProgram);
+  const createUserTokenAccountIx = createAssociatedTokenAccountInstruction(
```
anihamde, line 112: why not just use this [helper](https://solana-labs.github.io/solana-program-library/token/js/functions/createAssociatedTokenAccountIdempotentInstruction.html)?

sdk/js/src/svm.ts
```diff
@@ -82,6 +88,175 @@ export async function constructSubmitBidInstruction(
...
+      swapOpportunity.tokens.type === "input_specified"
+        ? new anchor.BN(swapOpportunity.tokens.inputToken.amount)
+        : bidAmount,
+    amountOutput:
+      swapOpportunity.tokens.type === "output_specified"
+        ? new anchor.BN(swapOpportunity.tokens.outputToken.amount)
+        : bidAmount,
+    referralFeeBps: new anchor.BN(0),
+    deadline,
+    feeToken: { input: {} },
```
anihamde, line 167: you can get this from the server. right now, the server is always returning input

sdk/js/src/svm.ts
anihamde, line 199: this is not necessarily the ATA. the router account here will be the token account that the referrer wants the SPL fee token to be sent to, so you can just use that

tilt-scripts/svm/test_swap.py
```diff
@@ -0,0 +1,95 @@
+import argparse
```
anihamde, line 1: i think it would be easier (and better) to start writing the tilt scripts in JS? i think you could probably write this whole script more flexibly (using SDK methods rather than writing endpoints and payloads manually) in JS

.pre-commit-config.yaml
guibescos: yes you should run this hook when `Cargo.toml` or `Cargo.lock` get updated

auction-server/api-types/src/opportunity.rs
guibescos: I think strum here will affect the `Display` which is used in the tracing.
So after this change it might be with a capital letter in the tracing?

auction-server/src/auction/service/verification.rs
m30m: yes, this is the ByServer variant (for limo orders)

sdk/js/src/index.ts
m30m: The `OpportunityCreate` type can only be `OpportunitySvmLimo` and not the swap type in the first place

auction-server/src/auction/entities/bid.rs
```diff
@@ -262,11 +262,33 @@ pub struct BidCreate<T: ChainTrait> {
...
-pub struct BidChainDataCreateSvm {
+pub struct BidChainDataOnChainCreateSvm {
     pub transaction: VersionedTransaction,
     pub slot:        Option<Slot>,
 }
 
+#[derive(Clone, Debug)]
+pub struct BidChainDataSwapCreateSvm {
+    pub transaction:    VersionedTransaction,
+    pub opportunity_id: Uuid,
```
m30m, line 273: switch to opportunity id

sdk/js/src/examples/simpleSearcherLimo.ts
m30m: I agree, I'd rather have 2 separate examples for limo and swap but at the moment this makes a bunch of stuff more complicated. I will probably need to duplicate testingSearcherLimo and make a testingSearcherSwap as well which is very dirty. I'd rather think more and make a cleaner attempt in a separate PR

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -271,9 +293,10 @@ impl TryFrom<repository::Opportunity<repository::OpportunityMetadataSvm>> for Op
                     order_address: program.order_address,
                 })
             }
-            repository::OpportunityMetadataSvmProgram::SwapKamino(program) => {
-                OpportunitySvmProgram::SwapKamino(OpportunitySvmProgramSwap {
+            repository::OpportunityMetadataSvmProgram::Swap(program) => {
+                OpportunitySvmProgram::Swap(OpportunitySvmProgramSwap {
                     user_wallet_address:  program.user_wallet_address,
+                    quote_tokens:         program.quote_tokens,
```
anihamde, line 299: aren't these quote token values stored in the Opportunity core fields? why the redundancy of including them in the program field?

auction-server/src/auction/service/verification.rs
```diff
@@ -533,82 +529,206 @@ impl Service<Svm> {
...
-                            .chain_config
-                            .express_relay
-                            .mint_input_account_position_swap,
+                        svm_config
+                            .swap_instruction_account_positions
+                            .mint_output_account,
                     )
                     .await?;
-                let mint_output = self
+
```
danimhr, line 567: Remove this line

auction-server/src/auction/service/verification.rs
```diff
@@ -533,82 +529,206 @@ impl Service<Svm> {
...
+                    .opportunity_service
+                    .get_live_opportunity_by_id(GetOpportunityByIdInput {
+                        opportunity_id: bid_data.opportunity_id,
+                    })
+                    .await
+                    .ok_or(RestError::BadParameters(
+                        "No swap opportunity with the given id found".to_string(),
+                    ))?;
+
+                let opp_swap_data = match opp.program {
```
danimhr, line 579: Extract a function to check these fields

auction-server/src/auction/service/verification.rs
```diff
@@ -533,82 +529,206 @@ impl Service<Svm> {
...
-                };
+                if router_fee_receiver_ta != expected_router_fee_receiver_ta {
+                    return Err(RestError::BadParameters(
+                        format!("Associated token account for router does not match. Expected: {:?} found: {:?}", expected_router_fee_receiver_ta, router_fee_receiver_ta),
+                    ));
+                }
 
-                let permission_account =
-                    get_quote_permission_key(&tokens, &user_wallet, swap_data.referral_fee_bps);
+                let permission_account = get_quote_permission_account(
```
danimhr, line 722: Add router_ta to the permission_account

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -49,7 +55,7 @@ pub struct GetQuoteInput {
     pub program:      ProgramSvm,
 }
 
-pub fn get_quote_permission_key(
+pub fn get_quote_permission_account(
```
anihamde, line 58: nit, this is not an account in the sense that it doesn't really exist, key feels more accurate a descriptor

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -316,13 +339,15 @@ impl From<api::OpportunityCreateSvm> for OpportunityCreateSvm {
...
-            } => OpportunitySvmProgram::SwapKamino(OpportunitySvmProgramSwap {
+                quote_tokens,
+            } => OpportunitySvmProgram::Swap(OpportunitySvmProgramSwap {
                 user_wallet_address,
                 // TODO*: see comment above about this arm
                 fee_token: FeeToken::InputToken,
                 referral_fee_bps,
                 input_token_program,
                 output_token_program,
+                quote_tokens: quote_tokens.into(),
```
danimhr, line 350: Why we need the quote_tokens here?

auction-server/src/opportunity/repository/models.rs
```diff
@@ -64,6 +67,7 @@ pub struct OpportunityMetadataSvmProgramSwap {
     #[serde_as(as = "DisplayFromStr")]
     pub user_wallet_address:  Pubkey,
     pub fee_token:            FeeToken,
+    pub quote_tokens:         QuoteTokens,
```
danimhr, line 70: I think it's not a good idea to store this value in the metadata. We have `sell_tokens` and `buy_tokens` in the opportunity itself.

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -129,13 +136,13 @@ impl Service<ChainTypeSvm> {
             ),
             chain_id:       quote_create.chain_id.clone(),
             sell_tokens:    vec![entities::TokenAmountSvm {
-                token:  input_mint,
```
anihamde, line 132: isn't the input token in this case that which the searcher is providing? https://github.com/pyth-network/per/blob/240b2ba21076ca0a1af86af23c0e34271d499c96/contracts/svm/programs/express_relay/src/lib.rs#L180-L182

auction-server/src/opportunity/service/get_opportunities.rs
```diff
@@ -17,7 +18,19 @@ pub struct GetOpportunitiesInput {
     pub query_params: GetOpportunitiesQueryParams,
 }
 
+pub struct GetOpportunityByIdInput {
```
danimhr, line 21: GetLiveOpportunityByIdInput

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -205,6 +213,13 @@ impl Service<ChainTypeSvm> {
 
     #[tracing::instrument(skip_all)]
     pub async fn get_quote(&self, input: GetQuoteInput) -> Result<entities::Quote, RestError> {
+        if FEE_SPLIT_PRECISION < input.quote_create.referral_fee_bps.into() {
```
anihamde, line 216: technically, we need a stronger check, since the instruction can fail if `referral_fee_bps < FEE_SPLIT_PRECISION` but `referral_fee_bps + other_fees > FEE_SPLIT_PRECISION`. could mark as a TODO to fetch and cache the other fees (stored in the express relay metadata account)

auction-server/src/auction/service/verification.rs
```diff
@@ -434,19 +437,21 @@ impl Service<Svm> {
             .iter()
             .filter(|instruction| {
                 let program_id = instruction.program_id(transaction.message.static_account_keys());
-                if *program_id != self.config.chain_config.express_relay.program_id {
-                    return false;
-                }
-
-                instruction.data.starts_with(&discriminator)
```
anihamde, line 441: unless there's something you're concerned about, i think better to filter for this here? ensure that only 1 call to submit_bid or 1 call to swap, and make sure that calls to both don't happen in the higher-level function?

the only reason is that there could be in the future other instructions that can be called alongside either of the bid instructions

auction-server/src/auction/service/verification.rs
```diff
@@ -533,82 +529,206 @@ impl Service<Svm> {
...
+                        None,
+                        output_token.token,
+                        Some(output_token.amount),
+                    ),
+                };
+                if let Some(expected_input_amount) = expected_input_amount {
+                    if expected_input_amount != swap_data.amount_input {
+                        return Err(RestError::BadParameters(
+                            format!(
+                                "Input amount in swap opportunity {} does not match the input amount in the swap instruction {}",
```
anihamde, line 622: nit but i think you can make the error messages a bit clearer to show that the searcher messed something up. Something like "Input amount in the submitted instruction is incorrect, does not match the input amount in the swap opportunity"

auction-server/src/auction/service/verification.rs
```diff
@@ -533,82 +529,206 @@ impl Service<Svm> {
...
+                    if expected_output_amount != swap_data.amount_output {
+                        return Err(RestError::BadParameters(
+                            format!(
+                                "Output amount in swap opportunity {} does not match the output amount in the swap instruction {}",
+                                expected_output_amount, swap_data.amount_output
+                            ),
+                        ));
+                    }
+                }
+                if expected_input_token != mint_input {
```
anihamde, line 638: nit but maybe these messages should go first before the amounts?

auction-server/src/auction/service/verification.rs
```diff
@@ -533,82 +529,206 @@ impl Service<Svm> {
...
+
+                if swap_data.referral_fee_bps != opp_swap_data.referral_fee_bps {
+                    return Err(RestError::BadParameters(
+                        format!(
+                            "Referral fee bps in swap opportunity {} does not match the referral fee bps in the swap instruction {}",
+                            opp_swap_data.referral_fee_bps, swap_data.referral_fee_bps
+                        ),
+                    ));
+                }
+                if user_wallet != opp_swap_data.user_wallet_address {
```
anihamde, line 664: this is probably the first one i would show, before referral fee and token mints and amounts

auction-server/src/auction/service/verification.rs
```diff
@@ -533,82 +529,206 @@ impl Service<Svm> {
...
+                            "User wallet address in swap opportunity {} does not match the user wallet address in the swap instruction {}",
+                            opp_swap_data.user_wallet_address, user_wallet
+                        ),
+                    ));
+                }
+                let (fee_token, fee_token_program) = match swap_data.fee_token {
+                    FeeToken::Input => (mint_input, opp_swap_data.input_token_program),
+                    FeeToken::Output => (mint_output, opp_swap_data.output_token_program),
+                };
+                if opp_swap_data.fee_token != swap_data.fee_token {
```
anihamde, line 676: i think this should also be above referral fee

auction-server/src/auction/service/verification.rs
```diff
@@ -533,82 +529,206 @@ impl Service<Svm> {
...
+                            "Fee token in swap opportunity {:?} does not match the fee token in the swap instruction {:?}",
+                            opp_swap_data.fee_token, swap_data.fee_token
+                        ),
+                    ));
+                }
+
+                let associated_token_program = Pubkey::from_str(
+                    "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL",
+                )
+                .map_err(|e| {
```
anihamde, line 688: this error feels kinda weird to me here, given that all the other errors are related to user input errors, makes it a bit confusing to read. maybe you could use this? https://docs.rs/solana-pubkey/2.1.9/src/solana_pubkey/lib.rs.html#309

auction-server/src/auction/service/verification.rs
```diff
@@ -533,82 +529,206 @@ impl Service<Svm> {
...
-                        QuoteTokens::OutputTokenSpecified {
-                            input_token:  mint_input,
-                            output_token: TokenAmountSvm {
-                                token:  mint_output,
-                                amount: swap_data.amount_output,
-                            },
-                        },
-                    ),
-                };
+                if router_fee_receiver_ta != expected_router_fee_receiver_ta {
```
anihamde, line 716: generally if we are manually checking all of these accounts, i'd rather simulate it at some point and return the error to them if there is an error

sdk/js/src/examples/simpleSearcherSvm.ts
```diff
@@ -89,12 +92,12 @@ export class SimpleSearcherLimo {
...
 
   /**
-   * Generates a bid for a given opportunity.
+   * Generates a bid for a given limo opportunity.
    * The transaction in this bid transfers assets from the searcher's wallet to fulfill the limit order.
    * @param opportunity The SVM opportunity to bid on.
    * @returns The generated bid object.
    */
-  async generateBid(opportunity: OpportunitySvm): Promise<BidSvm> {
+  async generateBidLimo(opportunity: OpportunitySvmLimo): Promise<BidSvm> {
```
anihamde, line 100: nit, i still prefer limo and swap examples to live in two different files bc they are completely different functions

sdk/js/src/types.ts
```diff
@@ -124,9 +122,41 @@ export type OpportunitySvm = {
...
+  permissionAccount: PublicKey;
+  routerAccount: PublicKey;
+  userWalletAddress: PublicKey;
+  feeToken: "input_token" | "output_token";
+  referralFeeBps: number;
+  // TODO: maybe type should be camelCase too?
```
anihamde, line 136: what's not camel case?

sdk/js/src/types.ts
```diff
@@ -124,9 +122,41 @@ export type OpportunitySvm = {
...
+  tokens: (
+    | {
+        inputToken: PublicKey;
```
anihamde, line 139: maybe make a type for this?

auction-server/api-types/src/bid.rs
```diff
@@ -265,6 +266,38 @@ pub struct BidCreateSvm {
...
+    #[schema(example = "SGVsbG8sIFdvcmxkIQ==", value_type = String)]
+    #[serde(with = "crate::serde::transaction_svm")]
+    pub transaction:    VersionedTransaction,
+    /// The id of the swap opportunity to bid on.
+    #[schema(example = "obo3ee3e-58cc-4372-a567-0e02b2c3d479", value_type = String)]
+    pub opportunity_id: OpportunityId,
+    /// The bid type. Should be "swap"
+    #[schema(example = "swap")]
+    #[serde(rename = "type")]
+    pub _type:          BidCreateSwapSvmTag, // this is mainly to distinguish next types of bids in the future
```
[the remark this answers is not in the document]
m30m: let's say we have another bid type later on with the exact same fields. This field helps us distinguish between them. Right now, since the structures are not quite equal we can distinguish from structure but this may not always be the case

auction-server/src/opportunity/api.rs
```diff
@@ -207,20 +206,14 @@ pub async fn get_opportunities(
     (status = 404, description = "No quote available right now", body = ErrorBodyResponse),
 ),)]
 pub async fn post_quote(
-    auth: Auth,
```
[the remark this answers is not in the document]
m30m: spoke about this with Dani,
At the end of the day, we don't want integrators to spin up a proxy to use this API so we should plan for DoS attacks.

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -55,7 +70,7 @@ pub struct OpportunitySvmProgramSwap {
 #[derive(Debug, Clone, PartialEq)]
 pub enum OpportunitySvmProgram {
     Limo(OpportunitySvmProgramLimo),
-    SwapKamino(OpportunitySvmProgramSwap),
```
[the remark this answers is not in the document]
m30m: I think this should happen in data analysis stage and not on the code. Do we want to update the code and add another enum variant per entity that is integrated?

auction-server/src/opportunity/service/get_quote.rs
m30m: I did this to avoid confusion between the main permission_key concept we have which is `permission_key = router + permission_account`.

auction-server/src/opportunity/service/get_quote.rs
m30m: agreed, added a TODO

auction-server/src/opportunity/service/get_quote.rs
m30m: Added a _virtual_ in the name

auction-server/src/auction/service/verification.rs
m30m: I guess the previous version was implemented that way because we were trying to find the type by looking at the transaction and looking at different match arms which makes sense.
But now, we distinguish directly based on input structure and then, for now, we want to make sure only that specific instruction exists in the tx, so this function seemed more ergonomic. We can change it to another format if the product requirement changes of course.

auction-server/src/auction/service/verification.rs
m30m: this function does not exist on our version but I used something similar.

sdk/js/src/types.ts
m30m: "input_token" and some other type values

auction-server/src/auction/service/verification.rs
m30m: we do simulate all these bids

anihamde: then why check these things here? for extra clarity for users?

auction-server/src/auction/service/verification.rs
```diff
@@ -594,13 +594,13 @@ impl Service<Svm> {
...
-                        output_token: opp_buy_token.clone(),
+                    (0, _) => QuoteTokens::InputTokenSpecified {
+                        output_token: opp_sell_token.token,
+                        input_token:  opp_buy_token.clone(),
                     },
-                    (_, 0) => QuoteTokens::InputTokenSpecified {
-                        input_token:  opp_sell_token.clone(),
-                        output_token: opp_buy_token.token,
+                    (_, 0) => QuoteTokens::OutputTokenSpecified {
+                        output_token: opp_sell_token.clone(),
```
anihamde: tiniest of nits, can you order input_token first for consistency? or output_token first is fine but let's just follow one order throughout

auction-server/src/auction/service/verification.rs
```diff
@@ -483,6 +494,188 @@ impl Service<Svm> {
         }
     }
 
+    async fn check_svm_swap_bid_fields(
```
anihamde, line 500: i like packaging the validations all here, you could pass in SwapAccounts here to cut the redundancy, but i think it's fine this way too

auction-server/src/auction/service/verification.rs
```diff
@@ -759,6 +804,32 @@ impl Service<Svm> {
...
+            .first()
+            .expect("Swap opportunity sell tokens must not be empty");
+        let opp_buy_token = opp
+            .core_fields
+            .buy_tokens
+            .first()
+            .expect("Swap opportunity buy tokens must not be empty");
+        match (opp_sell_token.amount, opp_buy_token.amount) {
+            (0, _) => QuoteTokens::InputTokenSpecified {
+                output_token: opp_sell_token.token,
```
anihamde, line 810: i think output_token is the buy_token from the standpoint of the searcher

anihamde (approved):

auction-server/src/auction/service/verification.rs
m30m: I think it's the responsibility of this function to extract such info.

auction-server/src/auction/service/verification.rs
m30m: yes. Also, I'd rather check everything twice than thinking about which of these are actually checked via the anchor constraints.

PR #328, opened 2025-01-15
add token programs to server

anihamde:
This PR adds the pulling and caching of token programs for mints to the server side. The token programs are passed in the QuoteSvm object to the caller.

### DIFF
--- a/auction-server/api-types/src/opportunity.rs
+++ b/auction-server/api-types/src/opportunity.rs
@@ -224,9 +224,11 @@ pub enum OpportunityCreateProgramParamsV1Svm {
         /// The referral fee in basis points.
         #[schema(example = 10, value_type = u16)]
         referral_fee_bps:     u16,
+        /// The token program of the input mint.
         #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
         #[serde_as(as = "DisplayFromStr")]
         input_token_program:  Pubkey,
+        /// The token program of the output mint.
         #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
         #[serde_as(as = "DisplayFromStr")]
         output_token_program: Pubkey,
--- a/auction-server/api-types/src/opportunity.rs
+++ b/auction-server/api-types/src/opportunity.rs
@@ -220,10 +220,16 @@ pub enum OpportunityCreateProgramParamsV1Svm {
         /// The user wallet address which requested the quote from the wallet.
         #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
         #[serde_as(as = "DisplayFromStr")]
-        user_wallet_address: Pubkey,
+        user_wallet_address:  Pubkey,
         /// The referral fee in basis points.
         #[schema(example = 10, value_type = u16)]
-        referral_fee_bps:    u16,
+        referral_fee_bps:     u16,
+        #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
+        #[serde_as(as = "DisplayFromStr")]
+        input_token_program:  Pubkey,
+        #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
+        #[serde_as(as = "DisplayFromStr")]
+        output_token_program: Pubkey,
     },
 }
 
@@ -266,6 +272,7 @@ pub enum OpportunityCreateSvm {
 /// The input type for creating a new opportunity.
 #[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug)]
 #[serde(untagged)]
+#[allow(clippy::large_enum_variant)]
 pub enum OpportunityCreate {
     #[schema(title = "evm")]
     Evm(OpportunityCreateEvm),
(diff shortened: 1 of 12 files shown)


auction-server/src/state.rs
```diff
@@ -117,7 +122,38 @@ impl ChainStoreSvm {
             log_sender: tx,
             _dummy_log_receiver: rx,
             config,
+            token_program_cache: RwLock::new(HashMap::new()),
+        }
+    }
+
+    pub async fn get_token_program(
```
m30m, line 129: I would limit the output of this function to be either TokenProgram or Token22Program

auction-server/api-types/src/opportunity.rs
```diff
@@ -220,10 +220,16 @@ pub enum OpportunityCreateProgramParamsV1Svm {
...
         #[serde_as(as = "DisplayFromStr")]
-        user_wallet_address: Pubkey,
+        user_wallet_address:  Pubkey,
         /// The referral fee in basis points.
         #[schema(example = 10, value_type = u16)]
-        referral_fee_bps:    u16,
+        referral_fee_bps:     u16,
+        #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
+        #[serde_as(as = "DisplayFromStr")]
+        input_token_program:  Pubkey,
```
danimhr, line 229: Add docstring plz.

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -137,13 +139,36 @@ impl Service<ChainTypeSvm> {
             }],
         };
 
+        let chain_store = self
```
danimhr, line 142: The store is going to be deprecated. Please add anything that you need to the ChainConfig in the opportunity module and then use the `self.get_config` with the chain_id you need. It's not a good idea to use the store anymore.

auction-server/src/state.rs
danimhr, line 129: This file is deprecated and we want to delete it. Please put this functionality in one the modules (Maybe opportunity) and provide a service for it.

danimhr:
The pre-commit is failing. please take care of it :D

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -99,6 +100,7 @@ impl Service<ChainTypeSvm> {
         &self,
         quote_create: entities::QuoteCreate,
         program: &ProgramSvm,
+        config: &ConfigSvm,
```
danimhr, line 103: Instead of passing config, you can call self.get_config

auction-server/src/opportunity/service/mod.rs
```diff
@@ -194,16 +207,68 @@ impl ConfigSvm {
...
+                        ),
+                        token_program_cache:     RwLock::new(HashMap::new()),
+                        accepted_token_programs: chain_store.config.accepted_token_programs.clone(),
                     },
                 )
             })
             .collect())
     }
+
+    pub async fn get_token_program(&self, mint: Pubkey) -> anyhow::Result<Pubkey> {
```
danimhr, line 229: You should put the cache inside the repo and make a service for `get_token_program`

auction-server/src/opportunity/service/mod.rs
```diff
@@ -194,16 +207,68 @@ impl ConfigSvm {
...
-                        auction_service: RwLock::new(None),
+                        auction_service:         RwLock::new(None),
+                        rpc_client:              TracedSenderSvm::new_client(
+                            chain_id.clone(),
+                            chain_store.config.rpc_read_url.as_str(),
+                            chain_store.config.rpc_timeout,
+                            RpcClientConfig::with_commitment(CommitmentConfig::processed()),
+                        ),
+                        token_program_cache:     RwLock::new(HashMap::new()),
+                        accepted_token_programs: chain_store.config.accepted_token_programs.clone(),
```
danimhr, line 222: Why do we need to whitelist the tokens?

anihamde: this is whitelisted token programs, a la TokenProgram and Token22

auction-server/src/opportunity/service/mod.rs
```diff
@@ -194,16 +207,68 @@ impl ConfigSvm {
...
                     },
                 )
             })
             .collect())
     }
+
+    pub async fn get_token_program(&self, mint: Pubkey) -> anyhow::Result<Pubkey> {
+        if let Some(program) = self.token_program_cache.read().await.get(&mint) {
```
danimhr, line 230: You can do this part like below:

```
let token_program = match self.token_program_cache.read().await.get(&mint) {
   Ok(token_program) => token_program,
   None => {
       self
            .rpc_client
            .get_account(&mint)
            .await
            .map_err(|err| {
                tracing::error!(
                    "Failed to retrieve owner program for mint account {mint}: {:?}",
                    err,
                    mint = mint
                );
                anyhow!(
                    "Failed to retrieve owner program for mint account {mint}: {:?}",
                    err,
                    mint = mint
                )
            })?
            .owner;
            self.token_program_cache
            .write()
            .await
            .insert(mint, token_program);
            token_program
   }
};
if !self.accepted_token_programs.contains(&token_program) {
            return Err(anyhow!(
                "Token program {program} for mint account {mint} is not an approved token program",
                program = token_program,
                mint = mint
            ));
        }
        Ok(token_program)

```

danimhr: With the code above, you basically remove the duplication for checking if the token_program is an accepted address or not.

auction-server/src/opportunity/service/get_token_program.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl Service<ChainTypeSvm> {
+    /// Find the token program for a given mint.
+    /// Pulls from the cache if already present, otherwise queries the RPC and saves in the cache.
+    pub async fn get_token_program(
+        &self,
+        input: GetTokenProgramInput,
+    ) -> Result<Pubkey, RestError> {
+        let config = self.get_config(&input.chain_id)?;
+        let cache_read = self.repo.in_memory_store.token_program_cache.read().await;
```
danimhr, line 26: It's better to add a functionality in repo fro this. The cache will be dropped automatically if you do so.
I mean like `get_token_program.rs`
We can have a function to read the cache in the repo instead of getting it here,

danimhr (approved):

PR #329, opened 2025-01-15
test: add fees overflow test

guibescos:
Added one last test. Had to refactor `assert_custom_error` because it's a new type of error.

### DIFF
--- a/contracts/svm/testing/src/helpers.rs
+++ b/contracts/svm/testing/src/helpers.rs
@@ -3,7 +3,7 @@ use {
     solana_sdk::{
         instruction::{
             Instruction,
-            InstructionError::Custom,
+            InstructionError,
         },
         native_token::LAMPORTS_PER_SOL,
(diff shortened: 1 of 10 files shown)


contracts/svm/testing/src/helpers.rs
```diff
@@ -56,16 +54,15 @@ pub fn warp_to_unix(svm: &mut litesvm::LiteSVM, unix_timestamp: i64) {
     svm.set_sysvar(&clock);
 }
 
-pub fn assert_custom_error(error: TransactionError, instruction_index: u8, custom_error: u32) {
+pub fn assert_custom_error(
+    error: TransactionError,
+    instruction_index: u8,
+    custom_error: InstructionError,
```
anihamde, line 60: maybe we could rename this `instruction_error`? since it's no longer just custom errors, it includes like `ArithmeticOverflow` and the like

anihamde (approved):

PR #331, opened 2025-01-17
feat: swap app

guibescos:
The goal of this PR is to put all the files in place for the swap app.


apps/swap/src/config/server.ts
```diff
@@ -0,0 +1,7 @@
+import "server-only";
+
+/* eslint-disable n/no-process-env */
+
+export const SOLANA_RPC = process.env.SOLANA_RPC || "https://localhost:8899";
```
m30m, line 5: shouldn't these be `http://`

apps/swap/src/app/page.tsx
```diff
@@ -0,0 +1,55 @@
...
+      console.error(error);
+    });
+  }, [expressRelayClient, publicKey, signTransaction, connection]);
+
+  return (
+    <main>
+      <WalletMultiButton />
+      <WalletDisconnectButton />
+      <p>Public Key: {publicKey?.toBase58()}</p>
+      {publicKey && (
```
m30m, line 50: I think this boolean logic should be the same as what you check at the beginning of handleClick and should include the signTransaction check

m30m (approved):

apps/swap/src/config/server.ts
guibescos: fixed

apps/swap/src/app/page.tsx
guibescos: fixed

PR #335, opened 2025-01-18
fix: better api format with more docs

m30m:


auction-server/api-types/src/opportunity.rs
```diff
@@ -178,13 +178,11 @@ pub enum OpportunityCreateEvm {
...
-    /// The token contract address.
-    #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
+    /// The token mint address.
+    #[schema(example = "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v", value_type = String)]
     #[serde_as(as = "DisplayFromStr")]
     pub token:  Pubkey,
-    /// The token amount, represented in the smallest unit of the respective token:
-    /// - For Solana, it is measured in lamports.
-    /// - For other tokens, it follows the smallest denomination of that token.
+    /// The token amount, represented in the smallest unit of the respective token.
```
anihamde, line 185: "the smallest unit of the respective token" is a bit confusing, i personally find the more verbose docstring clearer. maybe find another way to express that?

anihamde (approved):

PR #336, opened 2025-01-20
Fix/swap bid restrictions

anihamde:
This PR adds some restrictions to the instructions in the transaction of a swap bid. Acceptable instructions are:

- `ComputeBudget`
- `AssociatedTokenProgram`, createAta, createAtaIdempotent
- express relay program

### DIFF
--- a/auction-server/config.sample.yaml
+++ b/auction-server/config.sample.yaml
@@ -18,4 +18,3 @@ chains:
     accepted_token_programs:
       - TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA
       - TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb
-    associated_token_program_id: ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL
--- a/auction-server/config.sample.yaml
+++ b/auction-server/config.sample.yaml
@@ -18,3 +18,4 @@ chains:
     accepted_token_programs:
       - TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA
       - TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb
+    associated_token_program_id: ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL
--- a/auction-server/src/auction/service/verification.rs
+++ b/auction-server/src/auction/service/verification.rs
@@ -503,13 +503,12 @@ impl Service<Svm> {
         &self,
         tx: &VersionedTransaction,
     ) -> Result<(), RestError> {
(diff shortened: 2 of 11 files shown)


auction-server/config.sample.yaml
```diff
@@ -18,3 +18,4 @@ chains:
     accepted_token_programs:
       - TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA
       - TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb
+    associated_token_program_id: ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL
```
guibescos, line 21: mehhh, why not harcode this: I don't think this will ever change

auction-server/src/auction/service/verification.rs
```diff
@@ -498,6 +499,64 @@ impl Service<Svm> {
...
+                        ))
+                    })
+            })?;
+
+        Ok(())
+    }
+
+    fn validate_swap_transaction_instruction(
+        &self,
+        accounts: &[Pubkey],
```
guibescos, line 525: is this compatible with account lookup table?

auction-server/src/auction/service/verification.rs
```diff
@@ -498,6 +499,64 @@ impl Service<Svm> {
...
+                })?;
+            match ix_parsed {
+                AssociatedTokenAccountInstruction::Create => Ok(()),
+                AssociatedTokenAccountInstruction::CreateIdempotent => Ok(()),
+                _ => Err(RestError::BadParameters(format!(
+                    "Invalid associated token account instruction: {:?}",
+                    ix_parsed
+                ))),
+            }
+        } else if *program_id == self.config.chain_config.express_relay.program_id {
```
guibescos, line 550: would this work as a match statement?

auction-server/src/auction/service/verification.rs
anihamde: don't need lookup table right now because all the accounts we care about are programs/signers

auction-server/config.sample.yaml
anihamde: for other SVM chains in the future potentially

anihamde: not a big deal but i think this is good practice

auction-server/src/auction/service/verification.rs
anihamde: originally had it as that but it was a bit weird. `match` doesn't work with non-constant expressions, so you have to define the expressions outside of the match statement. i can make that change and we can choose whichever way you prefer

auction-server/src/auction/service/verification.rs
guibescos: got it

guibescos (approved):
lgtm

PR #337, opened 2025-01-20
feat: swap-python-sdk

m30m:


sdk/python/express_relay/models/svm.py
```diff
@@ -186,7 +185,21 @@ def __get_pydantic_json_schema__(
 SvmSignature = Annotated[_SvmSignature, _SignaturePydanticAnnotation]
 
 
-class BidSvm(BaseModel):
+class SwapBidSvm(BaseModel):
+    """
+    Attributes:
+        transaction: The transaction including the bid
```
danimhr, line 191: The transaction including the bid -> The transaction including the swap instruction

I'm not sure btw :D

PR #338, opened 2025-01-20
bump: prettier

guibescos:
Let's bump the prettier pre-commit hook to match the pre-commit that we have in package.json
This is required for CI to work on https://github.com/pyth-network/per/pull/331


m30m (approved):
All these `,`s made our code prettier 🙇

PR #339, opened 2025-01-20
fix: js quote sdk

guibescos:
Fix two current bugs from testing the swap flow.


anihamde (approved):
looks good, but make sure the vercel deployment works

sdk/js/src/const.ts
```diff
@@ -26,7 +26,7 @@ export const OPPORTUNITY_ADAPTER_CONFIGS: Record<
 export const SVM_CONSTANTS: Record<string, SvmConstantsConfig> = {
   "development-solana": {
     expressRelayProgram: new PublicKey(
-      "PytERJFhAKuNNuaiXkApLfWzwNwSNDACpigT3LwQfou"
+      "stag1NN9voD7436oFvKmy1kvRZYLLW8drKocSCt2W79"
```
m30m, line 29: this is gonna break tilt scripts because we use `development-solana` in local config but we deploy at `PytERJFhAKuNNuaiXkApLfWzwNwSNDACpigT3LwQfou` locally. Probably better to create a 3rd `development-local` entry both in js and python.

PR #340, opened 2025-01-20
fix: forgot this

guibescos:


sdk/js/src/types.ts
```diff
@@ -2,6 +2,7 @@ import { Address, Hex } from "viem";
 import type { components } from "./serverTypes";
 import { PublicKey, Transaction } from "@solana/web3.js";
```
anihamde, line 3: do you still need Transaction?

anihamde (approved):

sdk/js/src/types.ts
guibescos: seems like this is still used

PR #343, opened 2025-01-20
feat: small ui polish

m30m:


apps/swap/src/app/page.tsx
```diff
@@ -8,49 +8,81 @@ import {
...
+      setLog((log) => [...log, `Selling ${amount / 1e6} USDT for USDC`]);
       const quote = await expressRelayClient.getQuote({
         chainId: "development-solana",
         inputTokenMint: USDC,
         outputTokenMint: USDT,
         router: publicKey,
         userWallet: publicKey,
         specifiedTokenAmount: {
-          amount: Math.floor(Math.random() * 100000), // random to avoid same opportunity submitted recently error
+          amount: amount,
```
danimhr, line 38: we dont need to pass`amount: amount`
you can simply pass `amount`

apps/swap/src/app/page.tsx
```diff
@@ -8,49 +8,81 @@ import {
...
-      )}
+      <div className="m-auto w-2/4">
+        <h1>Express Relay Swap testing UI</h1>
+        <div className="my-3">
+          <WalletMultiButton />
+        </div>
+        <WalletDisconnectButton />
+        <p>Public Key: </p>
+        <pre>{publicKey?.toBase58()}</pre>
+        {canSwap && (
```
danimhr, line 75: Maybe it's nice to add loading as well here. In the handleClick function we can set the loading state and disable this button

apps/swap/src/app/page.tsx
```diff
@@ -8,49 +8,81 @@ import {
...
-      <WalletMultiButton />
-      <WalletDisconnectButton />
-      <p>Public Key: {publicKey?.toBase58()}</p>
-      {canSwap && (
-        <Button onClick={handleClick}>Click me to sell 1 USDT for USDC</Button>
-      )}
+      <div className="m-auto w-2/4">
+        <h1>Express Relay Swap testing UI</h1>
+        <div className="my-3">
+          <WalletMultiButton />
```
guibescos, line 70: feels inconsistent that one Button is inside this div but the other one isn't

guibescos (approved):
great improvements!

apps/swap/src/app/page.tsx
m30m: came up with better css

apps/swap/src/app/page.tsx
m30m: it's nice indeed but maybe too soon to add 😅

PR #344, opened 2025-01-20
feat: Add a new config for local tilt since the program addresses are now different

m30m:


integration.py
```diff
@@ -42,7 +42,7 @@ def main():
     adapter_factory_contract: {latest_env['adapterFactory']}
     legacy_tx: false
     poll_interval: 1
-  development-solana:
+  development-solana-local:
```
anihamde, line 45: can we call "local" development and the staging "staging" or "testnet"? or i'm fine with `solana-local` as well, but just makes it easier to distinguish

guibescos: I like this

m30m: you mean this?
local
solana-staging
solana

since we still support other chains, I'd like to keep the chain name in the chain_id and add some pre/post fix to it. So maybe:
solana-local
solana-staging
solana
?

guibescos: makes sense to me

guibescos (approved):

PR #346, opened 2025-01-21
fix:Only build in ci what is needed

m30m:


Dockerfile.node
```diff
@@ -19,7 +19,7 @@ USER per
 FROM builder-base AS builder
 ARG package
 ENV CI true
-RUN pnpm -r build
+RUN pnpm -r --filter $package build
```
guibescos, line 22: ohhh I think a better one is like `pnpm --filter $package..` which will build a package and it's dependencies

danimhr: Yeah the reasion not adding this filter is to make sure we are building all deps as well as the package

m30m: TIL

guibescos (approved):
send it

PR #351, opened 2025-01-22
Fix/simulation timestamp

anihamde:
This PR fixes the simulation timestamp to the current time after fetching the accounts from the RPC. Assuming insignificant clock drift, the current timestamp should sufficiently approximate the clock time on chain.

### DIFF
--- a/auction-server/src/auction/service/simulator.rs
+++ b/auction-server/src/auction/service/simulator.rs
@@ -28,6 +28,7 @@ use {
         account_utils::StateMut,
         address_lookup_table::state::AddressLookupTable,
         bpf_loader_upgradeable::UpgradeableLoaderState,
+        clock::Clock,
         commitment_config::CommitmentConfig,
         pubkey::Pubkey,
         signature::Signature,
@@ -320,7 +321,11 @@ impl Simulator {
     }
 
     #[tracing::instrument(skip_all)]
-    fn setup_lite_svm(&self, accounts_config_with_context: &Response<AccountsConfig>) -> LiteSVM {
+    fn setup_lite_svm(
+        &self,
+        accounts_config_with_context: &Response<AccountsConfig>,
+        timestamp: Option<i64>,
+    ) -> LiteSVM {
         let mut svm = LiteSVM::new()
             .with_sigverify(false)
             .with_blockhash_check(false)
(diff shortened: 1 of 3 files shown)


auction-server/src/auction/service/simulator.rs
```diff
@@ -320,18 +324,34 @@ impl Simulator {
...
+        if let Some(timestamp) = timestamp {
+            svm = self.warp_to_timestamp(svm, timestamp);
+        }
         accounts_config_with_context.value.apply(&mut svm);
         svm
     }
 
+    /// Warps the LiteSVM object clock to the given timestamp
+    /// This is necessary because LiteSVM does not natively support warping to a timestamp
+    fn warp_to_timestamp(&self, mut svm: LiteSVM, timestamp: i64) -> LiteSVM {
```
m30m, line 348: why not get `&mut svm `and modify in-place?

auction-server/src/auction/service/simulator.rs
```diff
@@ -395,15 +422,24 @@ impl Simulator {
     /// Right now, for simplicity, the method assume the bids are sorted, and tries to submit them in order
     /// and only return the ones that are successfully submitted.
     #[tracing::instrument(skip_all)]
-    pub async fn optimize_bids(&self, bids: &[Bid<Svm>]) -> RpcResult<Vec<Bid<Svm>>> {
+    pub async fn optimize_bids(
+        &self,
+        bids: &[Bid<Svm>],
+        warp_to_timestamp: bool,
```
m30m, line 428: no need to make it a parameter if always true

auction-server/src/auction/service/simulator.rs
```diff
@@ -369,6 +389,7 @@ impl Simulator {
     pub async fn simulate_transaction(
         &self,
         transaction: &VersionedTransaction,
+        warp_to_timestamp: bool,
```
m30m, line 392: no need to make it a parameter if always true

auction-server/src/auction/service/simulator.rs
```diff
@@ -320,18 +324,34 @@ impl Simulator {
     }
 
     #[tracing::instrument(skip_all)]
-    fn setup_lite_svm(&self, accounts_config_with_context: &Response<AccountsConfig>) -> LiteSVM {
+    fn setup_lite_svm(
```
m30m, line 327: I think we can just set it to the current time in this function for now and don't need to change any function signature

m30m (approved):

PR #355, opened 2025-01-22
Fix/rust sdk

danimhr:
This PR aims to fix rust sdk dependency issue and auction server swap router check bug

### DIFF
--- a/sdk/rust/src/svm.rs
+++ b/sdk/rust/src/svm.rs
@@ -100,7 +100,7 @@ pub struct Svm {
     client: RpcClient,
 }
 
-// After anchor releas version 0.31, we can remove all of the following functions
+// After anchor releases version 0.31, we can remove all of the following functions
 // And use the ones from the express-relay contract crate
 pub const SEED_METADATA: &[u8] = b"metadata";
 pub const SEED_CONFIG_ROUTER: &[u8] = b"config_router";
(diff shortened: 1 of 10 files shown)


sdk/rust/src/svm.rs
```diff
@@ -107,6 +100,56 @@ pub struct Svm {
...
+    pub fee_token:        FeeToken,
+}
+
+#[derive(BorshSerialize, BorshDeserialize)]
+struct SubmitBidArgs {
+    pub deadline:   i64,
+    pub bid_amount: u64,
+}
+
+pub fn deserialize_metadata(data: Vec<u8>) -> Result<ExpressRelayMetadata, ClientError> {
```
guibescos, line 142: should we remove this from `express_relay/src/sdk`

danimhr: I think we are going to use it later, so that's why I didnt delete the functionality

sdk/rust/src/svm.rs
```diff
@@ -129,9 +172,7 @@ impl Svm {
             .map_err(|_| {
                 ClientError::SvmError("Failed to fetch express relay metadata".to_string())
             })?;
-        deserialize_metadata(data).map_err(|_| {
-            ClientError::SvmError("Failed to deserialize express relay metadata".to_string())
-        })
+        deserialize_metadata(data)
```
guibescos, line 175: since this is the only place you call this method, should we inline it the code here?

not sure searchers should ever call `deserialize_metadata`

guibescos (approved):

PR #358, opened 2025-01-22
Feat: make fees valid again

m30m:
- Fetch metadata using RPC
- Use it to calculate and broadcast fees correctly
- Removed OpportunityCreateProgramParamsV1Svm::Swap variant because it did not make sense 🤔


sdk/js/src/types.ts
```diff
@@ -130,25 +130,27 @@ export type OpportunitySvmLimo = {
 
 export type SvmSwapTokens = (
   | {
-      inputToken: PublicKey;
-      outputToken: TokenAmountSvm;
+      outputAmount: bigint;
```
anihamde, line 133: maybe make this a separate type?

sdk/js/src/svm.ts
```diff
@@ -151,14 +153,31 @@ export async function constructSwapInstruction(
...
+    swapOpportunity.tokens.type === "input_specified" &&
+    swapOpportunity.feeToken === "output_token"
+  ) {
+    // scale bid amount by FEE_SPLIT_PRECISION/(FEE_SPLIT_PRECISION-fees) to account for fees
+    const denominator = FEE_SPLIT_PRECISION.sub(
+      new anchor.BN(
+        swapOpportunity.platformFeeBps + swapOpportunity.referralFeeBps,
+      ),
+    );
+    const numerator = bidAmount.mul(FEE_SPLIT_PRECISION);
```
anihamde, line 166: similar comments to above in terms of configuring this calculation

anihamde (commented):
didn't look too closely at the SDK changes; will do so more closely in the next review

auction-server/src/auction/service/mod.rs
```diff
@@ -74,6 +74,7 @@ pub mod get_bid;
 pub mod get_bids;
 pub mod get_live_bids;
 pub mod get_permission_keys_for_auction;
+pub mod get_program_id;
```
danimhr, line 77: This is name is very general. Maybe add express_relay to it?

auction-server/src/opportunity/service/get_express_relay_metadata.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// Fetches the express relay program metadata for fee calculations
+    /// Uses one hour cache to avoid repeated RPC calls
+    pub async fn get_express_relay_metadata(
+        &self,
+        input: GetExpressRelayMetadata,
+    ) -> Result<ExpressRelayMetadata, RestError> {
+        let config = self.get_config(&input.chain_id)?;
+        let program_id = config.get_auction_service().await.get_program_id();
+        let seed = express_relay_svm::state::SEED_METADATA;
+        let metadata_address = Pubkey::find_program_address(&[seed], &program_id).0;
```
danimhr, line 30: I think a better place to put this metadata is in the Kernel. We can cache it there and update whenever we need to update it. If it's hard to implement that way, please add a TODO for later

danimhr (commented):
I think the overall api objects can be a little bit different.

In general I think we should broadcast the quote to searchers, as same as the value we received in the request. We can add a `fee` field to some of the types to make sure searchers and protocol easily can calculate anything they need based on the `fee` variable and the `quote`.

RN it's a bit confusing and hard to split the fee calculation into another function which gets a Quote / Opportunity of Entity as input argument.

auction-server/src/opportunity/repository/mod.rs
```diff
@@ -55,8 +60,9 @@ pub struct InMemoryStoreEvm {
     pub spoof_info:  RwLock<HashMap<Address, entities::SpoofState>>,
 }
 pub struct InMemoryStoreSvm {
-    pub core_fields:         InMemoryStoreCoreFields<entities::OpportunitySvm>,
-    pub token_program_cache: RwLock<HashMap<Pubkey, Pubkey>>,
+    pub core_fields:            InMemoryStoreCoreFields<entities::OpportunitySvm>,
+    pub token_program_cache:    RwLock<HashMap<Pubkey, Pubkey>>,
+    pub express_relay_metadata: RwLock<Option<(Instant, ExpressRelayMetadata)>>,
```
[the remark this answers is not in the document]
m30m: the api Instant provides is very easy for the current usecase.

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -49,6 +51,25 @@ use {
     tokio::time::sleep,
 };
 
+// FeeToken and TokenSpecified combinations possible and how they are handled:
```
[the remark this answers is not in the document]
m30m: this is actually not true. In case `FeeToken=OutputToken, TokenSpecified=OutputTokenSpecified` the output_amount is after fees and we have a separate field for before_fees. Main rationale here was the main input/output amounts should show what is important for the searcher and not something internal.

auction-server/src/opportunity/repository/get_express_relay_metadata.rs
```diff
@@ -0,0 +1,31 @@
...
+
+impl Repository<InMemoryStoreSvm> {
+    pub async fn query_express_relay_metadata(&self) -> Option<ExpressRelayMetadata> {
+        let (insert_time, metadata) = self
+            .in_memory_store
+            .express_relay_metadata
+            .read()
+            .await
+            .clone()?;
+        if insert_time.elapsed() > MAX_METADATA_STALENESS {
```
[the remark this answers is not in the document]
m30m: added a TODO and removed the caching

auction-server/api-types/src/opportunity.rs
```diff
@@ -338,9 +319,13 @@ pub enum OpportunityParamsV1ProgramSvm {
         router_account: Pubkey,
 
         /// The referral fee in basis points.
-        #[schema(example = 10, value_type = u16)]
+        #[schema(example = 10)]
         referral_fee_bps: u16,
 
+        /// The platform fee in basis points.
+        #[schema(example = 10)]
+        platform_fee_bps: u64,
```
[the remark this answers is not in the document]
m30m: I used u64 to be consistent with the program state

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -120,16 +141,41 @@ impl Service<ChainTypeSvm> {
...
-                } => (input_token, 0, output_token.token, output_token.amount),
+            match (quote_create.tokens.clone(), fee_token.clone()) {
+                (
+                    entities::QuoteTokens::InputTokenSpecified {
+                        input_token,
+                        output_token,
+                    },
+                    entities::FeeToken::InputToken,
+                ) => {
+                    let denominator: u64 = FEE_SPLIT_PRECISION
```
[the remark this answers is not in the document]
anihamde: i think we should write a TODO to correct this at least. i think it's better to be accurate on this, and this will definitely be an audit issue whenever it gets audited. we don't have to do it now but at least something to add to backlog (and I don't think it will take too long to correct this)

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -120,17 +141,43 @@ impl Service<ChainTypeSvm> {
...
+                entities::QuoteTokens::InputTokenSpecified { input_token, .. },
+                entities::FeeToken::InputToken,
+            ) => {
+                // This is not exactly accurate and may overestimate the amount needed
+                // because of floor / ceil rounding errors.
+                let denominator: u64 = FEE_SPLIT_PRECISION
+                    - <u16 as Into<u64>>::into(quote_create.referral_fee_bps)
+                    - metadata.swap_platform_fee_bps;
+                let numerator = input_token.amount * FEE_SPLIT_PRECISION;
+                let amount_before_fees = numerator.div_ceil(denominator);
```
anihamde, line 170: slight nit, i think this would be better termed `amount_including_fees`. in general, i think using `including_fees` is clearer than `before_fees` which is more ambiguous in how it can be interpreted (before fees = with fees? without fees?)

anihamde: this applies to the `output_amount_before_fees` as well. i can think about this more in the api improvements and internal renamings, but maybe worth it to address it here

anihamde (approved):

auction-server/src/opportunity/service/get_quote.rs
m30m: renamed internally, you can address the fields externally

auction-server/api-types/src/opportunity.rs
```diff
@@ -318,6 +300,9 @@ pub enum OpportunityParamsV1ProgramSvm {
         #[schema(example = "DUcTi3rDyS5QEmZ4BNRBejtArmDCWaPYGfN44vBJXKL5", value_type = String)]
         #[serde_as(as = "DisplayFromStr")]
         order_address: Pubkey,
+        /// The slot where the opportunity params were fetched from using the RPC.
+        #[schema(example = 293106477, value_type = u64)]
+        slot:          Slot,
```
anihamde, line 305: this is good change, but it is backward incompatible. we should notify searchers of this in our next changelog

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -120,17 +138,43 @@ impl Service<ChainTypeSvm> {
...
+                input_token,
+                output_token,
+            } => (input_token, output_token.token),
+        };
+        let (input_amount, output_amount) = match (quote_create.tokens.clone(), fee_token.clone()) {
+            (
+                entities::QuoteTokens::InputTokenSpecified { input_token, .. },
+                entities::FeeToken::InputToken,
+            ) => {
+                // This is not exactly accurate and may overestimate the amount needed
```
anihamde, line 161: let's mark it as a TODO to fix this

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -237,20 +248,31 @@ impl From<OpportunitySvm> for api::OpportunitySvm {
...
-                        output_token: output_token.into(),
-                        input_token_program: program.input_token_program,
-                        output_token_program: program.output_token_program,
-                    },
+                    } => {
+                        let output_amount_excluding_fees = match program.fee_token {
+                            FeeToken::InputToken => output_token.amount,
+                            FeeToken::OutputToken => {
+                                let fees = (output_token.amount
+                                    * (program.platform_fee_bps
```
anihamde, line 263: this computation should be the `output_token.amount * program.platform_fee_bps / FEE_SPLIT_PRECISION + output_token.amount * program.referral_fee_bps / FEE_SPLIT_PRECISION`. and you shouldn't do `div_ceil`. just normal division, that will replicate exactly what happens in the smart contract

auction-server/api-types/src/opportunity.rs
```diff
@@ -364,40 +353,50 @@ pub enum FeeToken {
...
+        /// The amount that searcher will receive after deducting fees
+        output_amount:             u64,
+        /// The exact amount of output_token that the user wants to send in exchange
+        output_amount_before_fees: u64,
     },
 }
 
+#[serde_as]
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug, ToResponse)]
+pub struct QuoteTokensWithPrograms {
```
anihamde, line 386: slight nit, can we name this `QuoteTokensWithTokenPrograms`? ow it's a bit confusing: OpportunityProgramParams << QuoteTokensWithPrograms :D

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -385,19 +428,37 @@ impl Service<ChainTypeSvm> {
...
+            })
+            .await?;
+
+        let fee_token = match swap_data.fee_token {
+            FeeToken::Input => input_token.token,
+            FeeToken::Output => output_token.token,
+        };
+        let compute_fees = |amount: u64| {
+            metadata
+                .compute_swap_fees(input.quote_create.referral_fee_bps, amount)
```
anihamde, line 443: maybe this function should be a separate service? maybe a TODO?

sdk/js/src/svm.ts
```diff
@@ -151,14 +153,31 @@ export async function constructSwapInstruction(
...
+    // add denominator - 1 to round up
```
anihamde, line 167: let's mark this as a TODO to fix as well

sdk/python/express_relay/client.py
```diff
@@ -584,19 +579,30 @@ def get_svm_swap_instructions(
...
+        if (
+            swap_opportunity.tokens.side_specified == "input"
+            and swap_opportunity.fee_token == "output_token"
+        ):
+            # scale bid amount by FEE_SPLIT_PRECISION/(FEE_SPLIT_PRECISION-fees) to account for fees
+            denominator = FEE_SPLIT_PRECISION - (
+                swap_opportunity.platform_fee_bps + swap_opportunity.referral_fee_bps
+            )
+            numerator = bid_amount * FEE_SPLIT_PRECISION
+            # add denominator - 1 to round up
```
anihamde, line 591: same comment here

sdk/rust/src/svm.rs
```diff
@@ -293,39 +299,37 @@ impl Svm {
...
-        ) = match swap_data.tokens {
+
+        let bid_amount = match (&swap_data.tokens.tokens, &swap_data.fee_token) {
+            // scale bid amount by FEE_SPLIT_PRECISION/(FEE_SPLIT_PRECISION-fees) to account for fees
+            (QuoteTokens::InputTokenSpecified { .. }, ApiFeeToken::OutputToken) => {
+                let denominator = FEE_SPLIT_PRECISION
+                    - <u16 as Into<u64>>::into(swap_data.referral_fee_bps)
+                    - swap_data.platform_fee_bps;
+                let numerator = params.bid_amount * FEE_SPLIT_PRECISION;
+                numerator.div_ceil(denominator)
```
anihamde, line 310: the same

auction-server/src/opportunity/entities/opportunity_svm.rs
anihamde: any reason you are not using `compute_swap_fees` from express relay metadata as in `opportunity/service/get_quotes.rs`?

anihamde (commented):
mostly minor comments

auction-server/api-types/src/opportunity.rs
m30m: This just removes it from swap opps. The structure is flattened so, they see slot in the same place they are used to see

auction-server/src/opportunity/entities/opportunity_svm.rs
m30m: I don't have access to that in here. I thought about calculating this field in the beginning and putting in all the structures but that seems dirty as well.

auction-server/src/opportunity/service/get_quote.rs
m30m: We can do it whenever there is some other logic that wants to use it. For now, let's keep it simple

sdk/js/src/svm.ts
m30m: I will create a task on taskboard for it. Whatever we do, we should do it in 4 different places, so I'd rather track it in a single place.

anihamde (approved):

PR #359, opened 2025-01-23
Add multiple tx broadcaster to svm

danimhr:
This PR aims to add multiple tx broadcaster feature to svm


auction-server/src/auction/service/auction_manager.rs
```diff
@@ -633,18 +633,17 @@ impl Service<Svm> {
...
-                tracing::error!(error = ?e, "Failed to resend transaction");
-            }
+            join_all(
+                self.config.chain_config.tx_broadcaster_clients.iter().map(|tx_broadcaster_client| async {
+                    if let Err(e) = tx_broadcaster_client
+                        .send_transaction_with_config(
+                            &bid.chain_data.transaction,
+                            self.get_send_transaction_config(),
+                        ).await {
+                        tracing::error!(error = ?e, client = ?tx_broadcaster_client.url(), "Failed to get signature status");
```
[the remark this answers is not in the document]
danimhr: Oh the error message is wrong. It's failed the submission, I'll update this

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -660,7 +659,17 @@ impl Service<Svm> {
         let res = self
             .config
             .chain_config
-            .tx_broadcaster_client
+            .tx_broadcaster_clients
+            .first()
```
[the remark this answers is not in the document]
guibescos: fyi I'm trying to remove this code in https://github.com/pyth-network/per/pull/362

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -660,7 +659,17 @@ impl Service<Svm> {
...
+            .ok_or({
+                tracing::error!("No tx broadcaster clients available");
```
guibescos, line 665: should we make it fail at initialization like Ani suggested and here just use like expect and unwrap? seems like this error should never happen

danimhr: Updated

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -674,7 +693,7 @@ impl Service<Svm> {
         metrics::histogram!("transaction_landing_time_seconds_svm", &labels)
             .record(start.elapsed().as_secs_f64());
 
-        tracing::Span::current().record("total_tries", try_count);
+        tracing::Span::current().record("total_tries", retry_count + 1);
```
guibescos, line 696: why +1 here

integration.py
```diff
@@ -45,7 +46,9 @@ def main():
   local-solana:
     express_relay_program_id: PytERJFhAKuNNuaiXkApLfWzwNwSNDACpigT3LwQfou
     rpc_read_url: http://localhost:8899
-    rpc_tx_submission_url: http://localhost:8899
+    rpc_tx_submission_urls:
+      - http://localhost:8899
+      - http://localhost:8899
```
anihamde, line 51: remarking this to remove or at least comment

anihamde (approved):

guibescos (approved):
It works, as an alternative I had the idea of making a 
`struct MultipleRpcClient (Vec<RpcClient>)` with a constructor that fails if you pass 0 urls and that can implement all things like `send_transaction`

integration.py
danimhr: Added a comment on top of the json

PR #362, opened 2025-01-23
feat: refactor send_transaction, add land time metric

guibescos:
We will record the landing time and the result (success, failed, expired)


auction-server/src/auction/service/auction_manager.rs
```diff
@@ -618,36 +623,55 @@ impl Service<Svm> {
...
-                .await
-            {
-                tracing::error!(error = ?e, "Failed to resend transaction");
             }
         }
 
-        tracing::Span::current().record("total_tries", SEND_TRANSACTION_RETRY_COUNT_SVM + 1);
+        let labels = [
+            ("chain_id", self.config.chain_id.clone()),
+            ("result", result_label.to_string()),
```
danimhr, line 669: This label is not accurate. While it's for monitoring purposes, I'm fine with it. Maybe leave a comment here that there are cases (for expired label) which the tx may land later.

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -657,12 +681,6 @@ impl Service<Svm> {
     ) -> solana_client::client_error::Result<Signature> {
         tracing::Span::current().record("bid_id", bid.id.to_string());
         let tx = &bid.chain_data.transaction;
-        let res = self
```
danimhr, line 660: I think it's important to try to submit the tx (maybe it fails to submit because of RPC issues or anything like that) and then add it to pending_transactions.

auction-server/src/auction/service/auction_manager.rs
guibescos: added comment

auction-server/src/auction/service/auction_manager.rs
guibescos: done

danimhr (approved):

PR #363, opened 2025-01-24
feat: actively query for confirmations

guibescos:
Sometimes our websocket doens't get the confirmation right away, we can lead to more retries than necessary and inaccurate landing time metrics.

We can fix this by actively pulling for confirmations at the same time as we retry.


auction-server/src/auction/service/auction_manager.rs
```diff
@@ -647,6 +647,28 @@ impl Service<Svm> {
...
                 }
                 _ = retry_interval.tick() => {
+                    match self
+                        .config
+                        .chain_config
+                        .tx_broadcaster_client
+                        .get_signature_status(&signature)
+                        .await
+                    {
+                        Ok(status) => {
```
danimhr, line 657: This part is duplicate?

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -647,6 +647,28 @@ impl Service<Svm> {
                     }
                 }
                 _ = retry_interval.tick() => {
+                    match self
```
danimhr, line 650: This needs to be merged with main as well. We have multiple tx_broadcasters now

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -647,6 +647,28 @@ impl Service<Svm> {
...
+                                if status.is_err() {
+                                    result_label = "failed";
+                                } else {
+                                    result_label = "success";
+                                }
+                                break;
+                            }
+                        }
+                        Err(e) => {
+                            tracing::error!(error = ?e, "Failed to get signature status");
```
danimhr, line 668: Add the broadcaster here after merging with main

auction-server/src/auction/service/auction_manager.rs
guibescos: added constants

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -650,10 +653,30 @@ impl Service<Svm> {
...
 
+    /// Returns Some() if the transaction has landed, None if:
+    /// - the transaction is not yet confirmed
+    /// - the rpc calls failed
+    async fn get_signature_status(
+        &self,
+        signature: &Signature,
+    ) -> Option<Result<(), TransactionError>> {
+        let result = join_all(self.config.chain_config.tx_broadcaster_clients.iter().map(
+            |tx_broadcaster_client| async {
```
danimhr, line 664: instead of async ..., can you return ```tx_broadcaster_client.get_signature_status(signature)``` here

auction-server/src/auction/service/auction_manager.rs
guibescos: logged with client failed

auction-server/src/auction/service/auction_manager.rs
guibescos: outdated with the logging I added

danimhr (approved):

PR #367, opened 2025-01-27
fix: mark as submitted only those bids that were actually submitted

guibescos:
I fix this bug where some SVM bids were being marked as submitted despite `send_transaction` failing to submit them

### DIFF
--- a/auction-server/src/auction/service/auction_manager.rs
+++ b/auction-server/src/auction/service/auction_manager.rs
@@ -417,10 +417,6 @@ impl AuctionManager<Svm> for Service<Svm> {
         _permission_key: entities::PermissionKey<Svm>,
         bids: Vec<entities::Bid<Svm>>,
     ) -> Result<Vec<Result<entities::TxHash<Svm>>>> {
-        if bids.is_empty() {
-            return Err(anyhow::anyhow!("No bids to submit"));
-        }
-
         let send_futures: Vec<_> = bids
             .into_iter()
             .map(|mut bid| {
--- a/auction-server/src/auction/service/auction_manager.rs
+++ b/auction-server/src/auction/service/auction_manager.rs
@@ -1,6 +1,7 @@
 use {
     super::{
-        simulator::MySimulator, ChainTrait, Service
+        ChainTrait,
+        Service,
     },
     crate::{
         auction::entities::{
@@ -428,12 +429,16 @@ impl AuctionManager<Svm> for Service<Svm> {
             })
             .collect();
 
-        Ok(join_all(send_futures).await.into_iter().map(|res| {
-            res.map_err(|e| {
-                tracing::error!(error = ?e, "Error while submitting bid");
-                anyhow::anyhow!(e)
+        Ok(join_all(send_futures)
+            .await
+            .into_iter()
+            .map(|res| {
+                res.map_err(|e| {
+                    tracing::error!(error = ?e, "Error while submitting bid");
+                    anyhow::anyhow!(e)
+                })
             })
-        }).collect())
+            .collect())
     }
 
     #[tracing::instrument(skip_all, fields(bid_ids, tx_hash, auction_id, bid_statuses))]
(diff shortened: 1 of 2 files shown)


auction-server/src/auction/service/auction_manager.rs
```diff
@@ -229,7 +230,7 @@ impl AuctionManager<Evm> for Service<Evm> {
             .await?
             .tx_hash();
         tracing::Span::current().record("tx_hash", format!("{:?}", tx_hash));
-        Ok(tx_hash)
+        Ok((0..len).map(|_| Ok(tx_hash)).collect())
```
m30m, line 233: let's define `len` next to here. Or you can just directly map bids.iter()

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -415,7 +416,7 @@ impl AuctionManager<Svm> for Service<Svm> {
         &self,
         _permission_key: entities::PermissionKey<Svm>,
         bids: Vec<entities::Bid<Svm>>,
-    ) -> Result<entities::TxHash<Svm>> {
+    ) -> Result<Vec<Result<entities::TxHash<Svm>>>> {
         if bids.is_empty() {
```
m30m, line 420: can we simplify this function return type to `Vec<Result<entities::TxHash<Svm>>>` and remove the len 0 check?

auction-server/src/auction/service/handle_auction.rs
```diff
@@ -72,23 +72,38 @@ where
...
-                }))
-                .await;
+            Ok(tx_hashes) => {
+                // If at least one tx is submitted successfully, we submit the auction.
+                let tx_hash = tx_hashes.iter().find(|res| res.is_ok());
+                if let Some(Ok(tx_hash)) = tx_hash {
+                    tracing::debug!("Submitted transaction: {:?}", tx_hash);
+                    let auction = self.repo.submit_auction(auction, tx_hash.clone()).await?;
+
+                    // Now we update the status for the actually submitted bids
```
m30m, line 82: I am not sure if I'm following this or not.
tx_hashes is of len(winner_bids) and you are zipping it with auction.bids which has a different size 🤔 

The logic here should update all the bids and not only the winner bids. the get_new_status checks whether the bid is one of the winner bids and returns lost or won based on that.

auction-server/src/auction/service/auction_manager.rs
guibescos: mmmm I tried this and it's hard because in the EVM version we get a single error for the whole vector and returning a vector of results would requiring cloning that error which is not allowed.

I do agree about removing the 0-length check, it seems outside of the responsibility of this function.

auction-server/src/auction/service/handle_auction.rs
guibescos: you're right, the logic was wrong

auction-server/src/auction/service/auction_manager.rs
guibescos: on second thoughts since all we do with the errors is logging...
if you want I could do Result<Vec<Option>> or Vec<Option>.
Wdyt?

auction-server/src/auction/service/handle_auction.rs
```diff
@@ -72,23 +72,43 @@ where
...
+                    let bid_status_auction = entities::BidStatusAuction {
+                        tx_hash: auction_tx_hash,
+                        id:      auction.id,
+                    };
+
+                    join_all(auction.bids.into_iter().map(|bid| {
+                        self.update_bid_status(UpdateBidStatusInput {
+                            new_status: Service::get_new_status(
+                                &bid,
+                                &submitted_bids,
```
danimhr, line 104: With this change, we are marking the bids that are not submitted (failed to be submitted) on chain as lost, and we broadcast that their bid is lost (even when it's a winner bid and we were unable to submit it). Are we sure about this UX?

auction-server/src/auction/service/auction_manager.rs
```diff
@@ -415,11 +416,7 @@ impl AuctionManager<Svm> for Service<Svm> {
         &self,
         _permission_key: entities::PermissionKey<Svm>,
         bids: Vec<entities::Bid<Svm>>,
-    ) -> Result<entities::TxHash<Svm>> {
-        if bids.is_empty() {
-            return Err(anyhow::anyhow!("No bids to submit"));
-        }
-
+    ) -> Result<Vec<Result<entities::TxHash<Svm>>>> {
```
m30m, line 419: can't we remove the outer Result?

PR #368, opened 2025-01-28
Add submit quote api

danimhr:
This PR aims to add submit quote endpoint to auction server

### DIFF
--- a/auction-server/api-types/src/quote.rs
+++ b/auction-server/api-types/src/quote.rs
@@ -1,0 +1,40 @@
+use {
+    crate::bid::BidId,
+    serde::{
+        Deserialize,
+        Serialize,
+    },
+    serde_with::{
+        serde_as,
+        DisplayFromStr,
+    },
+    solana_sdk::{
+        signature::Signature,
+        transaction::VersionedTransaction,
+    },
+    utoipa::ToSchema,
+};
+
+/// Parameters needed to submit a quote from server.
+#[serde_as]
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug)]
+pub struct SubmitQuote {
+    /// The reference id for the quote that should be submitted.
+    #[schema(example = "1", value_type = u64)]
+    pub reference_id:   BidId,
+    /// The signature of the user for the quote.
+    #[schema(example = "Jb2urXPyEh4xiBgzYvwEFe4q1iMxG1DNxWGGQg94AmKgqFTwLAiTiHrYiYxwHUB4DV8u5ahNEVtMMDm3sNSRdTg", value_type = String)]
+    #[serde_as(as = "DisplayFromStr")]
+    pub user_signature: Signature,
+}
+
+/// Response to a quote submission.
+#[serde_as]
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug)]
+pub struct SubmitQuoteResponse {
+    /// The fully signed versioned transaction that was submitted.
+    /// The transaction is encoded in base64.
+    #[schema(example = "SGVsbG8sIFdvcmxkIQ==", value_type = String)]
+    #[serde(with = "crate::serde::transaction_svm")]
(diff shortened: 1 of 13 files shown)


auction-server/src/auction/service/submit_quote.rs
```diff
@@ -0,0 +1,88 @@
...
+                    .static_account_keys()
+                    .iter()
+                    .position(|p| p.eq(&user_wallet))
+                    .expect("User wallet not found in transaction");
+                bid.chain_data.transaction.signatures[user_signature_pos] = input.user_signature;
+
+                // TODO add relayer signature after program update
+                // self.add_relayer_signature(&mut bid);
+
+                // TODO change it to a better state (Wait for user signature)
```
m30m, line 64: we probably need to check the deadline here as well

auction-server/src/opportunity/entities/quote.rs
```diff
@@ -22,6 +25,7 @@ pub struct Quote {
     pub referrer_fee:    TokenAmountSvm,
     pub platform_fee:    TokenAmountSvm,
     pub chain_id:        ChainId,
+    pub reference_id:    BidId,
```
m30m, line 28: how about `quote_id`?

danimhr: I think it's not the quote_id. Based on what we defined till now, quote_id was the OpportunityId, so I think it's better to call it sth else.

auction-server/api-types/src/quote.rs
```diff
@@ -0,0 +1,40 @@
...
+    #[schema(example = "Jb2urXPyEh4xiBgzYvwEFe4q1iMxG1DNxWGGQg94AmKgqFTwLAiTiHrYiYxwHUB4DV8u5ahNEVtMMDm3sNSRdTg", value_type = String)]
+    #[serde_as(as = "DisplayFromStr")]
+    pub user_signature: Signature,
+}
+
+/// Response to a quote submission.
+#[serde_as]
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug)]
+pub struct SubmitQuoteResponse {
+    /// The fully signed versioned transaction that was submitted.
```
m30m, line 35: nit: maybe just mention signed by user? If we add relayer signature, it's not fully signed yet

auction-server/src/auction/service/submit_quote.rs
```diff
@@ -0,0 +1,95 @@
...
+                    )
+                    .map_err(|_| RestError::BadParameters("Invalid quote".to_string()))?;
+                let SwapAccounts { user_wallet, .. } = self
+                    .extract_swap_accounts(&bid.chain_data.transaction, &swap_instruction)
+                    .await
+                    .map_err(|_| RestError::BadParameters("Invalid quote".to_string()))?;
+                let swap_args = Self::extract_swap_data(&swap_instruction)
+                    .map_err(|_| RestError::BadParameters("Invalid quote".to_string()))?;
+
+                if swap_args.deadline < OffsetDateTime::now_utc().unix_timestamp() {
```
m30m, line 47: let's check this with some buffer. Maybe 2-3 seconds?

auction-server/src/opportunity/entities/quote.rs
m30m: hmm, but do we actually have quote_id somehwere? Or are we going to introduce it?

m30m (approved):

auction-server/src/opportunity/entities/quote.rs
danimhr: We dont have it yet. But my point is that the winner_bid is is not the quote id. So the reference_id make more sense for me :-? wdyt?

auction-server/api-types/src/quote.rs
danimhr: This is in response, it should be fully signed. so the FE can also submit the tx

PR #369, opened 2025-01-28
bump: litesvm

guibescos:


m30m (approved):
can you test locally before merging to make sure the new litesvm is working as expected? Overall, the package has a lot of breaking changes.

PR #373, opened 2025-01-31
fix: improve svm landing time metric

guibescos:
- make buckets that match what we see in production better
- start measuring before the first send_transaction


auction-server/src/server.rs
```diff
@@ -135,8 +135,20 @@ async fn fetch_access_tokens(db: &PgPool) -> HashMap<models::AccessTokenToken, m
         .collect()
 }
 
+pub const TRANSACTION_LANDING_TIME_SVM_METRIC: &str = "transaction_landing_time_seconds_svm";
+const TRANSACTION_LANDING_TIME_SVM_BUCKETS: &[f64; 14] = &[
+    0.1, 0.25, 0.5, 0.75, 1.0, 1.25, 1.5, 1.75, 2.0, 2.25, 2.5, 3.75, 5.0, 10.0,
```
danimhr, line 140: This is my bucket list suggestion:
```
[0.5, 1.0, 2.0, 4.0, 6.0, 8.0, 10.0, 15.0, 20.0, 30.0, 40.0, 50.0, 60.0]
```

We need to know more about instances more than 10 seconds.

auction-server/src/server.rs
```diff
@@ -135,8 +135,20 @@ async fn fetch_access_tokens(db: &PgPool) -> HashMap<models::AccessTokenToken, m
         .collect()
 }
 
+pub const TRANSACTION_LANDING_TIME_SVM_METRIC: &str = "transaction_landing_time_seconds_svm";
```
danimhr, line 138: Please add these constants to the `per_metrics.rs` file

auction-server/src/server.rs
guibescos: @m30m wanted more granularity between 1 and 2.5 because we see currently of transactions there

auction-server/src/server.rs
guibescos: done

danimhr (approved):

PR #377, opened 2025-02-04
Add relayer signer to ER swap

danimhr:
This PR aims to add relayer signer to ER swap


contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -403,15 +403,15 @@ pub struct Swap<'info> {
         token::authority = searcher,
         token::token_program = token_program_searcher
     )]
-    pub searcher_ta_mint_searcher: InterfaceAccount<'info, TokenAccount>,
+    pub searcher_ta_mint_searcher: Box<InterfaceAccount<'info, TokenAccount>>,
```
m30m, line 406: Why do we need all the boxing?

contracts/svm/testing/tests/swap.rs
```diff
@@ -200,13 +200,15 @@ pub struct SwapSetupResult {
     pub token_user:              Token,
```
m30m, line 200: please add some tests for the new logic

contracts/svm/programs/express_relay/src/lib.rs
danimhr: While we have a lot of accounts for swap, we'll get some kind of stack overflow error. We need to box these vars, to make sure we are using heap for them to resolve the issue

danimhr: https://stackoverflow.com/questions/70747729/how-do-i-avoid-my-anchor-program-throwing-an-access-violation-in-stack-frame#:~:text=To%20solve%20this%20problem%2C%20you%20could%20try%20Boxing%20your%20account%20structs%2C%20to%20move%20them%20to%20the%20heap%3A

m30m (approved):

PR #380, opened 2025-02-05
tests: write actual tests

guibescos:
This pull requests setups a structure to test the SVM opportunity service and more broadly the auction server.

How to review this PR:
- check out `auction-server/src/opportunity/service/add_opportunity.rs` to see an example of the tests that can be written with this setup.

The opportunity service interacts with 3 external dependencies:
- the database
- the solana rpc client
- the auction service

The strategy for testing is replacing each of them with mocks whose return values we control using the crate `mockall`.  To enable mocking with mockall, I refactor the interface between the opportunity repo and the database into a trait, then I use automock to create MockOpportunityTable a mock struct that implements the trait.

Next I'd like to focus on writing tests for SVM, for both opportunity service and auction service, covering Limo and Swap.

### DIFF
--- a/auction-server/src/opportunity/repository/mod.rs
+++ b/auction-server/src/opportunity/repository/mod.rs
@@ -12,6 +12,7 @@ use {
 
 mod add_opportunity;
 mod add_spoof_info;
+mod db;
 mod get_express_relay_metadata;
 mod get_in_memory_opportunities;
 mod get_in_memory_opportunities_by_key;
--- a/auction-server/src/opportunity/repository/mod.rs
+++ b/auction-server/src/opportunity/repository/mod.rs
@@ -1,5 +1,7 @@
 use {
     super::entities,
+    crate::kernel::db::DB,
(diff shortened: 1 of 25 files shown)


auction-server/src/opportunity/repository/mod.rs
```diff
@@ -24,12 +26,17 @@ mod refresh_in_memory_opportunity;
...
+    models::*,
+};
+
 pub const OPPORTUNITY_PAGE_SIZE_CAP: usize = 100;
 
 #[derive(Debug)]
-pub struct Repository<T: InMemoryStore> {
+pub struct Repository<T: InMemoryStore, U: OpportunityTable<T> = DB> {
     pub in_memory_store: T,
+    pub db:              U,
```
guibescos, line 39: I think it's probably worth it using dynamic dispatch here to avoid the generic `U` that appears in a lot of places now

danimhr (approved):

PR #381, opened 2025-02-05
Feat/swap last look sdks

danimhr:
This PR aims to add last look functionalities to sdks


sdk/rust/testing-searcher/src/main.rs
```diff
@@ -16,8 +17,19 @@ use {
     },
 };
 
+
+#[derive(Parser, Clone, Debug)]
+pub struct RunOptions {
+    /// The API key to use for authentication.
```
m30m, line 23: The API key to use for auction server authentication.

sdk/rust/simple-searcher/src/lib.rs
```diff
@@ -328,6 +340,29 @@ impl SimpleSearcher {
                 }
                 ServerUpdateResponse::BidStatusUpdate { status } => {
                     println!("Bid status update: {:?}", status);
+                    // It's possible to cancel bids with status awaiting_signature
+                    // Doing it here randomly for demonstration purposes
+                    if let BidStatus::Svm(BidStatusSvm::AwaitingSignature { .. }) =
```
m30m, line 345: maybe we should include chain_id in these messages? 🤔 
not for this PR but thinking in general

auction-server/src/auction/service/cancel_bid.rs
```diff
@@ -5,66 +5,64 @@ use {
...
+                        auction: entities::BidStatusAuction {
+                            id: auction.id,
+                            tx_hash,
+                        },
+                    },
+                })
+                .await
+            }
+            _ => Err(RestError::BadParameters(
+                "Bid is not cancellable".to_string(),
```
m30m, line 54: maybe specify that it's only cancellable when awaitingsignature

m30m (approved):

sdk/rust/simple-searcher/src/lib.rs
danimhr: hmmm, makes sense. Lets think more about it :-?

PR #389, opened 2025-02-14
fix: remove multisig from close instructions

guibescos:


m30m (approved):

sdk/js/src/svm.ts
```diff
@@ -306,7 +306,7 @@ export function getUnwrapSolInstruction(
   owner: PublicKey,
 ): TransactionInstruction {
   const ata = getAssociatedTokenAddress(owner, NATIVE_MINT, TOKEN_PROGRAM_ID);
-  return createCloseAccountInstruction(ata, owner, owner, [owner]);
+  return createCloseAccountInstruction(ata, owner, owner);
```
danimhr, line 309: Could you please update all sdks?
This is same for python and rust

guibescos: done

PR #396, opened 2025-02-20
feat: Better fee token determination

m30m:
This is useful for our current testing to make sure searchers handle both FeeToken variants

### DIFF
--- a/auction-server/src/opportunity/service/get_quote.rs
+++ b/auction-server/src/opportunity/service/get_quote.rs
@@ -43,7 +43,10 @@ use {
     futures::future::join_all,
     solana_sdk::pubkey::Pubkey,
     spl_associated_token_account::get_associated_token_address_with_program_id,
-    std::time::Duration,
+    std::{
+        str::FromStr,
+        time::Duration,
+    },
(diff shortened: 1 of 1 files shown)


auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -126,6 +129,19 @@ pub fn get_quote_virtual_permission_account(
     Pubkey::find_program_address(&seeds, &Pubkey::default()).0
 }
 
+/// Determines if the fee token should be the user token or the searcher token
+fn get_fee_token(user_mint: Pubkey, _searcher_mint: Pubkey) -> entities::FeeToken {
```
guibescos, line 133: why not make it FeeToken::SearcherToken in UserTokenSpecified
and FeeToken::UserToken in SearcherTokenSpecified

m30m: We prefer USDC/USDT as the fee and we should select that token if it exists

guibescos: sg

PR #398, opened 2025-02-20
feat: add indicative price

anihamde:
This PR enables ER to provide indicative prices when the `user_wallet_address` is not specified. It does this in a somewhat hacky way, reusing the `get_quote` api for ease.

TODO:

- [x] update SDKs
- [x] bump versions ahead of deployment

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -719,7 +719,7 @@ dependencies = [
 
 [[package]]
 name = "auction-server"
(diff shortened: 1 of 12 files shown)


danimhr:
I think you should update the swap UI we created for test to fix the Vercel error

danimhr (approved):

PR #400, opened 2025-02-24
fix: Use rpc simulations for swap

m30m:


auction-server/src/auction/service/verification.rs
```diff
@@ -1385,9 +1412,9 @@ impl Verification<Svm> for Service<Svm> {
             .await?;
         self.verify_signatures(&bid, &bid_chain_data, &bid_data.submit_type)
             .await?;
-        // we handle simulating swap bids separately, in the get_quote function
-        if bid_payment_instruction_type == BidPaymentInstructionType::SubmitBid {
-            self.simulate_bid(&bid).await?;
+        match bid_payment_instruction_type {
+            BidPaymentInstructionType::Swap => self.simulate_swap_bid(&bid).await?,
+            _ => self.simulate_bid(&bid).await?,
```
guibescos, line 1417: maybe be explicit with the variant here instead of the default branch

auction-server/src/auction/service/simulator.rs
```diff
@@ -320,10 +324,7 @@ impl Simulator {
     }
 
     #[tracing::instrument(skip_all)]
-    pub fn setup_lite_svm(
-        &self,
-        accounts_config_with_context: &Response<AccountsConfig>,
-    ) -> LiteSVM {
+    fn setup_lite_svm(&self, accounts_config_with_context: &Response<AccountsConfig>) -> LiteSVM {
```
guibescos, line 327: I don't understand why this got reformatted

guibescos (approved):
lgtm

auction-server/src/auction/service/simulator.rs
m30m: I removed `pub`

PR #401, opened 2025-02-25
Add better conclusion loop

danimhr:
This PR aims to update the conclusion loop to make sure we are not running a single loop at the same time for all of the pending auctions.


auction-server/src/auction/repository/add_auction.rs
```diff
@@ -24,6 +24,7 @@ impl<T: ChainTrait> Repository<T> {
             .insert(auction.id, auction);
     }
 
+    // NOTE: Do not call this function directly. Instead call `add_auction` from `Service`.
```
m30m, line 27: this seems dirty, anyway we can limit the access programatically?

m30m (approved):

PR #403, opened 2025-02-25
feat: Add compute limit to sample searchers

m30m:
Changed how js simple searcher was constructing the limo bids because the submitBid instruction was coming on top.
We should probably deprecate that function later since it doesn't make a lot of sense to add the PER instruction on top of the tx.

### DIFF
--- a/sdk/js/src/examples/simpleSearcherSvm.ts
+++ b/sdk/js/src/examples/simpleSearcherSvm.ts
@@ -106,6 +106,8 @@ export class SimpleSearcherSvm {
   }
 
   generateComputeBudgetInstructions(): TransactionInstruction[] {
+    // This limit assumes no other custom instructions exist in the transaction, you may need to adjust
+    // this limit depending on your integration
     const computeLimitInstruction = ComputeBudgetProgram.setComputeUnitLimit({
       units: 300000,
     });
--- a/sdk/js/src/examples/simpleSearcherSvm.ts
+++ b/sdk/js/src/examples/simpleSearcherSvm.ts
@@ -33,7 +33,7 @@ import {
   getPdaAuthority,
   OrderStateAndAddress,
 } from "@kamino-finance/limo-sdk/dist/utils";
-import { constructSwapBid } from "../svm";
+import { constructSubmitBidInstruction, constructSwapBid } from "../svm";
 
 const DAY_IN_SECONDS = 60 * 60 * 24;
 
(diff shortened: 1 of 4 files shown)


sdk/js/src/examples/simpleSearcherSvm.ts
```diff
@@ -105,6 +105,17 @@ export class SimpleSearcherSvm {
     return decimals;
   }
 
+  generateComputeBudgetInstructions(): TransactionInstruction[] {
+    const computeLimitInstruction = ComputeBudgetProgram.setComputeUnitLimit({
+      units: 300000,
```
anihamde, line 110: maybe a comment explaining this, otherwise searchers may blindly use this value without measuring the CUs their txs take up

sdk/js/src/examples/simpleSearcherSvm.ts
```diff
@@ -119,20 +130,12 @@ export class SimpleSearcherSvm {
...
-    });
-    const txRaw = new anchor.web3.Transaction().add(
-      feeInstruction,
-      ...ixsTakeOrder,
-    );
 
     const bidAmount = await this.getBidAmount(opportunity);
 
     const config = await this.getExpressRelayConfig();
-    const bid = await this.client.constructSvmBid(
```
anihamde, line 134: why not edit the `constructSvmBid` method?

sdk/python/express_relay/searcher/examples/simple_searcher_svm.py
```diff
@@ -150,6 +150,13 @@ async def generate_bid(self, opp: OpportunitySvm) -> BidSvm:
         if opp.program == "swap":
             return await self.generate_bid_swap(opp)
 
+    def generate_compute_budget_instructions(self) -> List[Instruction]:
+        compute_limit_instruction = set_compute_unit_limit(300_000)
```
anihamde, line 154: the same as above

anihamde (approved):

sdk/js/src/examples/simpleSearcherSvm.ts
```diff
@@ -142,12 +145,20 @@ export class SimpleSearcherSvm {
       config.relayerSigner,
       config.feeReceiverRelayer,
     );
-    bid.slot = opportunity.slot;
-
-    bid.transaction.recentBlockhash =
-      this.latestChainUpdate[this.chainId].blockhash;
-    bid.transaction.sign(this.searcher);
-    return bid;
+    const tx = new anchor.web3.Transaction().add(
```
guibescos, line 148: fwiw i don't think the order of the compute unit instructions within the tx matter so the refactor was unnecessary

sdk/js/src/examples/simpleSearcherSvm.ts
guibescos: for swap this value should almost always be the same and we can set it to be a reasonable value

anihamde: this is used for Limo though

sdk/js/src/examples/simpleSearcherSvm.ts
m30m: the refactor was just to bring both of the computebudget instructions on top which seems like a good practice

PR #404, opened 2025-02-26
Add temporary fix for multiple get quote requests at the same time

danimhr:
This PR fix the issue of the `Temporary Unavailable` server response occurring in specific cases when multiple simultaneous requests are made to fetch a quote with same body at the exact same time.

### DIFF
--- a/auction-server/src/opportunity/service/get_quote.rs
+++ b/auction-server/src/opportunity/service/get_quote.rs
@@ -489,7 +489,7 @@ impl Service<ChainTypeSvm> {
         {
             // This can only happen if the bid is already updated by another auction for another get_quote request
             // TODO We should handle this case more gracefully
-            return Err(RestError::QuoteNotFound);
+            return Err(RestError::DuplicateOpportunity);
         }
 
         let metadata = self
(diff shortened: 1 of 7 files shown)


auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -446,47 +443,54 @@ impl Service<ChainTypeSvm> {
...
-                }
-            },
-            None => {
-                tracing::error!(auction = ?auction, opportunity = ?opportunity, winner_bid = ?winner_bid, "Failed to get auction by id");
-                return Err(RestError::TemporarilyUnavailable);
-            }
-        };
+            // This can only happen if the bid is already updated by another auction for another get_quote request
+            // TODO We should handle this case more gracefully
+            return Err(RestError::QuoteNotFound);
```
anihamde, line 492: maybe a more verbose error? maybe it's not a user concern, but could be handled more specifically and be more informative if it was `DuplicateQuoteRequest`?

anihamde (approved):

auction-server/src/opportunity/service/get_quote.rs
danimhr: Updated

PR #405, opened 2025-02-26
Fix/server restart opp removal

anihamde:
This PR fixes a bug whereby server restarts would delete the in memory store of opportunities but not from the db, leaving a bunch of opportunities with null removal reason or time. This PR updates the removal time and removal reason for those opportunities in the db at the time of server restart.


m30m:
We probably need some db indexes if we want to run the query efficiently.

PR #408, opened 2025-03-02
Feat/test verify bid

danimhr:
This PR aims to add a test for `verify_bid` function for swap opportunity with mocking every external resource for auction module including opportunity service. The main objective here is mocking the resources not a complete test suit. Later we will add more tests for the function.

### DIFF
--- a/auction-server/src/api.rs
+++ b/auction-server/src/api.rs
@@ -87,7 +87,7 @@ async fn root() -> String {
 pub mod profile;
 pub(crate) mod ws;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum RestError {
     /// The request contained invalid parameters.
     BadParameters(String),
@@ -115,6 +115,8 @@ pub enum RestError {
     QuoteNotFound,
     /// Duplicate opportunity.
     DuplicateOpportunity,
+    /// Swao opportunity is not found.
+    SwapOpportunityNotFound,
 }
 
 impl RestError {
@@ -164,6 +166,10 @@ impl RestError {
             RestError::DuplicateOpportunity => (
                 StatusCode::BAD_REQUEST,
                 "Same opportunity is submitted recently".to_string(),
+            ),
+            RestError::SwapOpportunityNotFound => (
+                StatusCode::BAD_REQUEST,
+                "No swap opportunity with the given id found".to_string(),
             ),
         }
     }
--- a/auction-server/src/api.rs
+++ b/auction-server/src/api.rs
@@ -115,7 +115,7 @@ pub enum RestError {
     QuoteNotFound,
     /// Duplicate opportunity.
     DuplicateOpportunity,
-    /// Swao opportunity is not found.
+    /// Swap opportunity is not found.
     SwapOpportunityNotFound,
 }
 
--- a/auction-server/src/auction/repository/mod.rs
+++ b/auction-server/src/auction/repository/mod.rs
@@ -3,10 +3,7 @@ use {
         entities,
         service::ChainTrait,
     },
-    crate::kernel::{
-        db::DB,
-        entities::ChainId,
-    },
+    crate::kernel::entities::ChainId,
     axum_prometheus::metrics,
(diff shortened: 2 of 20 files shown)


auction-server/src/api.rs
```diff
@@ -115,6 +115,8 @@ pub enum RestError {
     QuoteNotFound,
     /// Duplicate opportunity.
     DuplicateOpportunity,
+    /// Swao opportunity is not found.
```
guibescos, line 118:
```suggestion
    /// Swap opportunity is not found.
```

auction-server/src/auction/repository/mod.rs
```diff
@@ -90,15 +87,15 @@ impl<T: ChainTrait> Default for InMemoryStore<T> {
 #[derive(Debug)]
 pub struct Repository<T: ChainTrait> {
     pub in_memory_store: InMemoryStore<T>,
-    pub db:              DB,
+    pub db:              Box<dyn models::DBTrait<T>>,
```
guibescos, line 90: very nice

auction-server/src/auction/repository/models.rs
```diff
@@ -599,3 +613,200 @@ impl<T: ChainTrait + ModelTrait<T>> Bid<T> {
...
+    ) -> anyhow::Result<entities::Auction<T>>;
+    async fn update_bid_status(
+        &self,
+        bid: &entities::Bid<T>,
+        new_status: &T::BidStatusType,
+    ) -> anyhow::Result<bool>;
+}
+
+#[async_trait]
+impl<T: ChainTrait> DBTrait<T> for DB {
```
guibescos, line 650: Something you could do here is divide this trait into two: `AuctionTable` and `BidTable`. 
It also makes it more consistent with the trait for opprotunities which is called `OpportunityTable`

auction-server/src/auction/service/mod.rs
```diff
@@ -1,3 +1,7 @@
+#[cfg(test)]
+use crate::opportunity::service::tests::MockService as OpportunityService;
+#[cfg(not(test))]
+use crate::opportunity::service::Service as OpportunityService;
```
guibescos, line 4: did you try updating this to use the #[double] macro?

auction-server/src/api.rs
```diff
@@ -165,6 +167,10 @@ impl RestError {
                 StatusCode::BAD_REQUEST,
                 "Same opportunity is submitted recently".to_string(),
             ),
+            RestError::SwapOpportunityNotFound => (
+                StatusCode::BAD_REQUEST,
```
guibescos, line 171: 404?

auction-server/src/auction/repository/models.rs
danimhr: I created a single trait because I’m not certain whether all the functions will use queries on a single table. For example, some queries might involve joins across multiple tables. Therefore, I thought it would be better to use a more general name.

auction-server/src/auction/service/mod.rs
danimhr: In that case we need to mock the service outside of the tests mod

auction-server/src/api.rs
danimhr: I dont want to change the api structure for now

auction-server/src/auction/service/mod.rs
guibescos: is that a problem?

auction-server/src/auction/repository/models.rs
guibescos: I somehow don't like the name because:
- too generic
- I think it's weird to have the word trait in the name of a trait
maybe something like `AuctionDB` ?

danimhr: AuctionDB does not include the bid. That's why I prefer the generic name. What about `Database`?

auction-server/src/auction/service/mod.rs
danimhr: not a problem :D 
Let me test it

PR #409, opened 2025-03-03
Feat: ws workers

m30m:
Use separate threads to process some of the time consuming requests in a single ws connection unblocking the rest. This allows higher concurrency per ws connection.

### DIFF
--- a/auction-server/src/api/ws.rs
+++ b/auction-server/src/api/ws.rs
@@ -80,12 +80,9 @@ use {
         time::Duration,
     },
     time::OffsetDateTime,
-    tokio::{
-        spawn,
-        sync::{
-            broadcast,
-            Semaphore,
-        },
+    tokio::sync::{
+        broadcast,
+        Semaphore,
     },
     tracing::{
         instrument,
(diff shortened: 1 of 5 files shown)


auction-server/src/api/ws.rs
```diff
@@ -193,6 +213,21 @@ impl Subscriber {
                     maybe_message_or_err.ok_or(anyhow!("Client channel is closed"))??
                 ).await
             },
+            response_received = self.response_receiver.recv() => {
+                match response_received{
+                    Ok(DeferredResponse{response, bid_id_to_add}) =>{
```
danimhr, line 218: Could you please reformatted spacing?

auction-server/src/api/ws.rs
```diff
@@ -444,11 +537,17 @@ impl Subscriber {
             }
         };
 
-        let response = match maybe_client_message {
-            Err(e) => Err(ServerResultResponse {
-                id:     None,
-                result: ServerResultMessage::Err(e.to_string()),
-            }),
+        match maybe_client_message {
```
danimhr, line 540: is it possible to call the `spawn_deferred` here instead of inside the functions?

m30m: not all the arms use that, so it becomes a bit awkward

PR #410, opened 2025-03-03
chore: make opportunity db a trait object

guibescos:
Make the notations cleaner for opportunity repo

### DIFF
--- a/auction-server/src/opportunity/repository/db.rs
+++ b/auction-server/src/opportunity/repository/db.rs
@@ -19,7 +19,9 @@ use {
         },
         opportunity::entities::Opportunity,
     },
+    axum::async_trait,
     sqlx::QueryBuilder,
+    std::fmt::Debug,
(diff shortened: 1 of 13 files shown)


auction-server/src/opportunity/repository/db.rs
```diff
@@ -31,7 +33,8 @@ use {
 };
 
 #[cfg_attr(test, automock)]
-pub trait OpportunityTable<T: InMemoryStore> {
+#[async_trait]
+pub trait OpportunityTable<T: InMemoryStore>: Debug + Send + Sync + 'static {
```
guibescos, line 37: tbh I have no idea why the trait bounds seem to fix it

danimhr: We can check the errors and see why each of them are needed :D

danimhr (approved):

PR #411, opened 2025-03-03
feat: merge test and release svm program

guibescos:


.github/workflows/test-and-release-svm-contract.yml
```diff
@@ -1,30 +1,41 @@
...
       - name: Install Solana Verify CLI
         run: |
           cargo install solana-verify --git https://github.com/Ellipsis-Labs/solana-verifiable-build --rev 121568e
       - name: Build
-        working-directory: ./contracts/svm
         run: solana-verify build
+      - name: Run tests
+        env:
+          SBF_OUT_DIR: ${{ github.workspace }}/contracts/svm/target/deploy
+        run: cargo test --target-dir ./target-tests
```
guibescos, line 32: we use `target-tests/` because `target/` becomes readonly after `solana-verify build` for some reasons I don't fully understand... but it's convenient because it means the artifact doesn't get updated at this step

guibescos: also avoid `cargo-test-sbf` because it rebuilds the binary

.github/workflows/test-and-release-svm-contract.yml
```diff
@@ -1,30 +1,41 @@
...
+    paths:
+      - contracts/svm/**
   push:
     branches:
       - main
     paths:
       - contracts/svm/**
 
+env:
+  CARGO_TERM_COLOR: always
```
anihamde, line 14: what does this do out of curiosity?

anihamde (approved):

.github/workflows/test-and-release-svm-contract.yml
guibescos: I don't know but it was in test-svm-contract.yml. I just copied it over 😄

PR #418, opened 2025-03-06
feat: return a quote when not enough funds

guibescos:
- in auction service, we let through transactions that fail with the user insufficient funds error
- in opportunity service, we check if the user has enough funds and set tx in the response to none if they don't

### DIFF
--- a/auction-server/src/auction/service/verification.rs
+++ b/auction-server/src/auction/service/verification.rs
@@ -1258,7 +1258,7 @@ impl Service<Svm> {
                     {
                         if instruction_error
                             == InstructionError::Custom(ErrorCode::InsufficientUserFunds.into())
-                        {   
+                        {
                             // This path only works as long as none of the other accepted programs use this error number, which is currently true
                             // TODO: Also check the instruction index here
                             return Ok(());
--- a/auction-server/src/auction/service/verification.rs
+++ b/auction-server/src/auction/service/verification.rs
@@ -1258,7 +1258,9 @@ impl Service<Svm> {
                     {
                         if instruction_error
                             == InstructionError::Custom(ErrorCode::InsufficientUserFunds.into())
-                        {
+                        {   
+                            // This path only works as long as none of the other accepted programs use this error number, which is currently true
+                            // TODO: Also check the instruction index here
                             return Ok(());
                         }
                     }
--- a/auction-server/src/auction/service/verification.rs
+++ b/auction-server/src/auction/service/verification.rs
@@ -1259,8 +1259,8 @@ impl Service<Svm> {
                         if instruction_error
                             == InstructionError::Custom(ErrorCode::InsufficientUserFunds.into())
                         {
-                            // This path only works as long as none of the other accepted programs use this error number, which is currently true
-                            // TODO: Also check the instruction index here
+                            // This path only works as long as none of the other accepted programs use this error number, which is currently true since express relay is the only program in the whitelist that uses anchor error codes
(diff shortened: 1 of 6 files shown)


auction-server/src/auction/service/verification.rs
```diff
@@ -1245,7 +1252,19 @@ impl Service<Svm> {
...
-                if simulation.value.err.is_some() {
+                if let Some(transaction_error) = simulation.value.err {
+                    if let TransactionError::InstructionError(_, instruction_error) =
+                        transaction_error
+                    {
+                        if instruction_error
+                            == InstructionError::Custom(ErrorCode::InsufficientUserFunds.into())
+                        {
+                            // This path only works as long as none of the other accepted programs use this error number, which is currently true since express relay is the only program in the whitelist that uses anchor error codes
+                            // TODO: Also check the instruction index here so we are sure it's coming from the swap instruction
```
anihamde, line 1263: wouldn't this TODO resolve the above comment? since ER doesn't make any CPIs to anything except system & token program?

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -526,12 +527,26 @@ impl Service<ChainTypeSvm> {
             ),
         };
 
-        let (transaction, expiration_time) = match input.quote_create.user_wallet_address {
-            None => (None, None),
-            Some(_) => (
+        let user_has_enough_balance =
+            if let Some(user_wallet_address) = input.quote_create.user_wallet_address {
+                self.check_user_token_balance(CheckUserTokenBalanceInput {
```
anihamde, line 532: are we checking here to prevent sybiling? why do we need to make an RPC call every time a quote is requested? couldn't we just avoid this check and use the error code itself?

anihamde: it seems we could handle this in verification.rs and just return a third option to allow the bid to remain with a flag that this failed due to `UserInsufficientFunds`. then we don't have to recheck the token balance on chain?

m30m: I think the changes we need for having more context **before sending the request to searchers** directly affect this change. Probably we can put a lot of info in the opportunity which we can use here. Let's chat about it.

guibescos: yeah my reasoning for doing it this way is that we will end up having these rpc calls in `get_quote` to get the existing ATAs info

auction-server/src/auction/service/verification.rs
guibescos: YES! let me clarify it

auction-server/src/opportunity/service/get_quote.rs
0xy0k: A few thoughts:
- Avoiding unnecessary RPC calls: Instead of making an RPC call every time get_quote is requested, better to cache or pre-fetch the required ATA balance info beforehand and reference it when needed?
- Handling in verification.rs: Agree that flagging bids with UserInsufficientFunds within verification.rs could reduce redundant checks. This would help streamline the process while preserving necessary validations.
- Context from opportunity struct: If additional context is already available before sending requests to searchers, leveraging this info to determine bid validity before an RPC call sounds like a great optimization.
- Tradeoff to consider: While keeping the RPC calls in get_quote ensures real-time accuracy, it introduces overhead. Could a hybrid approach be used where fetch balances once per session or on relevant state changes instead of per-quote?

PR #420, opened 2025-03-11
Add owner check to close account instruction

danimhr:
This PR aims to add owner check for close account instruction of swap bids


guibescos (approved):
lgtm

PR #421, opened 2025-03-11
Feat/wrap unwrap sol searcher

danimhr:
This PR aims to enable the searcher to wrap or unwrap SOL if one of the swap tokens is WSOL.

### DIFF
--- a/auction-server/src/auction/service/verification.rs
+++ b/auction-server/src/auction/service/verification.rs
@@ -2697,6 +2697,33 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_verify_bid_when_no_transfer_instruction() {
+        let (service, opportunities) = get_service(false);
+        let searcher = Keypair::new();
+        let swap_instruction = svm::Svm::get_swap_instruction(GetSwapInstructionParams {
+            searcher:             searcher.pubkey(),
+            opportunity_params:   get_opportunity_params(opportunities[2].clone()),
+            bid_amount:           1,
+            deadline:             (OffsetDateTime::now_utc() + Duration::minutes(1))
+                .unix_timestamp(),
+            fee_receiver_relayer: Pubkey::new_unique(),
+            relayer_signer:       service.config.chain_config.express_relay.relayer.pubkey(),
+        })
+        .unwrap();
+        let result = get_verify_bid_result(
+            service,
+            searcher,
+            vec![swap_instruction],
+            opportunities[2].clone(),
+        )
+        .await;
+        assert_eq!(
+            result.unwrap_err(),
+            RestError::InvalidInstruction(None, InstructionError::InvalidTransferInstructionsCount)
+        );
+    }
+
+    #[tokio::test]
     async fn test_verify_bid_when_no_transfer_instruction_is_allowed() {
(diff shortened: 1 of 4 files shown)


auction-server/src/auction/service/verification.rs
```diff
@@ -776,80 +798,137 @@ impl Service<Svm> {
...
+                        from: *tx
+                            .message
+                            .static_account_keys()
+                            .get(instruction.accounts[0] as usize)
+                            .ok_or(RestError::BadParameters(
+                                "Invalid account in sol transfer instruction".to_string(),
+                            ))?,
+                        to: *tx
+                            .message
+                            .static_account_keys()
```
guibescos, line 825: are we disallowing account lookup tables? what if to is in the ALT

auction-server/src/auction/service/verification.rs
```diff
@@ -913,85 +991,125 @@ impl Service<Svm> {
...
-            get_associated_token_address(&swap_accounts.user_wallet, &spl_token::native_mint::id());
-        if *account_to_close != ata {
-            return Err(RestError::InvalidInstruction(
-                None,
-                InstructionError::InvalidAccountToCloseCloseAccountInstruction {
-                    expected: ata,
-                    found:    *account_to_close,
-                },
-            ));
+        // User have to unwrap Sol
```
guibescos, line 1045:
```suggestion
        // User has to unwrap Sol
```

auction-server/src/auction/service/verification.rs
guibescos: if so maybe we should add a check that explicitly throws for this

auction-server/src/auction/service/verification.rs
```diff
@@ -2595,6 +2696,33 @@ mod tests {
...
+                .unix_timestamp(),
+            fee_receiver_relayer: Pubkey::new_unique(),
+            relayer_signer:       service.config.chain_config.express_relay.relayer.pubkey(),
+        })
+        .unwrap();
+        let result = get_verify_bid_result(
+            service,
+            searcher,
+            vec![swap_instruction],
+            opportunities[2].clone(),
```
guibescos, line 2717:
```suggestion
            opportunities[2].clone(), // User token wsol
```

auction-server/src/auction/service/verification.rs
```diff
@@ -3600,4 +3728,347 @@ mod tests {
...
+            &get_associated_token_address(&searcher.pubkey(), &spl_token::native_mint::id()),
+            &searcher.pubkey(),
+            &searcher.pubkey(),
+            &[],
+        )
+        .unwrap();
+        let mut transaction = Transaction::new_with_payer(
+            &[
+                swap_instruction,
+                transfer_instruction,
```
guibescos, line 3776: this order doesn't make sense in real life

guibescos (approved):

auction-server/src/auction/service/verification.rs
danimhr: Yes you are right. We decided to do the lookup integration later. So we have to make sure we do it for user and searcher accounts.

PR #422, opened 2025-03-11
Fix: simulator executable accounts

m30m:
There was no need to distinguish between upgradable_programs and normal programs as long as we set the program data before setting the program account so I simplified that logic.


auction-server/src/auction/service/simulator.rs
```diff
@@ -284,36 +279,25 @@ impl Simulator {
...
-            .map(|(_, programdata_address)| *programdata_address)
-            .collect::<Vec<_>>();
-        let indirect_accounts = self
-            .get_multiple_accounts_with_cache(&indirect_keys)
+        let program_datas = self
+            .get_multiple_accounts_with_cache(&program_data_addresses)
             .await?;
-        for ((program_key, _), program_account) in
-            programs_to_fetch.iter().zip(indirect_accounts.iter())
+        for (programdata_address, program_data_account) in
```
guibescos, line 294: nit: `programdata_address` -> `program_data_address` for consistency

guibescos (approved):
nice!

PR #423, opened 2025-03-11
feat: fetch and give to searcher extra information about atas and user balance

guibescos:


auction-server/api-types/src/opportunity.rs
```diff
@@ -337,6 +341,10 @@ pub enum OpportunityParamsV1ProgramSvm {
         /// Details about the tokens to be swapped. Either the searcher token amount or the user token amount must be specified.
         #[schema(inline)]
         tokens: QuoteTokensWithTokenPrograms,
+
+        /// Details about which token accounts need to be initialized an by whom
```
anihamde, line 345: `and`

auction-server/api-types/src/opportunity.rs
```diff
@@ -347,6 +355,24 @@ pub enum FeeToken {
...
+pub enum TokenAccountInitializationConfig {
+    Initialized,
+    SearcherPayer,
+    UserPayer,
+}
+
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug, ToResponse)]
+pub struct TokenAccountInitializationConfigs {
+    pub user_ata_mint_searcher:         TokenAccountInitializationConfig,
+    pub user_ata_mint_user:             Option<TokenAccountInitializationConfig>,
```
anihamde, line 369: why is this an option?

auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -61,15 +61,67 @@ impl PartialEq<ProgramFeeToken> for FeeToken {
...
+                    Self::SearcherPayer
+                }
+            }
+        }
+    }
+}
+
+
+impl Default for TokenAccountInitializationConfig {
+    fn default() -> Self {
```
anihamde, line 88: why is this the default? should there even be a default, or should we always be explicit about the variant we use?

auction-server/api-types/src/opportunity.rs
```diff
@@ -312,6 +312,10 @@ pub enum OpportunityParamsV1ProgramSvm {
         #[serde_as(as = "DisplayFromStr")]
         user_wallet_address: Pubkey,
 
+        /// The user's current balance of the user token
+        #[schema(example = 10)]
+        user_mint_user_balance: u64,
```
anihamde, line 317: why does this matter? to prevent searcher from requesting too much in cases of searcher token specified?

auction-server/api-types/src/opportunity.rs
guibescos: It's because this ata only needs to get initialized when mint_user is WSOL

auction-server/api-types/src/opportunity.rs
guibescos: the reasoning is since we have this info, we might as well make it available to the searchers

auction-server/src/opportunity/entities/opportunity_svm.rs
guibescos: you're totally right. I made this for tests but it shouldn't be a `Default`

auction-server/api-types/src/opportunity.rs
```diff
@@ -347,6 +355,25 @@ pub enum FeeToken {
...
+    SearcherPayer,
+    UserPayer,
+}
+
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug, ToResponse)]
+pub struct TokenAccountInitializationConfigs {
+    pub user_ata_mint_searcher:         TokenAccountInitializationConfig,
+    /// This will be None unless the mint user is wrapped SOL, the only case in which the user ata for mint user needs to be initialized during the swap
+    pub user_ata_mint_user:             Option<TokenAccountInitializationConfig>,
+    pub router_fee_receiver_ta:         TokenAccountInitializationConfig,
```
m30m, line 371: why `_ta` here but `_ata` in other places?

auction-server/api-types/src/opportunity.rs
m30m: Maybe we add another variant to `TokenAccountInitializationConfig` that represents this instead of wrapping it inside an option. Something like `NotNeeded`

auction-server/api-types/src/opportunity.rs
```diff
@@ -347,6 +355,25 @@ pub enum FeeToken {
...
+#[serde(rename_all = "snake_case")]
+pub enum TokenAccountInitializationConfig {
+    Initialized,
+    SearcherPayer,
+    UserPayer,
+}
+
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug, ToResponse)]
+pub struct TokenAccountInitializationConfigs {
+    pub user_ata_mint_searcher:         TokenAccountInitializationConfig,
```
m30m, line 368: Naming of these fields is a bit confusing. Can you add docstrings above them for more explanations?

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -277,6 +279,25 @@ impl Service<ChainTypeSvm> {
             }],
         };
 
+        let balances = self
+            .get_quote_request_account_balances(QuoteRequestAccountBalancesInput {
+                user_wallet_address,
+                mint_searcher: searcher_mint,
```
m30m, line 285: can't we make these names consistent?

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
```diff
@@ -0,0 +1,182 @@
...
+            .as_ref()
+            .map(|account| account.lamports)
+            .unwrap_or_default();
+
+        let token_balances: Vec<Option<u64>> = accounts[1..].iter()
+            .map(|account| {
+                account
+                    .as_ref()
+                    .map(|acc| {
+                        TokenAccount::unpack(acc.data.as_slice())
```
m30m, line 162: does this also work with Token2022? If yes, maybe mention it explicitly

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
```diff
@@ -0,0 +1,182 @@
...
+        } else {
+            self.user_ata_mint_user.unwrap_or_default()
+        }
+    }
+
+    pub fn get_token_account_initialization_configs(
+        &self,
+        mint_user_is_wrapped_sol: bool,
+    ) -> TokenAccountInitializationConfigs {
+        let rent = Rent::default();
```
m30m, line 62: Maybe add a TODO that this doesn't reflect the actual Rent values on-chain and should be fixed. But of course, not an important TODO

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub mint_user:              Pubkey,
+    pub token_program_searcher: Pubkey,
+    pub token_program_user:     Pubkey,
+}
+
+/// The balances of some of the accounts that will be used in the swap
+/// - If a token account doesn't exist, the balance is `None`
+/// - If a token account exists but has no balance, the balance is `Some(0)`
+pub struct QuoteRequestAccountBalances {
+    pub user_wallet:                    u64,
```
m30m, line 40: maybe rename to `user_sol_balance` or `user_lamports` ?

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    pub fn get_token_account_initialization_configs(
+        &self,
+        mint_user_is_wrapped_sol: bool,
+    ) -> TokenAccountInitializationConfigs {
+        let rent = Rent::default();
+
+        let user_payer = self.user_wallet.saturating_sub(rent.minimum_balance(0))
```
m30m, line 64: simplify and mention that this is just a heuristic

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub user_wallet:                    u64,
+    pub user_ata_mint_searcher:         Option<u64>,
+    pub user_ata_mint_user:             Option<u64>,
+    pub router_fee_receiver_ta:         Option<u64>,
+    pub relayer_fee_receiver_ata:       Option<u64>,
+    pub express_relay_fee_receiver_ata: Option<u64>,
+}
+
+impl QuoteRequestAccountBalances {
+    pub fn get_user_ata_mint_user_balance(&self, mint_user_is_wrapped_sol: bool) -> u64 {
```
m30m, line 49: I don't understand why we make this distinction. If the user balances is splitted between wrapped sol and sol, it's not easy to leverage both to make the trade.

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub user_wallet_address:    Pubkey,
+    pub router:                 Pubkey,
+    pub mint_searcher:          Pubkey,
+    pub mint_user:              Pubkey,
+    pub token_program_searcher: Pubkey,
+    pub token_program_user:     Pubkey,
+}
+
+/// The balances of some of the accounts that will be used in the swap
+/// - If a token account doesn't exist, the balance is `None`
```
m30m, line 37: if it's not too much work I think it's better to replace this Option<u64> to

enum TokenBalance {
  Uninitialized,
  Initialized(u64)
}

Or at least declare it as a type.

auction-server/api-types/src/opportunity.rs
```diff
@@ -337,6 +341,10 @@ pub enum OpportunityParamsV1ProgramSvm {
         /// Details about the tokens to be swapped. Either the searcher token amount or the user token amount must be specified.
         #[schema(inline)]
         tokens: QuoteTokensWithTokenPrograms,
+
+        /// Details about which token accounts need to be initialized and by whom
+        #[schema(inline)]
```
m30m, line 346: inlining would pollute this level of hierarchy too much. I'd rather keep them nested in a separate field

auction-server/api-types/src/opportunity.rs
guibescos: it's the same naming as in the program.
The program technically doesn't enforce that `router_fee_receiver_ta` is an ata

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
guibescos: I think the right call here is to return the `user_sol_balance` when `user_mint` is wsol since that's the balance the searchers will try to trade against.

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
```diff
@@ -0,0 +1,208 @@
...
+            .as_ref()
+            .map(|account| account.lamports)
+            .unwrap_or_default();
+
+        let token_balances: Vec<Option<u64>> = accounts[1..].iter()
+            .map(|account| {
+                account
+                    .as_ref()
+                    .map(|acc| {
+                        TokenAccount::unpack(acc.data.as_slice()) // This will also work for token2022 accounts since the first TokenAccount::LEN bytes have the same layout
```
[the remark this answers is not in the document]
guibescos: lol you're right

PR #425, opened 2025-03-12
chore: bump to anchor 0.31.0

guibescos:
I think we should do this for dependency hygiene.
Until now we had 2 differents versions of solana in our crate, this solves that!
Also removes many crates patches that we had.


Cargo.toml
```diff
@@ -39,33 +39,8 @@ async-stream = "0.3.5"
...
-# They later updated the crate to loosen the dependencies
-# in the commit we are using for this patch.
-# We will be able to remove this patch once the only version of solana that we use is 2 or later.
-[patch.crates-io.aes-gcm-siv]
-git = "https://github.com/RustCrypto/AEADs"
-rev = "e1e35e0c4f4943da0a99ceb8477c421dcfae2c33"
-
-# This patch disables debugging features in litesvm runtime_environments
-# which allows more programs to be loaded in the runtime
-[patch.crates-io.litesvm]
```
m30m, line 69: ser, you forgot the patch

guibescos: sir does litesvm 0.6.0 also require the patch?

.github/workflows/ci-pre-commit.yml
```diff
@@ -59,7 +59,10 @@ jobs:
         working-directory: contracts/evm
         run: forge install nomad-xyz/ExcessivelySafeCall@be417ab0c26233578b8d8f3a37b87bd1fcb4e286 --no-git --no-commit
       - name: Install Anchor CLI
-        run: npm install -g @coral-xyz/anchor-cli@v0.30.1
+        run: |
+          cargo install --git https://github.com/coral-xyz/anchor avm
+          avm install latest
+          avm use latest
```
guibescos, line 65: the previous way doesn't currently work 
https://github.com/coral-xyz/anchor/issues/3596

contracts/svm/Cargo.toml
```diff
@@ -19,6 +19,6 @@ codegen-units = 1
 anchor-lang = ">=0.29.0"
 anchor-client = ">=0.29.0"
 anchor-spl = ">=0.29.0"
-litesvm = "0.1.0"
-solana-program-test = "1.17.3"
-solana-sdk = "1.17.3"
+litesvm = ">=0.1.0"
+solana-program-test = ">=1.17.3"
+solana-sdk = ">=1.17.3"
```
guibescos, line 24: leaving these loose so people with anchor 0.29.0 can still use the express-relay crate if they want

m30m (approved):
very nice, thanks!

m30m:
I think we can simplify `auction/service/mod.rs:343` as well

PR #427, opened 2025-03-14
Feat: better determination for fee tokens

anihamde:
This PR adds a way to specify the hierarchy of fee tokens.

### DIFF
--- a/auction-server/src/opportunity/service/get_quote.rs
+++ b/auction-server/src/opportunity/service/get_quote.rs
@@ -21,7 +21,6 @@ use {
                 Service as AuctionService,
             },
         },
-        config::FeeTokenScore,
         kernel::entities::PermissionKeySvm,
         opportunity::{
             api::INDICATIVE_PRICE_TAKER,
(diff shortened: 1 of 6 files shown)


auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -129,13 +126,23 @@ pub fn get_quote_virtual_permission_account(
...
+/// If neither token tier has been specified, the fee token is set to the user token to simplify the logic for searchers.
+fn get_fee_token(
+    user_mint: Pubkey,
+    searcher_mint: Pubkey,
+    ordered_fee_tokens: &[Pubkey],
+) -> entities::FeeToken {
+    let user_token_tier = ordered_fee_tokens
+        .iter()
+        .position(|&x| x == user_mint)
+        .unwrap_or(u64::MAX as usize);
```
m30m, line 140: why not `usize::MAX`?

integration.py
```diff
@@ -53,6 +55,10 @@ def main():
     accepted_token_programs:
       - TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA
       - TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb
+    ordered_fee_tokens:
```
m30m, line 58: I'm not sure if it's a good idea to fix these in tilt too. For testing purposes, I would rather see all combinations and it can become annoying that we can't test with Sol as the fee token.

m30m (approved):

PR #428, opened 2025-03-14
feat: update sdks to use the atas information

guibescos:
Not sure about bumping yet because we might want to ship some more changes before the next release


sdk/js/src/svm.ts
```diff
@@ -275,13 +279,115 @@ function extractSwapInfo(swapOpportunity: OpportunitySvmSwap): {
...
+};
+
+type TokenAccountInitializationParams = {
+  owner: PublicKey;
+  mint: PublicKey;
+  program: PublicKey;
+  config: TokenAccountInitializationConfig;
+};
+
+function getTokenAccountToCreate(
```
m30m, line 300: maybe something like this to simplify?
```ts
function getTokenAccountToCreate(
  searcher: PublicKey,
  user: PublicKey,
  params: TokenAccountInitializationParams
): TokenAccountToCreate | undefined {
  if (params.config === "unneeded") return undefined;

  return {
    payer: params.config === "searcher_payer" ? searcher : user,
    owner: params.owner,
    mint: params.mint,
    program: params.program,
  };
}
```

sdk/js/src/svm.ts
```diff
@@ -328,33 +436,38 @@ export async function constructSwapBid(
...
-  if (swapOpportunity.referralFeeBps > 0) {
-    tokenAccountsToCreate.push({
-      owner: router,
-      mint: mintFee,
-      program: feeTokenProgram,
-    });
-  }
+
+  const tokenAccountsToCreate = getTokenAccountsToCreate(
+    {
```
m30m, line 443: I would rather pass in the `swapOpportunity` and call `extractSwapInfo` inside. Lot of parameters here and also in the function signature.

sdk/js/src/types.ts
```diff
@@ -141,6 +141,18 @@ export type SvmSwapTokens = (
   searcherToken: PublicKey;
   userToken: PublicKey;
 };
+
+export type TokenAccountInitializationConfig =
+  components["schemas"]["TokenAccountInitializationConfig"];
+
+export type TokenAccountInitializationConfigs = {
```
m30m, line 148: only if you like a more compressed version:
```ts
export type TokenAccountInitializationConfigs = Record<
  | "expressRelayFeeReceiverAta"
  | "relayerFeeReceiverAta"
  | "routerFeeReceiverAta"
  | "userAtaMintSearcher"
  | "userAtaMintUser",
  TokenAccountInitializationConfig
>;
```

sdk/python/express_relay/client.py
```diff
@@ -562,6 +581,87 @@ def get_svm_submit_bid_instruction(
         )
         return submit_bid_ix
 
+    @staticmethod
+    def get_token_account_to_create(
+        searcher: Pubkey,
+        user: Pubkey,
+        params: TokenAccountInitializationParams,
+    ) -> TokenAccountToCreate | None:
+        if params["config"] == "searcher_payer":
```
m30m, line 590: same simplification can happen here

sdk/rust/src/svm.rs
```diff
@@ -79,6 +84,39 @@ pub struct GetSubmitBidInstructionParams {
...
+
+pub struct TokenAccountToCreate {
+    pub payer:   Pubkey,
+    pub owner:   Pubkey,
+    pub mint:    Pubkey,
+    pub program: Pubkey,
+}
+
+pub struct TokenAccountInitializationParams {
+    pub config:  TokenAccountInitializationConfig,
```
m30m, line 114: nit: can we move config to the end so that field orders are the same across SDKs ?:D

m30m (approved):
Added a bunch of styling comments which are applicable on all the 3 sdks

PR #432, opened 2025-03-17
fix: wrap the right amount of sol

guibescos:
There was a bug where we were wrapping the wrong amount of SOL when `searcher_specfied`, `fee_token == "user"` and the `user_token` is wrapped SOL


sdk/rust/src/svm.rs
```diff
@@ -118,12 +118,12 @@ pub struct TokenAccountInitializationParams {
...
 pub struct GetSwapInstructionParams {
-    pub searcher:             Pubkey,
-    pub opportunity_params:   OpportunityParamsSvm,
-    pub bid_amount:           u64,
-    pub deadline:             i64,
-    pub fee_receiver_relayer: Pubkey,
-    pub relayer_signer:       Pubkey,
+    pub searcher:                  Pubkey,
+    pub opportunity_params:        OpportunityParamsSvm,
+    pub bid_amount_including_fees: u64,
```
guibescos, line 123: I changed the name because it's a breaking change and I want this to break people's code

sdk/js/src/svm.ts
```diff
@@ -137,7 +137,7 @@ export function createAtaIdempotentInstruction(
 export async function constructSwapInstruction(
   searcher: PublicKey,
   swapOpportunity: OpportunitySvmSwap,
-  bidAmount: anchor.BN,
+  bidAmountIncludingFees: anchor.BN,
```
m30m, line 140: this does not break the api, it just changes the behaviour since we are not passing them as objects. So I suggest changing it back to bidAmount and call the `getBidAmountIncludingFees` function internally.

sdk/rust/src/svm.rs
m30m: I suggest changing both back to bid_amount to remain consistent and calculate the other field internally.

sdk/js/src/svm.ts
```diff
@@ -161,34 +161,22 @@ export async function constructSwapInstruction(
...
-        swapOpportunity.platformFeeBps + swapOpportunity.referralFeeBps,
-      ),
-    );
-    const numerator = bidAmount.mul(FEE_SPLIT_PRECISION);
-    // add denominator - 1 to round up
-    bidAmount = numerator
-      .add(denominator.sub(new anchor.BN(1)))
-      .div(denominator);
-  }
+  const bidAmountIncludingFees = getBidAmountIncludingFees(
```
guibescos, line 164: I renamed the variable because I don't like mutating the argument `bidAmount` like we did previously

m30m (approved):

PR #435, opened 2025-03-18
Add rate limiter for total number of subscriber per ip

danimhr:
This PR aims to add rate limiter to total number of active subscribers per ip

### DIFF
--- a/auction-server/src/api/ws.rs
+++ b/auction-server/src/api/ws.rs
@@ -1,6 +1,7 @@
 use {
     super::{
         Auth,
+        RestError,
         WrappedRouter,
     },
     crate::{
@@ -32,6 +33,7 @@ use {
             State,
             WebSocketUpgrade,
         },
+        http::HeaderMap,
         response::IntoResponse,
         Router,
     },
@@ -68,8 +70,12 @@ use {
         StreamExt,
     },
     std::{
-        collections::HashSet,
+        collections::{
+            HashMap,
+            HashSet,
+        },
         future::Future,
+        net::IpAddr,
         sync::{
             atomic::{
                 AtomicUsize,
@@ -82,6 +88,7 @@ use {
     time::OffsetDateTime,
     tokio::sync::{
         broadcast,
+        RwLock,
         Semaphore,
     },
     tracing::{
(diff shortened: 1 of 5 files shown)


auction-server/src/api/ws.rs
```diff
@@ -91,26 +98,72 @@ use {
...
+    let ws_state = &store.store.ws;
+    let requester_ip = headers
+        .get(ws_state.requester_ip_header_name.as_str())
+        .and_then(|value| value.to_str().ok())
+        .and_then(|value| value.split(',').next()) // Only take the first ip if there are multiple
+        .and_then(|value| value.parse().ok());
+
+    match requester_ip {
+        Some(ip) => {
+            if let Some(ids) = ws_state.subscriber_per_ip.read().await.get(&ip) {
```
m30m, line 125: why not increment it here instead of websocket_handler?

auction-server/src/api/ws.rs
```diff
@@ -91,26 +98,72 @@ use {
 };
 
 pub struct WsState {
-    pub subscriber_counter: AtomicUsize,
-    pub broadcast_sender:   broadcast::Sender<UpdateEvent>,
-    pub broadcast_receiver: broadcast::Receiver<UpdateEvent>,
+    pub subscriber_per_ip:        RwLock<HashMap<IpAddr, HashSet<SubscriberId>>>,
```
m30m, line 101: why do we need the subscriber ids? why not just a u64?

auction-server/src/api/ws.rs
danimhr: imagine that I concurrently send reqs and try to create 100 ws connections at the same time

auction-server/src/api/ws.rs
danimhr: Yeah a counter is enough, just thought maybe it become useful later

auction-server/src/api/ws.rs
```diff
@@ -91,26 +98,106 @@ use {
...
+                let ids = write_gaurd.entry(ip).or_insert_with(HashSet::new);
+                if ids.len() >= MAXIMUM_SUBSCRIBERS_PER_IP {
+                    None
+                } else {
+                    let id = self.subscriber_counter.fetch_add(1, Ordering::SeqCst);
+                    ids.insert(id);
+                    Some(id)
+                }
+            }
+            None => Some(self.subscriber_counter.fetch_add(1, Ordering::SeqCst)),
```
m30m, line 137: this fetch_add logic can come out

PR #439, opened 2025-03-18
feat: reply when not enough funds

guibescos:


auction-server/src/auction/service/verification.rs
```diff
@@ -315,11 +322,12 @@ impl Verification<Evm> for Service<Evm> {
...
-    pub router:             Pubkey,
-    pub permission_account: Pubkey,
-    pub deadline:           OffsetDateTime,
-    pub submit_type:        SubmitType,
+    pub amount:                          u64,
+    pub router:                          Pubkey,
+    pub permission_account:              Pubkey,
+    pub deadline:                        OffsetDateTime,
+    pub submit_type:                     SubmitType,
+    pub express_relay_instruction_index: usize,
```
guibescos, line 330: This feels a bit weird here but I prefer this than a tuple as the return value of `extract_bid_data`

auction-server/src/opportunity/api.rs
```diff
@@ -228,10 +227,7 @@ pub async fn post_quote(
 
     let quote = store
         .opportunity_service_svm
-        .get_quote(GetQuoteInput {
-            quote_create,
-            program: ProgramSvm::Swap,
```
guibescos, line 233: the program in GetQuoteInput was always swap so I remove it

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
```diff
@@ -89,21 +89,16 @@ impl QuoteRequestAccountBalances {
...
 
         TokenAccountInitializationConfigs {
-            user_ata_mint_user:             if mint_user_is_wrapped_sol {
-                self.user_ata_mint_user.get_initialization_config(false)
-            } else {
-                TokenAccountInitializationConfig::Unneeded
-            },
+            user_ata_mint_user:             self
+                .user_ata_mint_user
+                .get_initialization_config(false),
```
guibescos, line 101: we need to ask searchers to initialize this account if it doesn't exist, 
otherwise the simulations will get rejected with account not initialized instead of `UserInsufficientFunds`

auction-server/src/auction/service/verification.rs
m30m: maybe just `instruction_index`?

auction-server/src/auction/service/verification.rs
```diff
@@ -471,18 +479,19 @@ impl Service<Svm> {
...
             })
-            .cloned()
-            .collect::<Vec<CompiledInstruction>>();
+            .map(|(index, instruction)| (index, instruction.clone()))
+            .collect::<Vec<(usize, CompiledInstruction)>>();
 
-        let instruction = match instructions.len() {
-            1 => Ok(instructions[0].clone()),
+        let (instruction_index, instruction) = match instructions.len() {
+            1 => Ok(instructions.into_iter().next().unwrap()),
```
m30m, line 494: nit: replace unwrap with expect

auction-server/src/auction/service/verification.rs
```diff
@@ -1374,7 +1391,18 @@ impl Service<Svm> {
             .await;
         match simulation {
             Ok(simulation) => {
-                if simulation.value.err.is_some() {
+                if let Some(transaction_error) = simulation.value.err {
+                    if let TransactionError::InstructionError(index, error) = transaction_error {
```
m30m, line 1395: I would put this boolean logic inside a variable called `isInsufficientUserFunds` for more readability.

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -70,11 +72,6 @@ use {
 /// Time to wait for searchers to submit bids.
 const BID_COLLECTION_TIME: Duration = Duration::from_millis(500);
 
-pub struct GetQuoteInput {
```
m30m, line 73: I think we wanted to encapsulate each service input in one of these structs. For the sake of consistency, we can keep it as is and just remove the program field

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
m30m: why aren't we passing `user_payer` as the argument here?

auction-server/src/auction/service/verification.rs
guibescos: `swap_instruction_index`

guibescos: mmm, not sure... there are multiple instruction in a bid (seen as a tx)
this is the instruction index in the bid where they call the express relay program

auction-server/src/opportunity/service/get_quote_request_account_balances.rs
guibescos: let's look at this in a future PR, there are some edgecases we need to think about

auction-server/src/auction/service/verification.rs
m30m: yeah since this is in a struct called bid data I thought it's clear

m30m (approved):

PR #446, opened 2025-03-21
feat: ata creation verification

guibescos:
We focus on preventing user to pay for any atas except those that are marked with `UserPayer`

### DIFF
--- a/auction-server/src/auction/service/verification.rs
+++ b/auction-server/src/auction/service/verification.rs
@@ -1345,7 +1345,9 @@ impl Service<Svm> {
             &token_account_initialization_configs.user_ata_mint_searcher,
         )?;
 
-        // we don't care about the other ata creation instructions but searcher must pay
+        // we rely on the simulation to check the other token accounts are created
+        // but we enforce here that the searcher pays for their creation
+        // this includes searcher token accounts and fee token accounts
         for account in create_ata_instructions {
             if account.payer != swap_accounts.searcher {
                 return Err(RestError::InvalidInstruction(
--- a/auction-server/src/auction/service/verification.rs
+++ b/auction-server/src/auction/service/verification.rs
@@ -3,8 +3,7 @@ use {
         auction_manager::TOTAL_BIDS_PER_AUCTION_EVM,
         ChainTrait,
         Service,
-    },
-    crate::{
+    }, crate::{
         api::{
             InstructionError,
             RestError,
@@ -38,10 +37,7 @@ use {
         opportunity::{
             self as opportunity,
             entities::{
-                get_swap_quote_tokens,
-                OpportunitySvm,
-                OpportunitySvmProgram::Swap,
-                QuoteTokens,
+                get_swap_quote_tokens, OpportunitySvm, OpportunitySvmProgram::Swap, QuoteTokens, TokenAccountInitializationConfig, TokenAccountInitializationConfigs
             },
             service::{
                 get_live_opportunities::GetLiveOpportunitiesInput,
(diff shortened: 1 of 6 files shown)


auction-server/src/auction/service/verification.rs
```diff
@@ -1061,6 +1089,78 @@ impl Service<Svm> {
...
+                )
+            ) {
+                if instruction.accounts.len() < 6 {
+                    return Err(RestError::BadParameters(
+                        "Invalid create ata instruction accounts".to_string(),
+                    ));
+                }
+                let invalid_account_message =
+                    "Invalid account in create ata instruction".to_string();
+                let payer = tx
```
m30m, line 1113: this looks very repetitive. Maybe we can simplify it with some inline function?
or just something like let (payer, ata, owner,... ) =

auction-server/src/auction/service/verification.rs
```diff
@@ -1168,8 +1268,118 @@ impl Service<Svm> {
...
+            &swap_accounts.token_program_user,
+            &token_account_initialization_configs.user_ata_mint_user,
+        )?;
+        validate_and_remove_create_ata_instruction(
+            &swap_accounts.mint_searcher,
+            &swap_accounts.token_program_searcher,
+            &token_account_initialization_configs.user_ata_mint_searcher,
+        )?;
+
+        // we don't care about the other ata creation instructions but searcher must pay
```
m30m, line 1366: why don't we care about fee receiver atas?

auction-server/src/auction/service/verification.rs
```diff
@@ -1168,8 +1268,118 @@ impl Service<Svm> {
...
+                return Err(RestError::BadParameters(format!(
+                    "Opportunity with id {} is not a swap opportunity",
+                    opp.id
+                )));
+            }
+        };
+
+        let mut create_ata_instructions = Self::extract_create_ata_instructions(tx)?;
+
+        let mut validate_and_remove_create_ata_instruction =
```
m30m, line 1288: I would signal that this only works for `user_wallet` atas

auction-server/src/auction/service/verification.rs
```diff
@@ -1168,8 +1268,118 @@ impl Service<Svm> {
     }
 
 
+    fn check_create_ata_instructions(
+        tx: &VersionedTransaction,
+        swap_accounts: &SwapAccounts,
+        opp: &OpportunitySvm,
+    ) -> Result<(), RestError> {
+        let token_account_initialization_configs = match &opp.program {
```
m30m, line 1276: there are a few places in the whole verification flow where we are doing something similar. another match like this and another one inside `get_swap_quote_tokens`. Ideally, we can do this once and pass it around? But it's ok if too much work.

auction-server/src/auction/service/verification.rs
guibescos: this part we get updated in https://github.com/pyth-network/per/pull/452 so let's address this then

auction-server/src/auction/service/verification.rs
guibescos: we rely on simulation for these

auction-server/src/auction/service/verification.rs
guibescos: refactored

auction-server/src/auction/service/verification.rs
guibescos: renamed

auction-server/src/auction/service/verification.rs
guibescos: `relayer_fee_receiver_ata` or `express_relay_fee_receiver_ata` are not accessed anywhere in the verification flow so we already rely in simulation to check those values

m30m: pls just add a comment describing this assumption

PR #451, opened 2025-03-25
program updates

m30m:
Changes:
- Add a secondary relayer for 0-downtime transitions between relayers
- Relax ata constraints on fee receivers: Only check their existence if the fees are positive
- Add a new instruction where swap fee bps is determined by the input

Tests to add:
- Assert second relayer works
- Test custom fees
- Test not checking atas when fees are 0

Next steps:
- Deploy program
- Accepting both instructions on the server side
- Update SDKs to use the new instruction
- Disallow the older swap instruction
- Configurable fees

### DIFF
--- a/contracts/svm/programs/express_relay/src/swap.rs
+++ b/contracts/svm/programs/express_relay/src/swap.rs
@@ -2,10 +2,7 @@ use {
     crate::{
         error::ErrorCode,
         state::ExpressRelayMetadata,
-        token::{
-            check_receiver_and_transfer_token_if_needed,
-            transfer_token_if_needed,
-        },
+        token::check_receiver_and_transfer_token_if_needed,
         FeeToken,
         Swap,
         SwapArgs,
@@ -14,6 +11,7 @@ use {
     },
     anchor_lang::{
         accounts::interface_account::InterfaceAccount,
+        error::ErrorCode as AnchorErrorCode,
         prelude::*,
     },
     anchor_spl::token_interface::TokenAccount,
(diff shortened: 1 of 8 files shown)


contracts/svm/programs/express_relay/src/swap.rs
```diff
@@ -128,9 +157,27 @@ pub struct SwapFees {
     pub express_relay_fee: u64,
 }
 impl ExpressRelayMetadata {
+    pub fn check_relayer_signer(&self, relayer_signer: &Pubkey) -> Result<()> {
+        if !self.relayer_signer.eq(relayer_signer)
+            && !self.secondary_relayer_signer.eq(relayer_signer)
```
guibescos, line 162: did you try putting the constraint in the context?

contracts/svm/programs/express_relay/src/token.rs
```diff
@@ -9,19 +9,38 @@ use {
     },
 };
 
+pub fn get_token_account_checked(
+    account: &AccountInfo,
+    expected_program_id: &Pubkey,
+) -> Result<TokenAccount> {
+    if account.data_len() == 0 {
+        return Err(ErrorCode::AccountNotInitialized.into());
```
guibescos, line 17: this check is not needed because `try_deserialize` will fail (i understand if you want to keep it)

contracts/svm/programs/express_relay/src/token.rs
```diff
@@ -9,19 +9,38 @@ use {
...
+    }
+
+    if account.owner != expected_program_id {
+        return Err(ErrorCode::ConstraintOwner.into());
+    }
+
+    let token_account = match TokenAccount::try_deserialize(&mut &account.data.borrow()[..]) {
+        Ok(ta) => ta,
+        Err(_) => return Err(ErrorCode::ConstraintTokenTokenProgram.into()),
```
guibescos, line 26: wrong error here

contracts/svm/programs/express_relay/src/token.rs
```diff
@@ -9,19 +9,38 @@ use {
...
+pub fn get_token_account_checked(
+    account: &AccountInfo,
+    expected_program_id: &Pubkey,
+) -> Result<TokenAccount> {
+    if account.data_len() == 0 {
+        return Err(ErrorCode::AccountNotInitialized.into());
+    }
+
+    if account.owner != expected_program_id {
+        return Err(ErrorCode::ConstraintOwner.into());
```
guibescos, line 21:
```suggestion
        return Err(ErrorCode::ConstraintTokenTokenProgram.into());
```

contracts/svm/programs/express_relay/src/token.rs
guibescos, line 26:
```suggestion
        Err(_) => return Err(ErrorCode::AccountDidNotDeserialize.into()),
```

guibescos (approved):

contracts/svm/programs/express_relay/src/swap.rs
m30m: not really, I think this is a better place since it's just not a single context where we need this

PR #461, opened 2025-03-28
fix:  wsol edgecase 

guibescos:


auction-server/src/api.rs
```diff
@@ -208,42 +208,42 @@ impl std::fmt::Display for InstructionError {
...
-                    found, expected
+                    expected, found
                 )
             }
             InstructionError::InvalidTokenProgramInCreateAtaInstruction { expected, found } => {
                 write!(
                     f,
                     "Invalid token program in create associated token account instruction. Expected: {:?} found: {:?}",
-                    found, expected
+                    expected, found
```
guibescos, line 246: drive-by

auction-server/src/auction/service/verification.rs
```diff
@@ -845,6 +846,11 @@ impl Service<Svm> {
 
         // User have to wrap Sol
         if swap_accounts.mint_user == spl_token::native_mint::id() {
+            // Sometimes the user doesn't have enough SOL, but we want the transaction to fail in the Expert Relay program with InsufficientUserFunds
```
anihamde, line 849: the relay truly is an expert 😄

auction-server/src/auction/service/verification.rs
```diff
@@ -2216,7 +2228,7 @@ mod tests {
                 token_program_searcher: spl_token::id(),
                 fee_token,
                 referral_fee_bps,
-                user_mint_user_balance: 0,
```
anihamde, line 2219: how were all these passing previously?

sdk/js/src/svm.ts
```diff
@@ -250,6 +251,7 @@ function extractSwapInfo(swapOpportunity: OpportunitySvmSwap): {
   tokenProgramSearcher: PublicKey;
   tokenInitializationConfigs: TokenAccountInitializationConfigs;
   memo?: string;
+  userMintUserBalance: anchor.BN;
```
anihamde, line 254: why doesn't the server just provide the amount to wrap?

anihamde (approved):

auction-server/src/auction/service/verification.rs
guibescos: the `user_mint_user_balance` wasn't used for anything

sdk/js/src/svm.ts
guibescos: we can't do this since it can depend on the bid amount (when the trade is searcher_specified)

sdk/js/src/svm.ts
```diff
@@ -39,6 +39,7 @@ function getExpressRelayProgram(chain: string): PublicKey {
 }
 
 export const FEE_SPLIT_PRECISION = new anchor.BN(10000);
+const RENT_TOKEN_ACCOUNT_LAMPORTS = 2039280;
```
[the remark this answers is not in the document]
guibescos: no, the rent methods require an rpc call in js and python

PR #463, opened 2025-03-31
fix: 0 length accounts

guibescos:
atas can sometimes be uninitialized but exist (when someone has airdropped some sol to the address for example)


auction-server/src/opportunity/service/get_quote_request_account_balances.rs
```diff
@@ -196,7 +196,10 @@ impl Service<ChainTypeSvm> {
                 account
                     .as_ref()
                     .map(|acc| {
-                        StateWithExtensions::<TokenAccount>::unpack(&acc.data[..TokenAccount::LEN])
+                        if acc.data.is_empty() {
+                            return Ok(0);
```
m30m, line 200: can you add some comments? I guess this is just uninitialized account right?

PR #465, opened 2025-04-02
feat: cancellable flag 2

guibescos:


auction-server/src/auction/service/submit_quote.rs
```diff
@@ -59,39 +67,60 @@ impl Service<Svm> {
...
+            self.send_transaction(&bid).await;
+        }
         Ok(())
     }
 
+    pub async fn sign_bid_and_submit_auction(
+        &self,
+        bid: entities::Bid<Svm>,
+        auction: entities::Auction<Svm>,
+        submit_bid: bool,
```
danimhr, line 113: if we decide to always `submit_bid` when anyone calling this function with `user_signature` then we may not need this field anymore

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -604,6 +611,16 @@ impl Service<ChainTypeSvm> {
             (None, None)
         };
 
+        let transaction = match (&transaction, input.quote_create.cancellable) {
+            (Some(_), false) => Some(
+                auction_service
+                    .sign_bid_and_submit_auction(winner_bid.clone(), auction.clone(), false)
```
danimhr, line 617: I would suggest to only add the relayer signature here to the winner bid?
It'll make things simpler specifically on the `submit_bid` function side.

danimhr:
Please take a look at the `conclude_auction_loop` function as well. I think we need to change sth on the SVM `get_bid_results` line 465:
```
        let statuses = if bids.iter().any(|bid| bid.status.is_submitted()) {
```
We may want to send an on-chain query to fetch the status of the bids with status `SentToUserForSubmission`.

auction-server/src/auction/repository/submit_auction.rs
```diff
@@ -16,8 +16,12 @@ impl<T: ChainTrait> Repository<T> {
         tracing::Span::current().record("auction_id", auction.id.to_string());
         tracing::Span::current().record("tx_hash", format!("{:?}", transaction_hash));
 
-        let auction = self.db.submit_auction(&auction, &transaction_hash).await?;
-        self.update_in_memory_auction(auction.clone()).await;
-        Ok(auction)
+        if auction.submission_time.is_none() {
```
danimhr, line 19: To make sure no race condition can happen here, the DB level should return True/False (Or raise error) if update failed. Based on that we can decide to update the memory and what to return.

PR #466, opened 2025-04-02
feat: Add counter metrics to track quote submission

Dewyer:
# About

These changes should add extra metrics to submit quote about the kind of result that was received during the execution.

# Sidenote
I made changes to the tiltfile that should add linux compatibility to the test searchers, while still working on mac os as well.


danimhr:
@Dewyer We may close this PR?

PR #467, opened 2025-04-03
Update submit quote error messages

danimhr:
This PR aims to update the error messages in the submit quote


m30m (approved):
Maybe we can give something more specfic than 400 in some of these cases. Not all of these are bad requests :D

PR #469, opened 2025-04-03
fix: forward compatibility

guibescos:


sdk/python/express_relay/models/svm.py
```diff
@@ -380,6 +381,13 @@ class BidStatusSvm(BaseModel):
     type: BidStatusVariantsSvm
     result: SvmSignature | None = Field(default=None)
 
+    @field_validator("type", mode="before")
+    @classmethod
+    def check_type(cls, data):
```
anihamde, line 386: maybe just a comment for clarity that this is for forward compatibility and not bc an SDK should expect arbitrary bid status variants?

anihamde (approved):

PR #472, opened 2025-04-04
feat: Add ppm precision to swap v2 fees

Dewyer:
# About
In this PR i increased the possible precision for the dynamic platform and referrer fees.
This required a bit of trickery hence the swap v2 account still used the swap v1 argument instruction data, so that the two implementations could share a context. Which has to broken here because the size of the referer fee ppm needs to be increased to u64 from u16 breaking wire compatibility. This introduces some duplication. The hack were the constraints are moved out of the account validator don't work because the instruction data still needs to deserialize properly even if we werent using fee token data.

### DIFF
--- a/contracts/svm/programs/express_relay/src/lib.rs
+++ b/contracts/svm/programs/express_relay/src/lib.rs
@@ -455,11 +455,11 @@ pub struct SwapV2Args {
     pub deadline:              i64,
     pub amount_searcher:       u64,
     pub amount_user:           u64,
-    /// The referral fee is specified in mill percentile
+    /// The referral fee is specified in parts per million
     pub referral_fee_ppm:      u64,
     /// Token in which the fees will be paid
     pub fee_token:             FeeToken,
-    /// The platform fee is specified in mill percentile
+    /// The platform fee is specified in parts per million
     pub swap_platform_fee_ppm: u64,
 }
 
@@ -524,12 +524,12 @@ pub struct Swap<'info> {
     pub mint_searcher: Box<InterfaceAccount<'info, Mint>>,
 
     #[account(mint::token_program = token_program_user)]
-    /// CHECK: we check that this is set correctly based on the fee token manually since the V2 arguments are incompatible on the wire
     pub mint_user: Box<InterfaceAccount<'info, Mint>>,
 
     #[account(
         mint::token_program = token_program_fee,
     )]
+    /// CHECK: we check that this is set correctly based on the fee token manually since the V2 arguments are incompatible on the wire
     pub mint_fee: Box<InterfaceAccount<'info, Mint>>,
 
     // Token programs
--- a/contracts/svm/programs/express_relay/src/lib.rs
+++ b/contracts/svm/programs/express_relay/src/lib.rs
@@ -182,8 +182,8 @@ pub mod express_relay {
         )
     }
 
-
     pub fn swap_internal(ctx: Context<Swap>, data: SwapV2Args) -> Result<()> {
+        ctx.accounts.check_raw_constraints(data.fee_token)?;
         check_deadline(data.deadline)?;
         ctx.accounts
             .express_relay_metadata
@@ -227,6 +227,7 @@ pub mod express_relay {
 
         Ok(())
     }
+
     pub fn swap(ctx: Context<Swap>, data: SwapArgs) -> Result<()> {
         let data = ctx.accounts.convert_to_v2(&data);
         swap_internal(ctx, data)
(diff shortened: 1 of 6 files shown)


contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -546,3 +502,228 @@ pub struct Swap<'info> {
 
     pub relayer_signer: Signer<'info>,
 }
+
+#[derive(Accounts)]
+#[instruction(data: Box<SwapV2Args>)]
```
guibescos, line 507: this macro is only used to define some constraints, so the old way will magically work if the offset of fee_token is the same in `SwapV2Args` and `SwapArgs`

contracts/svm/programs/express_relay/src/state.rs
```diff
@@ -2,6 +2,9 @@ use anchor_lang::prelude::*;
 
 pub const FEE_SPLIT_PRECISION: u64 = 10_000;
 
+pub const FEE_SPLIT_PRECISION_PPM: u64 = 100_000_000;
```
guibescos, line 5: isn't this supposed to be `1_000_000`

contracts/svm/programs/express_relay/src/state.rs
```diff
@@ -2,6 +2,9 @@ use anchor_lang::prelude::*;
...
+pub const FEE_BPS_TO_PPM: u64 = 10_000;
```
guibescos, line 6: and this 100

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -438,31 +439,31 @@ pub struct SwapArgs {
...
 pub struct SwapV2Args {
-    // deadline as a unix timestamp in seconds
+    /// deadline as a unix timestamp in seconds
     pub deadline:              i64,
     pub amount_searcher:       u64,
     pub amount_user:           u64,
-    // The referral fee is specified in basis points
-    pub referral_fee_bps:      u16,
-    // Token in which the fees will be paid
+    /// The referral fee is specified in mill percentile
```
guibescos, line 458: what is a `mill percentile`

contracts/svm/programs/express_relay/src/lib.rs
```diff
@@ -523,21 +524,19 @@ pub struct Swap<'info> {
     pub mint_searcher: Box<InterfaceAccount<'info, Mint>>,
 
     #[account(mint::token_program = token_program_user)]
+    /// CHECK: we check that this is set correctly based on the fee token manually since the V2 arguments are incompatible on the wire
```
guibescos, line 527: this comment should go above `mint_fee`

contracts/svm/testing/tests/swap.rs
```diff
@@ -23,6 +24,7 @@ use {
         },
     },
     testing::{
+        assert_all_token_balances,
```
guibescos, line 27: can you replace the other appearances of `token_balance_matches` with the macro, if not too hard

guibescos (commented):
maybe you could add a test for wrong fee token program

contracts/svm/programs/express_relay/src/lib.rs
Dewyer: 1 millionth percentile, i thought thats what we wanted to raise it to, but i can divi down the numbers as well

contracts/svm/testing/tests/swap.rs
```diff
@@ -1913,12 +1838,12 @@ fn test_swap_v2_mint_searcher_ppm_fee() {
...
+                user.pubkey() => 9961.89,
+                get_express_relay_metadata_key() => 17.7760,
+                express_relay_metadata.fee_receiver_relayer => 4.4440,
             },
             raw: {
-                router_ta_mint_searcher => 0.158900,
+                router_ta_mint_searcher => 15.8900,
             }
         }
     );
```
guibescos, line 1849: maybe you could add a test for wrong fee token program

contracts/svm/testing/tests/swap.rs
```diff
@@ -846,6 +846,58 @@ fn test_swap_wrong_mint_fee() {
...
+        &mut svm,
+        &instructions,
+        &searcher,
+        &[&searcher, &user, &relayer_signer],
+    )
+    .unwrap_err();
+    assert_custom_error(
+        result.err,
+        4,
+        InstructionError::Custom(AnchorErrorCode::ConstraintTokenMint.into()),
```
guibescos, line 897: wait i thought we were supposed to get `ConstraintRaw`

Dewyer: true

PR #473, opened 2025-04-04
fix: do arithmetic in u128

guibescos:


auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -357,12 +357,17 @@ impl From<OpportunitySvm> for api::OpportunitySvm {
...
-                                user_token.amount - router_fee - platform_fee
+                                let router_fee = u128::from(user_token.amount)
+                                    * u128::from(program.referral_fee_bps)
+                                    / u128::from(FEE_SPLIT_PRECISION);
+                                let platform_fee = u128::from(user_token.amount)
+                                    * u128::from(program.platform_fee_bps)
+                                    / u128::from(FEE_SPLIT_PRECISION);
+                                user_token
+                                    .amount
+                                    .saturating_sub(router_fee as u64)
```
anihamde, line 368: maybe just do the casts in the lines where the variables are defined? all the logic of the type conversions is contained in that single line there, which improves readability imo

anihamde: can also include the comment about the safety of the casts in each of those lines for clarity

anihamde (approved):

PR #474, opened 2025-04-07
Add dynamic deadline to quote

danimhr:
This PR introduces a `minimum_lifetime` parameter for `get_quote` requests. It allows the Protocol to optimize pricing by requesting quotes with varying deadline windows, either shorter or longer, based on its strategy.

### DIFF
--- a/auction-server/src/opportunity/entities/opportunity_svm.rs
+++ b/auction-server/src/opportunity/entities/opportunity_svm.rs
@@ -10,7 +10,10 @@ use {
         OpportunityCreate,
     },
     crate::{
-        auction::entities::BidPaymentInstructionType,
+        auction::{
+            entities::BidPaymentInstructionType,
+            service::verification::DEFAULT_SWAP_BID_MINIMUM_LIFE_TIME,
+        },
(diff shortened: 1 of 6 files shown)


auction-server/src/opportunity/entities/opportunity_svm.rs
```diff
@@ -104,6 +107,7 @@ pub struct OpportunitySvmProgramSwap {
     pub token_program_searcher:               Pubkey,
     pub token_account_initialization_configs: TokenAccountInitializationConfigs,
     pub memo:                                 Option<String>,
+    pub minimum_lifetime:                     Option<u32>,
```
m30m, line 110: some of these types including this one can just be `u32`.  from searcher's perspective it's just a new param that they may use to adjust the their bids.

danimhr: I think this is inside the entity. At the entity level, it's an optional variable and can be None. We replace it with default value when wanna send it to searchers

m30m (approved):

PR #476, opened 2025-04-08
Fix tracing error level in instrument

danimhr:
This PR aims to fix the issue with logging the errors of the traces

### DIFF
--- a/auction-server/src/main.rs
+++ b/auction-server/src/main.rs
@@ -19,12 +19,12 @@ use {
         io::IsTerminal,
         time::Duration,
     },
-    tracing::Metadata,
+    tracing::{
+        Level,
+        Metadata,
+    },
     tracing_subscriber::{
-        filter::{
-            self,
-            LevelFilter,
-        },
+        filter,
         layer::SubscriberExt,
         util::SubscriberInitExt,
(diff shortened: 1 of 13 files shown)


auction-server/src/main.rs
```diff
@@ -46,6 +46,17 @@ fn is_internal(metadata: &Metadata) -> bool {
     metadata.target().starts_with("auction_server")
 }
 
+fn is_loggable(metadata: &Metadata) -> bool {
+    metadata.level() <= &Level::INFO
```
Dewyer, line 50: this would mean only info level logs will be logged right? the logger already has an info level filter, and stuff higher would be filtered out here

Dewyer: I dont think the logger config needs to be changed, and the tracer should pick up stuff on the trace level by default and the traces would be considered internall as well, also there is a crate / module level filter we could register instead too, that allos more fine grained controll over where logs forwarded from https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html

Dewyer: mb

Dewyer (approved):

PR #478, opened 2025-04-08
Update quote opportunity to include absolute deadline

danimhr:
This PR aims to update the quote opportunties to include absolute minimum deadline instead of relative lifetime field.

### DIFF
--- a/auction-server/src/api.rs
+++ b/auction-server/src/api.rs
@@ -77,7 +77,6 @@ use {
             atomic::Ordering,
             Arc,
         },
-        time::Duration,
     },
     time::OffsetDateTime,
     tower_http::cors::CorsLayer,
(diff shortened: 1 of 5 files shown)


auction-server/src/api.rs
```diff
@@ -538,7 +537,7 @@ impl RestError {
             ),
             RestError::InvalidDeadline { deadline, minimum } => (
                 StatusCode::BAD_REQUEST,
-                format!("Bid deadline of {:?} is too short, bid must be valid for at least {:?} seconds", deadline, minimum),
+                format!("Bid deadline {:?} is too early; it must be remain valid at least until {:?}", deadline, minimum),
```
m30m, line 540: ` it must be remain` -> ` it must remain`

m30m (approved):

PR #481, opened 2025-04-08
chore: expose cancellable to searchers

guibescos:


CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to the searcher sdks will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [Unreleased]
+
+### Added
+
+- The minimum lifetime of swap quotes is now configurable. Users can provide a `minimum_lifetime` which represents for how long the quote they receive should be valid and defaults to the old value of 10 seconds. Searchers receive a `minimum_deadline` as a timestamp in the opportunity parameters that they should use as the deadline for their quotes. [482](https://github.com/pyth-network/per/pull/482)
```
danimhr, line 11: Thank you for updating here :D

auction-server/src/auction/service/verification.rs
```diff
@@ -2074,24 +2073,11 @@ mod tests {
...
             permission_account: permission_account_user_token_wsol,
-            program: OpportunitySvmProgram::Swap(OpportunitySvmProgramSwap {
-                user_wallet_address,
-                platform_fee_bps: 0,
-                token_program_user: spl_token::id(),
-                token_program_searcher: spl_token::id(),
-                fee_token: fee_token.clone(),
-                referral_fee_bps,
-                user_mint_user_balance: LAMPORTS_PER_SOL,
-                token_account_initialization_configs: TokenAccountInitializationConfigs {
```
danimhr, line 2085: I think `token_account_initialization_configs` variable was different before. now you are setting it to default_test ....

Maybe you can get the default and override the `token_account_initialization_configs`?

auction-server/src/opportunity/repository/models.rs
```diff
@@ -99,6 +99,7 @@ pub struct OpportunityMetadataSvmProgramSwap {
     pub user_mint_user_balance:               u64,
     pub token_account_initialization_configs: TokenAccountInitializationConfigs,
     pub memo:                                 Option<String>,
+    pub cancellable:                          bool,
```
danimhr, line 102: I think we should make it optional as the old values does not contain this field.
With this version if you fetch the historical opportunity, it should crash and return 5xx as the swap opportunities in the db does not contain cancellable field inside their metadata in their json

danimhr: There also may exists other fields like `user_mint_user_balance` with the same issue as well :D

sdk/js/src/index.ts
```diff
@@ -520,12 +520,13 @@ export class Client {
         ? quoteRequest.userWallet.toBase58()
         : null,
       memo: quoteRequest.memo ?? null,
-      version: "v1" as const,
+      cancellable: quoteRequest.cancellable ?? undefined,
```
danimhr, line 523: it's better to use the same structure. if we wanna use undefined instead of null, could you please update the other fields like memo, and minimum_lifetime to be `undefined` instead of `null`?

sdk/js/src/serverTypes.d.ts
```diff
@@ -800,6 +800,11 @@ export interface components {
           slot: number;
         }
       | {
+          /**
+           * @description If true, bids to this opportunity can be cancelled by the searcher.
+           * @example true
+           */
+          cancellable: boolean;
```
danimhr, line 807: isn't the default value true here?
So if not provided or true, the quotes are cancellable?

auction-server/src/auction/service/verification.rs
guibescos: good catch

sdk/js/src/serverTypes.d.ts
guibescos: lgtm

danimhr (approved):

PR #483, opened 2025-04-09
Feat: add metrics around deadline validation

anihamde:
This PR adds some metrics around the validation of the deadline in `submit_quote`. This will help to track the rate of swaps failing due to late submission.

### DIFF
--- a/auction-server/src/auction/service/submit_quote.rs
+++ b/auction-server/src/auction/service/submit_quote.rs
@@ -17,7 +17,7 @@ use {
         },
         per_metrics::{
             SUBMIT_QUOTE_DEADLINE_BUFFER_METRIC,
-            SUBMIT_QUOTE_DEADLINE_CHECK_METRIC,
+            SUBMIT_QUOTE_DEADLINE_TOTAL,
         },
     },
     axum_prometheus::metrics,
@@ -141,7 +141,7 @@ impl Service<Svm> {
 
         if deadline_buffer_secs < MIN_DEADLINE_BUFFER_SECS {
             metrics::counter!(
-                SUBMIT_QUOTE_DEADLINE_CHECK_METRIC,
+                SUBMIT_QUOTE_DEADLINE_TOTAL,
                 &[
                     ("chain_id", chain_id.clone()),
                     ("result", "error".to_string()),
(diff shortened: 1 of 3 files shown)


auction-server/src/auction/service/submit_quote.rs
```diff
@@ -119,6 +122,41 @@ impl Service<Svm> {
...
+        &self,
+        chain_id: ChainId,
+        swap_args: express_relay::SwapArgs,
+    ) -> Result<(), RestError> {
+        let deadline_buffer_secs = swap_args.deadline - OffsetDateTime::now_utc().unix_timestamp();
+
+        tracing::info!("Deadline buffer: {}", deadline_buffer_secs);
+
+        metrics::histogram!(
+            "submit_quote_deadline_buffer",
```
danimhr, line 135: Instead of using a string here please add a constant name for this metric in the `per_metrics.rs` file?
sth like this: `TRANSACTION_LANDING_TIME_SVM_METRIC`

auction-server/src/auction/service/submit_quote.rs
```diff
@@ -119,6 +122,41 @@ impl Service<Svm> {
...
     }
 
+    fn check_deadline_buffer(
+        &self,
+        chain_id: ChainId,
+        swap_args: express_relay::SwapArgs,
+    ) -> Result<(), RestError> {
+        let deadline_buffer_secs = swap_args.deadline - OffsetDateTime::now_utc().unix_timestamp();
+
+        tracing::info!("Deadline buffer: {}", deadline_buffer_secs);
```
danimhr, line 132: Do we need this info?
we have the histogram

danimhr (approved):

auction-server/src/auction/service/submit_quote.rs
anihamde: lol debugging comment forgot to save before committing :/

auction-server/src/auction/service/submit_quote.rs
```diff
@@ -119,6 +122,41 @@ impl Service<Svm> {
         Ok(())
     }
 
+    fn check_deadline_buffer(
+        &self,
+        chain_id: ChainId,
+        swap_args: express_relay::SwapArgs,
+    ) -> Result<(), RestError> {
+        let deadline_buffer_secs = swap_args.deadline - OffsetDateTime::now_utc().unix_timestamp();
```
danimhr, line 130: This can be a negative value. Are we okay with having negative values in the histogram?

PR #484, opened 2025-04-09
chore: improve auction service injection by removing rw lock

Dewyer:


auction-server/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license-file = "license.txt"
 anchor-lang-idl = { version = "0.1.1", features = ["convert"] }
 
 [dependencies]
+arc-swap = "1.7.1"
```
danimhr, line 11: nit: Maybe add it to workspace?

danimhr (approved):

PR #486, opened 2025-04-09
Feat/bid submission failed status

danimhr:
This PR adds additional states for bids submitted by users that we choose not to submit on-chain for some reason.

### DIFF
--- a/auction-server/api-types/src/bid.rs
+++ b/auction-server/api-types/src/bid.rs
@@ -79,6 +79,15 @@ pub enum BidStatusEvm {
     },
 }
 
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum SubmissionFailedReason {
+    /// The bid was cancelled by the owner.
+    Cancelled,
+    /// The bid was submitted very late.
+    DeadlinePassed,
+}
+
 #[serde_as]
 #[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug)]
 #[serde(tag = "type", rename_all = "snake_case")]
@@ -148,6 +157,16 @@ pub enum BidStatusSvm {
         #[schema(example = "Jb2urXPyEh4xiBgzYvwEFe4q1iMxG1DNxWGGQg94AmKgqFTwLAiTiHrYiYxwHUB4DV8u5ahNEVtMMDm3sNSRdTg", value_type = String)]
         #[serde_as(as = "DisplayFromStr")]
         result: Signature,
+    },
+    /// The Submission of the bid failed.
+    /// This can happend if the user try to submit the bid after it being cancelled by the owner.
+    /// Or if the user try to submit the bid after late while the bid is in the state of AwaitingSignature.
+    #[schema(title = "Cancelled")]
+    SubmissionFailed {
+        #[schema(example = "Jb2urXPyEh4xiBgzYvwEFe4q1iMxG1DNxWGGQg94AmKgqFTwLAiTiHrYiYxwHUB4DV8u5ahNEVtMMDm3sNSRdTg", value_type = String)]
+        #[serde_as(as = "DisplayFromStr")]
+        result: Signature,
+        reason: SubmissionFailedReason,
     },
 }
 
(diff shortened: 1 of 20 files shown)


auction-server/api-types/src/bid.rs
```diff
@@ -79,6 +79,15 @@ pub enum BidStatusEvm {
     },
 }
 
+#[derive(Serialize, Deserialize, ToSchema, Clone, PartialEq, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum SubmissionFailedReason {
+    /// The bid was cancelled by the owner.
```
m30m, line 85: owner means searcher?

auction-server/api-types/src/bid.rs
```diff
@@ -149,6 +158,16 @@ pub enum BidStatusSvm {
         #[serde_as(as = "DisplayFromStr")]
         result: Signature,
     },
+    /// The Submission of the bid failed.
+    /// This can happend if the user try to submit the bid after it being cancelled by the owner.
+    /// Or if the user try to submit the bid after late while the bid is in the state of AwaitingSignature.
```
m30m, line 163: what do you mean after late?

auction-server/src/auction/repository/models.rs
```diff
@@ -573,6 +607,20 @@ impl ModelTrait<Svm> for Svm {
                 bid.id,
                 BidStatus::AwaitingSignature as _,
             )),
+            entities::BidStatusSvm::SubmissionFailed { auction, reason } => {
+                let (status, expected_status) = match reason {
+                    entities::BidSubmissionFailedReason::Cancelled => (BidStatus::SubmissionFailedCancelled, BidStatus::Cancelled),
+                    entities::BidSubmissionFailedReason::DeadlinePassed => (BidStatus::SubmissionFailedDeadlinePassed, BidStatus::AwaitingSignature),
```
m30m, line 613: what happens if bid is both cancelled and deadline is passed?

auction-server/src/auction/service/submit_quote.rs
```diff
@@ -48,11 +48,30 @@ impl Service<Svm> {
         let winner_bid = auction
             .bids
             .iter()
-            .find(|bid| bid.status.is_awaiting_signature() || bid.status.is_sent_to_user_for_submission() || bid.status.is_submitted() || bid.status.is_cancelled())
+            .find(|bid| {
```
m30m, line 51: can't we only filter out lost bids and then match on `winner_bid.status`? The current if/else pattern and filtering seems a bit ad-hoc.

auction-server/src/auction/service/submit_quote.rs
```diff
@@ -181,7 +198,21 @@ impl Service<Svm> {
...
+                new_status: entities::BidStatusSvm::SubmissionFailed {
+                    auction: entities::BidStatusAuction {
+                        id: auction.id,
+                        tx_hash,
+                    },
+                    reason:  entities::BidSubmissionFailedReason::DeadlinePassed,
+                },
+            })
+            .await?;
+            return Err(RestError::BadParameters("Quote is expired.".to_string()));
```
m30m, line 214: can we use different status codes here? It gives us more visibility in grafana. For example 410 seems appropriate here. And maybe 409 when quote is cancelled 🤔

auction-server/api-types/src/bid.rs
danimhr: Yes. The owner of the bid

auction-server/src/auction/repository/models.rs
danimhr: Fixed

auction-server/src/auction/service/submit_quote.rs
danimhr: The lost bid is gonna be updated on a separate thread. So It's not safe to do it.

m30m (approved):
We should probably handle the cancel_bid where current bid status is `SubmissionFailedDeadlinePassed` more gracefully. We only return `Bid is only cancellable in awaiting_signature state` as error but it shouldn't return an error.

auction-server/src/auction/service/submit_quote.rs
danimhr: Done

PR #487, opened 2025-04-10
chore: Add test utils for svm rpc calls

Dewyer:
# About
We had a convo about how to actually go about testing svm close functionality, i looked into it, and thought about an inteligent canning solution to be the best.

Taken some insipration from mockall but ultimately, the key here is gona be adding domain specific trickery to apease the rpc caller :D and make the test writing a breeze at the same time.

LMK wdyt


auction-server/src/kernel/rpc_client_svm_tester.rs
```diff
@@ -0,0 +1,208 @@
...
+        RpcClient::new_sender(
+            RpcClientTesterSender {
+                tester: self.clone(),
+            },
+            RpcClientConfig::default(),
+        )
+    }
+
+    /// Panics if not all the canned responses have been consumed during the test
+    pub async fn check_all_uncanned(&self) {
```
guibescos, line 129: could we do this at drop?

guibescos (approved):

auction-server/src/opportunity/service/get_express_relay_metadata.rs
```diff
@@ -76,3 +76,104 @@ impl Service {
...
+    #[tokio::test]
+    async fn test_metadata_rpc_fetch() {
+        let rpc_client = RpcClientSvmTester::new();
+        let mock_db = MockDatabase::default();
+
+        let (service, _) =
+            Service::new_with_mocks_svm(DEFAULT_CHAIN_ID.to_string(), mock_db, &rpc_client);
+
+        let admin_pubkey = Pubkey::new_unique();
+        rpc_client
```
danimhr, line 141: I think instead of using canned stuff, we could just mock the rpc_client itself?
Why didnt we mock the rpc response instead of using a wrapper around the client?

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -874,14 +874,14 @@ mod tests {
         } = params;
 
         let chain_id = DEFAULT_CHAIN_ID.to_string();
-        let rpc_client = MockRpcClient::default();
```
danimhr, line 877: For example we could mock a specific function call response here

auction-server/src/opportunity/service/get_express_relay_metadata.rs
```diff
@@ -76,3 +76,104 @@ impl Service {
...
+
+        let params_received = service
+            .get_express_relay_metadata(GetExpressRelayMetadata {
+                chain_id: DEFAULT_CHAIN_ID.to_string(),
+            })
+            .await
+            .expect("Failed to get express relay metadata in test");
+
+        assert_eq!(params_received.admin, admin_pubkey);
+        rpc_client.check_all_uncanned().await;
```
danimhr, line 177: We can remove this line if we do this check while dropping

auction-server/src/kernel/rpc_client_svm_tester.rs
Dewyer: its async, async drop last time i checked wasnt too nice

auction-server/src/opportunity/service/get_express_relay_metadata.rs
Dewyer: i was like wondering which way to go, and tbh solana domain logic is so in all the services that moving that logic out and not testing it might cause further issues idk

auction-server/src/opportunity/service/get_quote.rs
Dewyer: but at the same time, canning these svm rpc repsonses didnt prove to be that difficult, overall, its a few lines of code

auction-server/src/kernel/rpc_client_svm_tester.rs
Dewyer: yeah there is just nothing really to poll once u drop i suppose

PR #488, opened 2025-04-10
fix: opportunities error handling

guibescos:


danimhr:
As we removed the EVM side codes, I think we can close this PR and it's kind of not needed anymore

PR #489, opened 2025-04-10
Add bid submission failed status to sdk

danimhr:
This PR aims to add bid submission failed status to sdks

### DIFF
--- a/sdk/python/express_relay/models/svm.py
+++ b/sdk/python/express_relay/models/svm.py
@@ -5,6 +5,7 @@ from typing import Annotated, Any, ClassVar, Literal
 
 from express_relay.models.base import (
     BidStatusVariantsSvm,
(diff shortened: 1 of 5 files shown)


Dewyer (approved):

sdk/python/express_relay/models/svm.py
```diff
@@ -399,6 +402,14 @@ def check_result(self):
             ), "bid result should not be empty when status is not pending or lost"
         return self
 
+    @model_validator(mode="after")
+    def check_failed_reason(self):
+        if self.type == BidStatusVariantsSvm.SUBMISSION_FAILED:
+            assert (
```
Dewyer, line 408: q: for backwards compat reasons, we might want to allow reason to be null right? not sure if its populated for older data

PR #505, opened 2025-04-14
feat: Remove generics from opportunity service

Dewyer:


auction-server/src/opportunity/repository/mod.rs
```diff
@@ -22,29 +22,22 @@ mod refresh_in_memory_opportunity;
 mod remove_opportunities;
 mod remove_opportunity;
 
+use crate::opportunity::entities::OpportunitySvm;
```
danimhr, line 25: Maybe it's better to not important the object. You may import the `entities` and then use `entities::OpportunitySvm`

auction-server/src/opportunity/service/remove_invalid_or_expired_opportunities.rs
```diff
@@ -67,38 +53,9 @@ where
                         reason = ?reason,
                         "Removing Opportunity",
                     );
-                    match self.repo.remove_opportunity(opportunity, reason).await {
```
danimhr, line 70: What happened to the remove opportunity logic here?

Dewyer: wrong way, good catch

danimhr (approved):

PR #507, opened 2025-04-15
Remove verification from the opportunity module

danimhr:
This PR aims to remove verification from the opportunity module


auction-server/src/opportunity/service/add_opportunity.rs
```diff
@@ -48,26 +45,16 @@ impl Service {
         &self,
         input: AddOpportunityInput,
     ) -> Result<OpportunitySvm, RestError> {
+        // Make sure the chain id is valid
+        self.get_config(&input.opportunity.core_fields.chain_id.clone())?;
```
Dewyer, line 49: this is also done in the caller of this fn, when preparing the payload, we can also do it here

Dewyer (approved):

PR #508, opened 2025-04-15
chore:  add test for  opportunity removal

Dewyer:


auction-server/src/opportunity/service/remove_invalid_or_expired_opportunities.rs
```diff
@@ -63,3 +63,209 @@ impl Service {
...
+            }],
+            buy_tokens: vec![TokenAmountSvm {
+                token:  Pubkey::new_unique(),
+                amount: 1,
+            }],
+            creation_time,
+            refresh_time,
+            router: Pubkey::new_unique(),
+            permission_account: Pubkey::new_unique(),
+            program: OpportunitySvmProgram::Swap(OpportunitySvmProgramSwap {
```
guibescos, line 129: can you use `OpportunitySvmProgramSwap::default_test_with_user_wallet_address` here?

auction-server/src/opportunity/service/remove_invalid_or_expired_opportunities.rs
```diff
@@ -63,3 +63,209 @@ impl Service {
...
+        // check opportunities
+        let all_opportunities_by_test_key = service
+            .repo
+            .get_in_memory_opportunities()
+            .await
+            .remove(&live_opportunity.get_key())
+            .expect("opportunity should exist");
+
+        assert_eq!(all_opportunities_by_test_key, vec![live_opportunity]);
+        assert!(ws_receiver.is_empty());
```
guibescos, line 225: why is this empty, didn't we remove `expired_opportunity`

Dewyer: we removed one, but the point is that we didnt remove all of them under the permission key, this is specifically what the code is checking for, not sure if its sound logic but it works

PR #511, opened 2025-04-15
feat: Add local token support to rfq stress test script

Dewyer:


scripts/rfq-stress-test/tokens.prod.json
```diff
@@ -0,0 +1,5 @@
+[
+  "So11111111111111111111111111111111111111112",
+  "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
+  "Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB"
```
guibescos, line 4: `tokens.sample.json` ?

guibescos (approved):

scripts/rfq-stress-test/tokens.prod.json
Dewyer: well, these are the actual tokens on chain i think so we can keep it in there imo, u can copy paste this into tokens and change it as u like

Dewyer: also it should make it clear to users that these are mint addresses

PR #512, opened 2025-04-15
feat: Remove ethers by moving bytes hex impl

Dewyer:


auction-server/src/opportunity/service/add_opportunity.rs
```diff
@@ -20,7 +20,7 @@ pub struct AddOpportunityInput {
 #[derive(Debug, Clone)]
 enum OpportunityAction {
     Add,
-    Refresh(OpportunitySvm),
+    Refresh(Box<OpportunitySvm>),
```
danimhr, line 23: Why boxed?

auction-server/src/opportunity/service/add_opportunity.rs
```diff
@@ -136,7 +135,7 @@ mod tests {
         let buy_token = Pubkey::new_unique();
         let buy_amount = 1;
 
-        let permission_key = Bytes::from_hex("0xdeadbeef").unwrap();
+        let permission_key = PermissionKeySvm::try_from(&[1; 65][..]).expect("permission key");
```
danimhr, line 138: Shouldnt it be length 65?

auction-server/src/opportunity/service/remove_invalid_or_expired_opportunities.rs
```diff
@@ -109,7 +106,7 @@ mod test {
     ) -> OpportunitySvm {
         OpportunitySvm {
             id: Uuid::from_u128(id),
-            permission_key: Bytes::from_hex("0xdeadbeef").unwrap(),
+            permission_key: PermissionKeySvm::try_from(&[2; 65][..]).expect("permission key"),
```
danimhr, line 109: Same

auction-server/src/opportunity/service/add_opportunity.rs
Dewyer: clippy and enum size, its an easy change and keeps this enums size to like n bytes instead of 400

auction-server/src/opportunity/service/add_opportunity.rs
Dewyer: i think its length 65, 65 1ns

auction-server/src/opportunity/service/remove_invalid_or_expired_opportunities.rs
Dewyer: should init w 65 2s

auction-server/src/opportunity/service/add_opportunity.rs
danimhr: nit: While this will put things inside heap, maybe it's a better idea to ask clippy to ignore the error for this line

danimhr (approved):

PR #517, opened 2025-04-17
Update submit quote error codes

danimhr:
This PR aims to update error codes of the submit quote service / api

### DIFF
--- a/auction-server/src/auction/service/submit_quote.rs
+++ b/auction-server/src/auction/service/submit_quote.rs
@@ -98,23 +98,22 @@ impl Service {
         if !self.is_within_deadline_buffer(bid_latest_version.chain_id.clone(), swap_args) {
             if bid_latest_version.status.is_sent_to_user_for_submission() {
                 // TODO we are losing information here, need a better way for handling this situation
-                // NOTE: These bids maybe submitted by the user, so we need to update the status to submission failed
+                // NOTE: These bids maybe submitted by the user, so we don't need to update the status to submission failed
                 tracing::warn!(bid_id = ?bid_latest_version.id, auction_id = ?auction.id, "A non cancellable bid is submitted after the deadline buffer");
-                return Err(RestError::QuoteIsExpired);
-            }
-
-            let tx_hash = bid_latest_version.chain_data.transaction.signatures[0];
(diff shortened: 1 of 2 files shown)


Dewyer (approved):

auction-server/src/auction/service/submit_quote.rs
```diff
@@ -52,16 +52,7 @@ impl Service {
...
-                "This quote has already been finalized. No further changes are allowed."
-                    .to_string(),
-            ))?;
-
-        if winner_bid.status.is_submitted() {
-            return Err(RestError::BadParameters(
-                "Quote is already submitted on-chain.".to_string(),
-            ));
-        }
+            .ok_or(RestError::QuoteIsFinalized)?;
```
anihamde, line 55: from purely readability, this error is confusing given the name of the function, `get_winner_bid`: it seems weird that a get method returns an error when the quote is already submitted. maybe the error message should be more verbose, or the function name should be updated to reflect that this get method is only called to retrieve the bid to then submit it? so maybe rename to `get_winner_bid_for_submission` or something?

auction-server/src/auction/service/submit_quote.rs
```diff
@@ -93,12 +84,39 @@ impl Service {
...
         // Make sure the bid is still not cancelled, we also get the latest saved version of the auction
         let (auction, bid_latest_version) = self.get_winner_bid(auction.id).await?;
+        if bid_latest_version.status.is_submitted() {
+            return Ok(());
+        }
+
+        if !self.is_within_deadline_buffer(bid_latest_version.chain_id.clone(), swap_args) {
+            if bid_latest_version.status.is_sent_to_user_for_submission() {
+                // TODO we are losing information here, need a better way for handling this situation
+                // NOTE: These bids maybe submitted by the user, so we don't need to update the status to submission failed
```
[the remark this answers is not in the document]
danimhr: Lets discuss more about this

auction-server/src/auction/service/submit_quote.rs
danimhr: Agree. Will update it later or feel free to create the renaming PR sir

PR #519, opened 2025-04-17
fix: improve idlgen

guibescos:


anihamde (approved):
this is just improving the documentation, right?

PR #520, opened 2025-04-22
fix: Use last IP coming in the Webscoket client IP header

Dewyer:


danimhr:
I'm not sure if we should get the last one. This mentioned that the left most ip (comma separated) is the original ip
https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html#x-forwarded-for

PR #521, opened 2025-04-22
Feat: log other quoters

anihamde:
This PR adds the basic logic to log the responses of other quoters whenever a quote request is created. The other quoters' info is stored in the opportunity metadata. This PR leverages a Rust crate derived from [this fork](https://github.com/anihamde/jupiter-swap-api-client) of the Jupiter Swap API client, to source quotes from the Jupiter Ultra endpoint.

Currently the other quoters' info is added to the opportunity retroactively following the creation and broadcasting of the opportunity.

TODOs

- [x] getting the responses of other quoters should run in a separate thread rather than in the main thread, which will block creation of the opportunity and delay the auction
- [x] consider implications of rate limits


auction-server/src/opportunity/repository/models.rs
```diff
@@ -89,6 +90,7 @@ pub struct OpportunityMetadataSvmProgramSwap {
     #[serde(default = "default_cancellable")]
     pub cancellable:                          bool,
     pub minimum_lifetime:                     Option<u32>,
+    pub other_quotes:                         Vec<OtherQuote>,
```
guibescos, line 93: It doesn't feel great to have such a big struct here. 
Could we alternatively just return the searcher token amount?

PR #523, opened 2025-04-23
feat: Add token mint whitelisting

Dewyer:


auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -204,6 +194,26 @@ impl Service {
             } => (user_token, searcher_token.token),
         };
         let config = self.get_config(&quote_create.chain_id)?;
+
+        // validate token mints being whitelisted at the earliest point to fail fast
+        if !config.token_whitelist.is_token_mint_allowed(&mint_user) {
+            return Err(RestError::TokenMintNotAllowed("User".to_string()));
```
guibescos, line 200: I feel like in the API namespace it's better so use the input_token_mint/output_token_mint terminology

PR #525, opened 2025-04-23
fix: remove redundantly serialized memo

Dewyer:


auction-server/src/opportunity/repository/models.rs
```diff
@@ -85,9 +85,11 @@ pub struct OpportunityMetadataSvmProgramSwap {
     pub user_mint_user_balance:               u64,
     #[serde(default = "TokenAccountInitializationConfigs::searcher_payer")]
     pub token_account_initialization_configs: TokenAccountInitializationConfigs,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
danimhr, line 88: Can you check if the deserialization also works?

danimhr (approved):

auction-server/src/opportunity/repository/models.rs
Dewyer: the roundtrip test checks ser and deser too

PR #527, opened 2025-04-28
Add referral fee ppm to auction server

danimhr:
This PR aims to add referral fee ppm to auction server quote apis

### DIFF
--- a/auction-server/src/api.rs
+++ b/auction-server/src/api.rs
@@ -387,7 +387,7 @@ impl std::fmt::Display for SwapInstructionError {
             ),
             SwapInstructionError::PlatformFee { expected, found } => write!(
                 f,
-                "Invalid platform fee bps {} in swap instruction data. Value does not match the platform fee bps in swap opportunity {}",
+                "Invalid platform fee ppm {} in swap instruction data. Value does not match the platform fee ppm in swap opportunity {}",
                 found, expected
             ),
         }
--- a/auction-server/src/api.rs
+++ b/auction-server/src/api.rs
@@ -319,8 +319,12 @@ pub enum SwapInstructionError {
         found:    express_relay::FeeToken,
     },
     ReferralFee {
-        expected: u16,
-        found:    u16,
+        expected: u64,
+        found:    u64,
+    },
+    PlatformFee {
+        expected: u64,
+        found:    u64,
     },
     AssociatedRouterTokenAccount {
         expected: Pubkey,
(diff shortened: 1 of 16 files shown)


auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -337,14 +338,19 @@ impl Service {
             &tokens_for_permission,
             &user_wallet_address,
             &router_token_account,
-            referral_fee_info.referral_fee_bps,
+            referral_fee_info.referral_fee_ppm,
         );
 
+        let referral_fee_bps =
```
anihamde, line 344: maybe the TODO note about deleting this should live here...

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -337,14 +338,19 @@ impl Service {
...
+            u16::try_from(referral_fee_info.referral_fee_ppm / FEE_BPS_TO_PPM)
```
anihamde, line 345: this should technically be ceiling divide, right?

contracts/svm/programs/express_relay/src/swap.rs
```diff
@@ -202,14 +202,14 @@ impl ExpressRelayMetadata {
 
     pub fn compute_swap_fees_with_default_platform_fee(
         &self,
-        referral_fee_bps: u16,
+        referral_fee_ppm: u64,
         amount: u64,
     ) -> Result<SwapFeesWithRemainingAmount> {
-        let referral_fee_ppms = u64::from(referral_fee_bps) * FEE_BPS_TO_PPM;
         let swap_platform_fees_ppms = self.swap_platform_fee_bps * FEE_BPS_TO_PPM;
```
anihamde, line 208: typo: ppms?

anihamde (approved):

auction-server/src/api.rs
```diff
@@ -381,6 +385,11 @@ impl std::fmt::Display for SwapInstructionError {
                 "Associated token account for router does not match. Expected: {:?} found: {:?}",
                 expected, found
             ),
+            SwapInstructionError::PlatformFee { expected, found } => write!(
+                f,
+                "Invalid platform fee bps {} in swap instruction data. Value does not match the platform fee bps in swap opportunity {}",
```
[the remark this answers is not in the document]
danimhr: RN it's in bps. Let me see if we can update it or not

auction-server/src/opportunity/service/get_quote.rs
danimhr: I think if we delete the other place, then the compiler will raise error

auction-server/src/opportunity/service/get_quote.rs
danimhr: I'm not sure about this. Hopefully this will never happen, but I prefer less fees during this conversions. Is it gonna produce bug anywhere?

PR #529, opened 2025-04-28
Update token account creation config based on fees

danimhr:
This PR aims to update the token account creation logic when fees are 0

### DIFF
--- a/auction-server/src/opportunity/service/get_quote.rs
+++ b/auction-server/src/opportunity/service/get_quote.rs
@@ -31,6 +31,7 @@ use {
                 self,
                 OpportunitySvmProgram,
                 OpportunitySvmProgramSwap,
+                TokenAccountInitializationConfig,
                 TokenAccountInitializationConfigs,
                 TokenAmountSvm,
             },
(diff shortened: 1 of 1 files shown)


auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -334,6 +335,19 @@ impl Service {
                 }
             };
 
+        if referral_fee_info.referral_fee_ppm == 0 {
+            // If the referral fee is 0, we can skip the initialization of the router token account
+            token_account_initialization_configs.router_fee_receiver_ta =
+                TokenAccountInitializationConfig::Unneeded;
+        }
+        if metadata.swap_platform_fee_bps == 0 {
+            // If the platform fee is 0, we can skip the initialization of the router token account
```
guibescos, line 344:
```suggestion
            // If the platform fee is 0, we can skip the initialization of the platform fee token accounts
```

guibescos (approved):
Makes sense, my first intuition is this should live in `get_token_account_initialization_configs` but happy to go with your call

PR #534, opened 2025-05-01
Add minimum number of DB pool connections

danimhr:
This PR aims to add min db pool connection config

### DIFF
--- a/auction-server/src/server.rs
+++ b/auction-server/src/server.rs
@@ -184,8 +184,13 @@ pub fn setup_metrics_recorder() -> Result<PrometheusHandle> {
 const NOTIFICATIONS_CHAN_LEN: usize = 1000;
 
 // TODO move to kernel repo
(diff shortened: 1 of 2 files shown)


auction-server/src/server.rs
```diff
@@ -184,16 +184,21 @@ pub fn setup_metrics_recorder() -> Result<PrometheusHandle> {
...
 
 // TODO move to kernel repo
-async fn create_pg_pool(database_url: &str, max_connections: u32) -> Result<PgPool> {
+async fn create_pg_pool(
+    database_url: &str,
+    min_connections: u32,
+    max_connections: u32,
+) -> Result<PgPool> {
     PgPoolOptions::new()
+        .min_connections(min_connections)
```
anihamde, line 193: what was the behavior before? defaulted to 0 min connections?

anihamde (approved):

auction-server/src/server.rs
danimhr: Yes

PR #546, opened 2025-05-09
fix: disable router_fee_receiver_ta optimisation

m30m:


guibescos (approved):

auction-server/src/opportunity/service/get_quote.rs
```diff
@@ -335,11 +335,12 @@ impl Service {
...
             // If the referral fee is 0, we can skip the initialization of the router token account
-            token_account_initialization_configs.router_fee_receiver_ta =
-                TokenAccountInitializationConfig::Unneeded;
+            // TODO: enable this at some point
+            // token_account_initialization_configs.router_fee_receiver_ta =
+            //     TokenAccountInitializationConfig::Unneeded;
         }
         if metadata.swap_platform_fee_bps == 0 {
-            // If the platform fee is 0, we can skip the initialization of the router token accounts
+            // If the platform fee is 0, we can skip the initialization of the express relay and relayer token accounts
```
guibescos, line 343:
```suggestion
            // If the platform fee is 0, we can skip the initialization of the express relay and relayer fee token accounts
```

PR #548, opened 2025-05-12
fix: properly expire awaiting_signature statuses in db

anihamde:
This PR fixes a bug in the db update query for `awaiting_signature` -> `expired`. We were seeing large numbers of `awaiting_signature` statuses lasting forever (and bid conclusion times not being updated), which suggested an issue in the db query to update the statuses.

Later after this fix is deployed, we should perform a migration to expire all `awaiting_signature` bids with no conclusion time that are at least somewhat old.


danimhr (commented):
Please update the version and create the queries as well :D

danimhr (approved):

PR #549, opened 2025-05-12
Add feature permission revoke for profiles

danimhr:
This PR adds the ability to revoke a specific feature permission for a given profile.


auction-server/migrations/20250512094618_permission.up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TYPE permission_state AS ENUM ('enabled', 'disabled');
+
+CREATE TABLE permission
+(
+    id         UUID             PRIMARY KEY,
+    profile_id UUID             NOT NULL REFERENCES profile(id) ON DELETE CASCADE,
+    feature    VARCHAR(255)     NOT NULL,
+    state      permission_state NOT NULL,
```
anihamde, line 8: this could just be a boolean?

auction-server/api-types/src/profile.rs
```diff
@@ -60,6 +60,35 @@ pub struct Profile {
...
+#[derive(Serialize, Deserialize, ToSchema, Clone, ToResponse)]
+#[serde(rename_all = "snake_case")]
+pub enum PermissionFeature {
+    /// The feature for which the permission is for.
+    CancelQuote,
+}
+
+#[derive(Serialize, Deserialize, ToSchema, Clone, ToResponse)]
+#[serde(rename_all = "snake_case")]
+pub enum PermissionState {
```
anihamde, line 72: this can also be a boolean? unless you anticipate this having more states soon?

auction-server/src/state.rs
```diff
@@ -278,4 +281,60 @@ impl Store {
...
+    pub async fn is_permitted(
+        &self,
+        profile_id: models::ProfileId,
+        feature: models::PermissionFeature,
+    ) -> Result<bool, RestError> {
+        let permissions = self.permissions.read().await;
+        if let Some(permission) = permissions.get(&(profile_id, feature.clone())) {
+            Ok(permission.state == models::PermissionState::Enabled)
+        } else {
+            Ok(true)
```
anihamde, line 337: i think cancel quote shouldn't be allowed if a permission hasn't been created? aka this should be whitelist, not blacklist

anihamde: in other words, you need to have Permission enabled in order to cancel quote. if permission is disabled, this fails. if the permission doesn't exist for this `profile_id`, it fails as well

m30m (commented):
I think this adds a double meaning to the term `permission` into the rep. Maybe use something like `Privilege`?

auction-server/src/state.rs
anihamde: i think cancellation should be a privileged action because it has costs on the system. if anyone abuses it, then that can cause bad impacts on downstream consumers. thus, it should be enabled for trusted parties and not for any anonymous integrator

danimhr: Yeah I agree with your point. For this reason the api is behind the authentication module. It means you can only access the API with API KEY. So by default it's enabled if you have API KEY. If you have API KEY and abuse the feature, then we'll revoke your access by adding a disabled row to your profile Id Privilege model.

anihamde: you're right that it is behind the authentication guard. but if an API key passes the authentication guard but doesn't have the Privilege, as it stands right now the cancellation will be allowed. I think that the default behavior should be that it is not allowed.

for example, right now, if there is an API key for a random entity that isn't a searcher, then that entity can make a bid and then cancel it.

danimhr: I'm still not fully convinced this is better. Using an API key is already a form of whitelisting, right? We only generate keys for trusted partners, and if any feature is misused, we can just revoke the privilege of that feature.

Introducing an additional layer feels like double whitelisting, which seems unnecessary to me.

Also, I’m not too keen on the idea of having to create a migration every time we add a new feature behind the privileged module—just to grant access to all existing profiles or API keys.

m30m (approved):

PR #555, opened 2025-05-14
Add profile to ws and axum http reqs

danimhr:
This PR aims to add profile name to ws and axum requests

### DIFF
--- a/auction-server/src/api.rs
+++ b/auction-server/src/api.rs
@@ -49,7 +49,11 @@ use {
         },
         TypedHeader,
     },
-    axum_prometheus::metrics,
+    axum_prometheus::{
+        metrics,
+        AXUM_HTTP_REQUESTS_DURATION_SECONDS,
+        AXUM_HTTP_REQUESTS_TOTAL,
+    },
     clap::crate_version,
     express_relay_api_types::{
         self as api_types,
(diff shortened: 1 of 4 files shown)


auction-server/src/api.rs
```diff
@@ -750,11 +745,43 @@ impl WrappedRouter {
...
+    let status = response.status().as_u16().to_string();
+
+    let labels = [
+        ("method", method.to_string()),
+        ("endpoint", endpoint),
+        ("status", status),
+        ("profile", profile),
+    ];
+
+    metrics::counter!("axum_http_requests_total", &labels).increment(1);
```
m30m, line 778: let's use the constants from the package:
```rs
axum_prometheus::{AXUM_HTTP_REQUESTS_DURATION_SECONDS, AXUM_HTTP_REQUESTS_TOTAL}
```

auction-server/src/api.rs
```diff
@@ -750,11 +745,43 @@ impl WrappedRouter {
...
+    metrics::histogram!("axum_http_requests_duration_seconds_bucket", &labels).record(latency);
```
m30m, line 779: I don't think you need to add the '_bucket' suffic

PR #558, opened 2025-05-21
Add batch inserter to clickhouse

danimhr:
This PR aims to insert rows to clickhouse using an inserter

### DIFF
--- a/auction-server/src/kernel/analytics_db.rs
+++ b/auction-server/src/kernel/analytics_db.rs
@@ -1,0 +1,53 @@
+use {
+    clickhouse::Row,
+    serde::Serialize,
+    std::time::Duration,
+    tokio::sync::mpsc,
+};
+
+pub struct ClickhouseInserter<T> {
+    pub sender: mpsc::Sender<T>,
+}
+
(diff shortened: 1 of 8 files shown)


auction-server/src/kernel/analytics_db.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub sender: mpsc::Sender<T>,
+}
+
+impl<T: Row + Serialize + Send + Sync + 'static> ClickhouseInserter<T> {
+    async fn run(client: clickhouse::Client, table_name: String, mut rx: mpsc::Receiver<T>) {
+        match client.inserter(&table_name) {
+            Ok(inserter) => {
+                let mut inserter = inserter
+                    .with_period(Some(Duration::from_secs(1)))
+                    .with_max_rows(100)
```
anihamde, line 18: maybe these should be constants defined above? also can you leave some comments to make these params more understandable?

anihamde: in particular some comments to explain how the batching exactly works, since it isn't obvious from these params

PR #560, opened 2025-05-26
Add lazer price for analytics

danimhr:
This PR aims to add price for bids and opportunities based on their token amounts. It uses the Pyth Lazer price feed subscription.


auction-server/src/kernel/pyth_lazer.rs
```diff
@@ -0,0 +1,312 @@
...
+            price_feeds.insert(price_feed.id, price_feed);
+        }
+        PythLazer {
+            url,
+            api_key,
+            price_feeds,
+        }
+    }
+
+    pub async fn subscribe(&self) -> anyhow::Result<broadcast::Receiver<StreamUpdatedMessage>> {
```
anihamde, line 244: re most of the content in this file: it seems like this should be defined/maintained in lazer consumer packages? [this](https://github.com/pyth-network/pyth-crosschain/blob/main/lazer/sdk/rust/client/src/lib.rs) is a simple version in crosschain, any reason you defined a lot of the classes and subscription logic here?

auction-server/src/auction/repository/add_bid_analytics.rs
```diff
@@ -97,12 +118,12 @@ impl Repository {
                     searcher_token_mint: mint_searcher.to_string(),
                     searcher_token_amount: amount_searcher,
                     // TODO Fill this in
-                    searcher_token_usd_price: None,
+                    searcher_token_usd_price,
 
                     user_token_mint: mint_user.to_string(),
                     user_token_amount: amount_user,
                     // TODO Fill this in
```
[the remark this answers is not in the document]
anihamde: although if this is meant to be price, it's being calculated incorrectly?

auction-server/src/auction/repository/add_bid_analytics.rs
```diff
@@ -97,12 +118,12 @@ impl Repository {
                     searcher_token_mint: mint_searcher.to_string(),
                     searcher_token_amount: amount_searcher,
                     // TODO Fill this in
-                    searcher_token_usd_price: None,
+                    searcher_token_usd_price,
```
anihamde, line 121: one thought: the schema we decided upon doesn't have a field for `price_timestamp`, but i think it would be good to add that. that way, we can tell if any prices are stale. if it's not too hard to add, i would add that

auction-server/src/kernel/pyth_lazer.rs
```diff
@@ -0,0 +1,312 @@
...
+        })?;
+        ws_stream.send(Message::Text(message)).await?;
+        let (update_sender, update_receiver) = broadcast::channel(1000);
+
+        tokio::spawn(WsClient::run(ws_stream, update_sender));
+        Ok(update_receiver)
+    }
+}
+
+pub fn calculate_final_amount(
```
anihamde, line 268: this function is named confusingly. it's returning the notional value, not the amount.

auction-server/src/auction/service/update_bid_status.rs
```diff
@@ -19,6 +21,49 @@ pub struct UpdateBidStatusInput {
...
+
+                let user_mint = self
+                    .opportunity_service
+                    .get_token_mint(GetTokenMintInput {
+                        chain_id: self.config.chain_id.clone(),
+                        mint:     data.accounts.mint_user,
+                    })
+                    .await?;
+
+                HashMap::from([
```
anihamde, line 47: maybe this should just be what `get_token_mint` caches and returns?

auction-server/src/kernel/pyth_lazer.rs
```diff
@@ -0,0 +1,312 @@
...
+                    tracing::error!(exponent = ?exponent, "Overflow in exponent divisor calculation for lazer price");
+                    return None;
+                }
+            };
+            let value = (amount as u128)
+                .checked_mul(price as u128)? // scale up price
+                .checked_div(unit_divisor)?
+                .checked_div(exponent_divisor)?;
+
+            let value_f64 = value as f64; // scale back down to original price
```
anihamde, line 299: why cast to `f64`? why not leave it as a `u128` or cast it to a `u64`?

auction-server/src/kernel/workers.rs
```diff
@@ -0,0 +1,74 @@
...
+                                }
+                            }
+                        }
+                        None => {
+                            tracing::warn!(id=?price.id, "Lazer price feed not found");
+                            None
+                        }
+                    }
+                });
+                let mut prices = store.prices.write().await;
```
anihamde, line 63: should this be in a dashmap? allows for better concurrency control

auction-server/src/opportunity/entities/token.rs
```diff
@@ -36,3 +36,10 @@ impl From<api::TokenAmountSvm> for TokenAmountSvm {
         }
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct TokenMint {
+    pub mint:     Pubkey,
+    pub decimals: u8,
+    pub owner:    Pubkey,
```
anihamde, line 44: i prefer `token_program` for clarity. technically owner is unambiguous, but there are some other terms that are close enough (the mint, freeze, etc. authorities) such that `token_program` is just clearer

auction-server/src/kernel/workers.rs
anihamde: this should also have some protection against updating with stale prices. if `update_time < currently_stored_time`, do nothing

auction-server/src/kernel/pyth_lazer.rs
```diff
@@ -0,0 +1,312 @@
...
+            .headers_mut()
+            .insert("Authorization", bearer_token.parse()?);
+
+        let (mut ws_stream, _) = connect_async(request).await?;
+        let message = serde_json::to_string(&SubscribeRequest {
+            id:             1,
+            request_type:   SubscribeRequestType::Subscribe,
+            price_feed_ids: self.price_feeds.keys().cloned().collect(),
+            properties:     vec![SubscribeRequestProperties::Price],
+            channel:        SubscribeRequestChannel::FixedRate200Ms,
```
anihamde, line 257: why always 200 ms? you can pull the lazer feeds' metadata and get the highest frequency channel if you want?

auction-server/src/state.rs
```diff
@@ -71,13 +72,20 @@ pub struct ServerState {
 
 pub type PrivilegeKey = (models::ProfileId, models::PrivilegeFeature);
 
+#[derive(Clone, Debug)]
+pub struct Price {
```
anihamde, line 76: i think you can pull this type from an existing pyth sdk crate?

auction-server/src/auction/repository/add_bid_analytics.rs
```diff
@@ -59,6 +67,19 @@ impl Repository {
             }
             entities::BidTransactionData::Swap(transaction_data) => {
                 let status_reason = Svm::get_bid_status_reason(&bid.status);
+                let mint_user = transaction_data.accounts.mint_user;
+                let user_token_usd_price = calculate_final_amount(
```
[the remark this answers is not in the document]
danimhr: I suggest we deploy it first, then review any errors that occur and check how much data has the price set to NULL. Based on that, we can decide on our next steps.

auction-server/src/kernel/pyth_lazer.rs
danimhr: Let me refactor this part in another PR. Lets deploy the current version and then I will replace this file with their Rust SDK

auction-server/src/auction/service/update_bid_status.rs
danimhr: I didnt want to use the Opportunity Entity variable inside the Auction service :D

auction-server/src/kernel/workers.rs
danimhr: We have RwLock around it, so we can use HashMap here. I will add time in another PR if you are okay?

auction-server/src/state.rs
danimhr: Let me refactor this part in a separate PR

auction-server/src/kernel/pyth_lazer.rs
danimhr: Sorry forgot to add precision before. Now it's fixed. Tnx for the comment

auction-server/src/auction/repository/add_bid_analytics.rs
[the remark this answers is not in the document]
anihamde: i think create a separate column for the time in the db and insert the price update time as well

auction-server/src/auction/repository/add_bid_analytics.rs
```diff
@@ -68,13 +68,13 @@ impl Repository {
             entities::BidTransactionData::Swap(transaction_data) => {
                 let status_reason = Svm::get_bid_status_reason(&bid.status);
                 let mint_user = transaction_data.accounts.mint_user;
-                let user_token_usd_price = calculate_final_amount(
+                let user_token_usd_price = calculate_notional_value(
```
anihamde, line 71: this variable should be named `user_token_notional_value` instead of `user_token_usd_price`

anihamde: or you can leave it as price and just not calculate the notional value yourself. instead you can literally just store the decimals-adjusted price here with a bunch of precision

anihamde: otherwise, the column name in the db `buy_token_usd_price` is very misleading. this refers to the USD price of the token, not the notional value of the total amount of the tokens

anihamde (approved):

auction-server/src/kernel/workers.rs
danimhr: Added a task for it

PR #564, opened 2025-05-30
Fix: zero amount opportunity bug

anihamde:
This PR fixes a bug whereby opportunities with 0 amount specified could be created, added to the db, and broadcast to subscribers.


m30m (approved):
can you add a check on line 236 and return an error (fees to high or something) if denominator is 0?

PR #566, opened 2025-06-03
Feat: delete pg history

anihamde:
This PR adds a loop to delete old rows in the Postgres `bid` and `opportunity` tables. The arguments are optional, and, if not provided, will not induce any deletion.

### DIFF
--- a/auction-server/src/config.rs
+++ b/auction-server/src/config.rs
@@ -93,7 +93,7 @@ pub struct DeletePgRowsOptions {
     /// How often to delete rows from the database.
     #[arg(long = "delete-interval-seconds")]
     #[arg(env = "DELETE_INTERVAL_SECONDS")]
-    #[arg(default_value = "60")]
+    #[arg(default_value = "1")]
     pub delete_interval_secs: u64,
 
     /// The threshold staleness for whether a row should be deleted.
--- a/auction-server/src/kernel/workers.rs
+++ b/auction-server/src/kernel/workers.rs
@@ -85,7 +85,7 @@ pub async fn run_price_subscription(
 }
 
 
-const DELETE_BATCH_SIZE: u64 = 1000;
+const DELETE_BATCH_SIZE: u64 = 5000;
 
 pub async fn run_delete_pg_db_history(
     db: &PgPool,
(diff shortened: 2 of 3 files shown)


auction-server/src/kernel/workers.rs
```diff
@@ -69,6 +78,51 @@ pub async fn run_price_subscription(
...
+            tracing::info!("Starting delete PG DB history worker, deleting every {} seconds rows that are {} seconds stale...", delete_interval_secs, delete_threshold_secs);
+            let mut delete_history_interval =
+                tokio::time::interval(Duration::from_secs(delete_interval_secs));
+
+            while !SHOULD_EXIT.load(Ordering::Acquire) {
+                tokio::select! {
+                    _ = delete_history_interval.tick() => {
+                        let threshold = OffsetDateTime::now_utc() - Duration::from_secs(delete_threshold_secs);
+
+                        sqlx::query!(
```
danimhr, line 104: This query may take a long time, I suggest to delete limited number of rows each time. Sth like 1000

auction-server/src/config.rs
```diff
@@ -75,6 +75,46 @@ pub struct RunOptions {
...
     pub secret_key: String,
+
+    #[command(flatten)]
+    pub delete_pg_rows: DeletePgRowsOptions,
+}
+
+#[derive(Args, Clone, Debug)]
+#[command(next_help_heading = "Delete PG Rows Options")]
+#[group(id = "DeletePgRows")]
+pub struct DeletePgRowsOptions {
```
danimhr, line 86: Isnt it better to have a default variable here?
If we wanna turn it on and off, we can add another variable.

auction-server/src/config.rs
```diff
@@ -75,6 +75,32 @@ pub struct RunOptions {
...
+    /// Whether to enable the deletion of rows from the database.
+    #[arg(long = "delete-enabled")]
+    #[arg(env = "DELETE_ENABLED")]
+    #[arg(default_value = "true")]
+    pub delete_enabled: bool,
+
+    /// How often to delete rows from the database.
+    #[arg(long = "delete-interval-seconds")]
+    #[arg(env = "DELETE_INTERVAL_SECONDS")]
+    #[arg(default_value = "60")]
```
danimhr, line 96: Maybe run it every second?

danimhr (approved):
Before merging, please make sure you are running the deletion loop every second.
I think you just need to update the default value for deletion loop.

+ I think 1000 is very small for deletion. We may have more bid than that per second. I suggest to increase it to 5000 and

PR #576, opened 2025-07-26
Fix: delete pg auction

anihamde:
This PR addresses the growing size of the auction postgres table. In order to manage this, we can delete historical rows from the auction PG table similar to how we do for the bid and opportunity tables. This PR adds a `tx_hash` column to the analytics bid tables and inserts that value as part of the `add_bid_analytics` method. It also adds a deletion loop for the auction postgres table.

### DIFF
--- a/auction-server/migrations/20250726182247_auction_creation_time_idx.up.sql
+++ b/auction-server/migrations/20250726182247_auction_creation_time_idx.up.sql
@@ -1,0 +1,1 @@
+CREATE INDEX auction_creation_time_idx ON auction (creation_time);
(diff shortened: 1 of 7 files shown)


auction-server/migrations/20250726182247_auction_creation_time_idx.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX auction_creation_time_idx ON auction (creation_time);
```
anihamde, line 1: this is going to be a big migration, but it may be necessary to efficiently delete rows from auction long-term. we could optionally make this query `CREATE INDEX CONCURRENTLY` which allows the index to [be created non-atomically](https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) and without blocking writes

danimhr (approved):

PR #582, opened 2025-08-05
Feat: allow arbitrary programs in swap transactions

anihamde:
This PR allows arbitrary program instructions in swap transactions.

Previously, the server maintained effectively a whitelist of program instructions that were allowed to be included in swap transactions, in order to protect users from signing transactions with unintended consequences.

In order to allow arbitrary program instructions while maintaining security, this PR changes the code to allow any of the aforesaid approved program instructions and allow an arbitrary program instruction provided that it does not invoke the user account (which is a signer of the transaction) in any of its accounts (including lookup table accounts). If this second condition is met, the user's signature on the transaction should not be actionable towards any logic in the arbitrary program instruction.

Some outstanding TODOs:

- [ ] make the extraction of lookup table accounts more efficient by bundling multiple lookups into a single call
- [ ] fix transfer checks to allow any transfers that do not invoke the user account

### DIFF
--- a/auction-server/src/api.rs
+++ b/auction-server/src/api.rs
@@ -131,7 +131,6 @@ pub enum InstructionError {
     MissingCreateAtaInstruction(Pubkey),
     InvalidMemoInstructionCount { expected: usize, found: usize },
     InvalidMemoString { expected: String, found: String },
-    UnsupportedInvocationOfUserWallet,
     UnapprovedProgramId(Pubkey),
 }
 
@@ -274,12 +273,6 @@ impl std::fmt::Display for InstructionError {
                     f,
                     "Invalid memo string in memo instruction. Expected: {:?} found: {:?}",
                     expected, found
-                )
-            }
-            InstructionError::UnsupportedInvocationOfUserWallet => {
-                write!(
-                    f,
-                    "Invocation of user wallet is not supported in this instruction"
                 )
             }
             InstructionError::UnapprovedProgramId(program_id) => {
--- a/auction-server/src/api.rs
+++ b/auction-server/src/api.rs
@@ -132,6 +132,7 @@ pub enum InstructionError {
     InvalidMemoInstructionCount { expected: usize, found: usize },
     InvalidMemoString { expected: String, found: String },
     UnsupportedInvocationOfUserWallet,
+    UnapprovedProgramId(Pubkey),
 }
 
 impl std::fmt::Display for InstructionError {
@@ -280,6 +281,9 @@ impl std::fmt::Display for InstructionError {
                     f,
                     "Invocation of user wallet is not supported in this instruction"
                 )
+            }
+            InstructionError::UnapprovedProgramId(program_id) => {
+                write!(f, "Instruction has unapproved program id: {:?}", program_id)
             }
         }
     }
--- a/auction-server/src/api.rs
+++ b/auction-server/src/api.rs
@@ -159,24 +159,24 @@ impl std::fmt::Display for InstructionError {
             }
             InstructionError::InvalidUserTransferInstructionsCount { found } => {
                 match found {
-                    0 => write!(f, "At least one sol transfer instruction from the user wallet account is required"),
-                    _ => write!(f, "Invalid number ({}) of sol transfer instructions from the user wallet account", found),
+                    0 => write!(f, "At least one SOL transfer instruction from the user wallet account is required"),
+                    _ => write!(f, "Invalid number ({}) of SOL transfer instructions from the user wallet account", found),
                 }
             }
             InstructionError::InvalidUserAccountTransferInstruction => {
-                write!(f, "Invalid sol transfer instruction from the user account.")
+                write!(f, "Invalid SOL transfer instruction from the user account.")
             }
             InstructionError::InvalidToAccountTransferInstruction { expected, found } => {
                 write!(
                     f,
-                    "Invalid to account in sol transfer instruction. Expected: {:?} found: {:?}",
+                    "Invalid to account in SOL transfer instruction. Expected: {:?} found: {:?}",
                     expected, found
                 )
             }
             InstructionError::InvalidAmountTransferInstruction { expected, found } => {
                 write!(
                     f,
-                    "Invalid amount in sol transfer instruction. Expected: {:?} found: {:?}",
+                    "Invalid amount in SOL transfer instruction. Expected: {:?} found: {:?}",
                     expected, found
                 )
             }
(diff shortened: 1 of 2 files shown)


auction-server/src/auction/service/verification.rs
```diff
@@ -2288,27 +2358,40 @@ mod tests {
...
-            let result = get_verify_bid_result(
-                service.clone(),
-                searcher,
-                vec![instruction],
-                opportunities.user_token_specified.clone(),
-            )
-            .await;
+            let program_id = instruction.program_id;
+            let transaction = Transaction::new_with_payer(
+                &[instruction.clone(), swap_instruction.clone()],
```
guibescos: i don't think you need `swap_instruction` here and in the other tests that only call `check_approved_program_instruction`

auction-server/src/auction/service/verification.rs
```diff
@@ -248,31 +249,63 @@ impl Service {
...
+        &self,
+        program_id: &Pubkey,
+        ix: &CompiledInstruction,
+        tx: &VersionedTransaction,
+        user_wallet: &Pubkey,
+        ix_index: usize,
+    ) -> Result<(), RestError> {
+        if self
+            .check_approved_program_instruction(program_id, ix)
+            .is_ok()
```
guibescos: it's doesn't feel great that now we swallow the error from `check_approved_program_instruction`. this might make it harder to debug

guibescos (approved):

auction-server/src/auction/service/verification.rs
anihamde: yeah i agree. i refactored this to bubble up the error from `check_approved_program_instruction` if the user account is invoked

auction-server/src/auction/service/verification.rs
```diff
@@ -531,113 +534,80 @@ impl Service {
...
-                        found:    transfer_instruction.to,
-                    },
-                ));
-            }
-        }
-        // No transfer instruction is allowed
-        else if !transfer_instructions.is_empty() {
+        // otherwise, if user mint is not SOL, then user should not be system transferring any SOL
+        else if !user_transfer_instructions.is_empty() {
+            let user_transfer_instruction = user_transfer_instructions[0].clone();
```
guibescos: nit: unnecessary clone
I like the previous syntax with 
```
                transfer_instructions
                    .first()
                    .map(|instruction| instruction.index)
```

auction-server/src/api.rs
```diff
@@ -155,21 +153,17 @@ impl std::fmt::Display for InstructionError {
...
-                write!(f, "Transfer instruction is not allowed")
-            }
             InstructionError::CloseAccountInstructionNotAllowed => {
                 write!(f, "Close account instruction is not allowed")
             }
-            InstructionError::InvalidTransferInstructionsCount => {
-                write!(f, "Exactly one sol transfer instruction is required")
+            InstructionError::InvalidUserTransferInstructionsCount { found } => {
+                match found {
+                    0 => write!(f, "At least one sol transfer instruction from the user wallet account is required"),
```
guibescos: nit: maybe capitalize sol?

auction-server/src/auction/service/verification.rs
```diff
@@ -1085,15 +1054,15 @@ impl Service {
                     ..
                 } = transaction_data.accounts;
 
+                let quote_tokens = get_swap_quote_tokens(&opp);
+                let opportunity_swap_data = get_opportunity_swap_data(&opp);
+                self.check_svm_swap_bid_fields(bid_data, opportunity_swap_data, &quote_tokens)
```
guibescos: why the reorder here?

auction-server/src/auction/service/verification.rs
```diff
@@ -537,23 +536,40 @@ impl Service {
...
         transaction_data: &entities::BidTransactionDataSwap,
         opportunity_swap_data: &OpportunitySvmProgramSwap,
     ) -> Result<(), RestError> {
-        let user_transfer_instructions = self
-            .extract_transfer_instructions(tx, transaction_data.accounts.user_wallet)
-            .await?;
+        let transfer_instructions: Vec<TransferInstructionData> =
+            self.extract_transfer_instructions(tx).await?;
+
+        let relayer_transfer_instructions = transfer_instructions
```
guibescos, line 553: imo this feels like paranoia, one of the assumptions i had when writing this code was that relayer would never have any funds or permissions beyond being the relayer

auction-server/src/auction/service/verification.rs
```diff
@@ -583,8 +583,9 @@ impl Service {
 
         // User has to wrap Sol
         if transaction_data.accounts.mint_user == spl_token::native_mint::id() {
-            // Sometimes the user doesn't have enough SOL, but we want the transaction to fail in the Express Relay program with InsufficientUserFunds
-            // Therefore we allow the user to wrap less SOL than needed so it doesn't fail in the transfer instruction
+            // Sometimes the user doesn't have enough SOL, but we want the transaction to fail in the Express Relay program with InsufficientUserFunds.
+            // Sometimes the user will already have a WSOL account, and they don't need to wrap the full user_amount of SOL.
```
guibescos: I don't think we handle this case in `get_user_amount_to_wrap`

auction-server/src/auction/service/verification.rs
```diff
@@ -2752,6 +2753,85 @@ mod tests {
...
+                InstructionError::InvalidPayerInCreateAtaInstruction {
+                    expected: searcher.pubkey(),
+                    found:    relayer,
+                }
+            )
+        );
+    }
+
+    #[tokio::test]
+    async fn test_verify_bid_when_sol_transfer_relayer() {
```
guibescos: I think `test_verify_bid_when_relayer_transfer_instruction` already covered this case

guibescos (dismissed):
lgtm, the relayer stuff feels a bit paranoid

auction-server/src/auction/service/verification.rs
anihamde: yeah this is 100% true. doesn't hurt to have multiple layers of defense

auction-server/src/auction/service/verification.rs
anihamde: i should clarify: i meant the user already has a WSOL account, so if they didn't have enough SOL in their account to fund they would just wrap what they can and the tx could still succeed

auction-server/src/auction/service/verification.rs
anihamde: good callout, i did this earlier while implementing something but decided not to do that and forgot to undo it

anihamde: i do need the `user_wallet` definition ahead of the call to `validate_swap_transaction_instructions`, but apart from that i can redo the ordering to what it was before

